PR #42, opened 2022-06-06
AV-1741: Fast deploy with default snapshot

felipemadero:
- use ANR DeployBlockchains call to do multiple deploys on top of already started network 
- fast deploy, avoiding network restarts, by using default snapshot with preloaded validator groups (subnet ids)
- improve health interval checks to be more responsive at determining if a network had started or a subnet was deployed
- improve user messages in failure case, for commands: clean, stop, start
- replace binary/pluging paths in snapshots, to be able to load snapshots saved in different environments
- move snapshots dir to a CLI managed directory, to permit future usage/manipulation
- made start cmd to also start server and network if needed


cmd/root.go
```diff
@@ -94,6 +99,40 @@ func init() {
...
+	if err != nil {
+		// no logger here yet
+		fmt.Printf("failed creating the snapshots dir %s: %s\n", snapshotsDir, err)
+		os.Exit(1)
+	}
+
+	// Copy default snapshot if not present
+	defaultSnapshotPath := filepath.Join(snapshotsDir, "anr-snapshot-"+constants.DefaultSnapshotName)
+	if _, err := os.Stat(defaultSnapshotPath); os.IsNotExist(err) {
+		defaultSnapshotBytes, err := ioutil.ReadFile("defaultSnapshot.tar.gz")
```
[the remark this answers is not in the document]
felipemadero: I agree. Shall live outside the repo. Probably for the next release related to pkg management.

cmd/start.go
```diff
@@ -25,6 +28,17 @@ is already running or if no subnets have been deployed.`,
 }
 
 func startNetwork(cmd *cobra.Command, args []string) error {
+	sd := subnet.NewLocalSubnetDeployer(log, baseDir)
+
+	if err := sd.StartServer(); err != nil {
+		return err
+	}
+
+	avalancheGoBinPath, pluginDir, err := sd.GetBinPaths()
```
holisticode, line 37: It's kind of weird that `GetBinPaths` is hooked on the `LocalSubnetDeployer`. The paths are not really dependent on the deployer, but are general for the whole tool.

We can maybe change this in a future PR though.

cmd/start.go
```diff
@@ -35,30 +49,47 @@ func startNetwork(cmd *cobra.Command, args []string) error {
...
+		client.WithExecPath(avalancheGoBinPath),
+	}
+	_, err = cli.LoadSnapshot(
+		ctx,
+		snapshotName,
+		loadSnapshotOpts...,
+	)
 	if err != nil {
-		return fmt.Errorf("failed to start network with the persisted snapshot: %s", err)
+		if !strings.Contains(err.Error(), "already bootstrapped") {
```
holisticode, line 69: We are "institutionalizing" this pattern which I don't like much. We should not check for error texts to identify error conditions.

If you don't want to change this in this PR, we should have a TODO and ticket and do it later (soon).

cmd/start.go
```diff
@@ -35,30 +49,47 @@ func startNetwork(cmd *cobra.Command, args []string) error {
...
 	if err != nil {
 		return fmt.Errorf("failed waiting for network to become healthy: %s", err)
 	}
 
 	fmt.Println()
-	ux.Logger.PrintToUser("Network ready to use. Local network node endpoints:")
-	for _, u := range endpoints {
-		ux.Logger.PrintToUser(u)
+	if len(endpoints) == 0 {
+		ux.Logger.PrintToUser("Network ready to use. No local network node endpoints")
```
holisticode, line 86: Not sure I follow. Why can there be 0 endpoints and how does this mean that the network is ready to use? What does it mean "No local network node endpoints"?

cmd/stop.go
```diff
@@ -36,13 +37,17 @@ func stopNetwork(cmd *cobra.Command, args []string) error {
...
 	} else {
-		snapshotName = defaultSnapshotName
+		snapshotName = constants.DefaultSnapshotName
 	}
 
 	ctx := binutils.GetAsyncContext()
 
 	_, err = cli.RemoveSnapshot(ctx, snapshotName)
 	if err != nil {
+		if strings.Contains(err.Error(), "not bootstrapped") {
```
holisticode, line 47: Again, pattern checking the error string

pkg/subnet/local.go
```diff
@@ -146,19 +153,65 @@ func (d *SubnetDeployer) doDeploy(chain string, chain_genesis string) error {
...
+		client.WithPluginDir(pluginDir),
+		client.WithExecPath(avalancheGoBinPath),
+	}
+	loadSnapshotsInfo, err := cli.LoadSnapshot(
+		ctx,
+		constants.DefaultSnapshotName,
+		loadSnapshotOpts...,
+	)
+	if err != nil {
+		if !strings.Contains(err.Error(), "already bootstrapped") {
```
holisticode, line 166: string error checking

pkg/subnet/local.go
```diff
@@ -146,19 +153,65 @@ func (d *SubnetDeployer) doDeploy(chain string, chain_genesis string) error {
...
+		ux.Logger.PrintToUser("Network has already been booted. Wait until healthy...")
+	} else {
+		ux.Logger.PrintToUser("Booting Network. Wait until healthy...")
+	}
+
+	d.log.Debug(loadSnapshotsInfo.String())
+
+	subnetIDs, blockchainsInfo, _, err := d.WaitForHealthy(ctx, cli, d.healthCheckInterval)
+	if err != nil {
+		_ = binutils.KillgRPCServerProcess()
```
holisticode, line 178: This seems to have been here before (so not from your PR) but:
1. I wonder if we should do this here, if we return the `err` doesn't it kill the server anyways?
2. The err should maybe not be shadowed

pkg/subnet/local.go
```diff
@@ -146,19 +153,65 @@ func (d *SubnetDeployer) doDeploy(chain string, chain_genesis string) error {
 	ctx := binutils.GetAsyncContext()
 
 	ux.Logger.PrintToUser("VM ready. Trying to boot network...")
-	info, err := cli.Start(
+	loadSnapshotOpts := []client.OpOption{
```
holisticode, line 156: Not sure I follow, are you wanting to start the snapshot in all cases?

I am not sure I agree, but I am also not sure I disagree. 

If we do this, then the subnet ID will always be the same. Also the owner of the subnet will always be the same, and finally, the number of control keys is always the same and 1.

In short, I wonder if loading via snapshot should be an optional path, or an independent path. For example, load a snapshot first, and deploy to that snapshot. Or just deploy from scratch.

I have not made up my mind about it yet but it feels like an important decision. I might be wrong though.

pkg/subnet/local.go
```diff
@@ -146,19 +153,65 @@ func (d *SubnetDeployer) doDeploy(chain string, chain_genesis string) error {
...
+	d.log.Debug(loadSnapshotsInfo.String())
+
+	subnetIDs, blockchainsInfo, _, err := d.WaitForHealthy(ctx, cli, d.healthCheckInterval)
+	if err != nil {
+		_ = binutils.KillgRPCServerProcess()
+		return fmt.Errorf("failed to query network health: %s", err)
+	}
+
+	// choose next subnetID
+	sort.Strings(subnetIDs)
```
holisticode, line 183: So this is also something new here and I might just not understand.

Do we want to make all decisions for the user and deploy everything? What if that is not desired?
Also, why do have to do this in the first place? What is happening here? Aren't this subnets and blockchains supposed to be in the snapshot?

I would really like to see more documentation about what is going on here.

pkg/subnet/local.go
```diff
@@ -146,19 +153,65 @@ func (d *SubnetDeployer) doDeploy(chain string, chain_genesis string) error {
...
+	subnetId := ""
+	if len(subnetIDs) > 0 {
```
holisticode, line 185: What if it's not > 0? `subnetId` is empty? But then why do all this? And if it's > 1? Why is it not a loop?

pkg/subnet/local.go
```diff
@@ -146,19 +153,65 @@ func (d *SubnetDeployer) doDeploy(chain string, chain_genesis string) error {
...
 
-	d.log.Debug(info.String())
-	ux.Logger.PrintToUser("Network has been booted. Wait until healthy. Please be patient, this will take some time...")
+	d.log.Debug(deployBlockchainsInfo.String())
+
+	fmt.Println()
+	ux.Logger.PrintToUser("Blockchain has been deployed. Wail until network acknowledges...")
 
-	endpoints, err := d.WaitForHealthy(ctx, cli, d.healthCheckInterval)
+	_, _, endpoints, err := d.WaitForHealthy(ctx, cli, d.healthCheckInterval)
```
holisticode, line 214: This also feels odd, sometimes we check and sometimes we just ignore the returns. It feels like the pattern could be refactored...

pkg/subnet/local.go
```diff
@@ -280,47 +333,56 @@ func (d *SubnetDeployer) setupLocalEnv() (string, error) {
...
+				d.log.Debug("network is not healthy. polling again...")
 				continue
 			}
 			if !resp.ClusterInfo.CustomVmsHealthy {
 				d.log.Debug("network is up but custom VMs are not healthy. polling again...")
 				continue
 			}
+			d.log.Debug("network is up and custom VMs are up")
+			blockchains := []string{}
+			for vmID, vmInfo := range resp.ClusterInfo.CustomVms {
```
holisticode, line 375: I wonder if we are overloading the Health check here. What do you think of storing this into the `d *SubnetDeployer` struct for example, or some other means? `WaitForHealthy` would then really just do that, and the storing of data (for later printing) would be done elsewhere?

cmd/stop.go
```diff
@@ -36,13 +37,17 @@ func stopNetwork(cmd *cobra.Command, args []string) error {
 	if len(args) > 0 {
 		snapshotName = args[0]
 	} else {
-		snapshotName = defaultSnapshotName
+		snapshotName = constants.DefaultSnapshotName
```
holisticode, line 40: Are we overriding the default snapshot here? What if we cleanup the snapshot directory, or it is gone, the doesn't this mean that we loose the first snapshot default from the installation? Resulting the in the tool not being usable anymore?

holisticode (commented):
Nice job with the snapshots.

I am convinced we want to use snapshots, but I wonder if we want to make this the only path to use the tool. I personally prefer the snapshot path to be optional, but you're making it the only way to use the tool if I understand correctly. Let's discuss if that's really the best way.

Also, there are parts of the code which I didn't understand with the new blockchain deploy. Please add some more comments.

cmd/root.go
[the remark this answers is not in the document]
felipemadero: Really the bootstrap snapshot install shall be downloaded, because for the install script option, we don't have the repo available. Either downloaded+installed from the install script and/or downloaded+installed at first deploy or start time. Will work on the second option for the moment, and download it (for the moment) from the repo.

A note, I think install script shall download avalanchego, evm, snapshot, so first deploy time is much faster, and we conveniently split the waiting times. Also think subnet create command shall download the pluging binary. Same criteria. Finally, install script can also be used in local, so shall replace build script.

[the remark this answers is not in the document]
felipemadero: In this PR the bootstrap snapshot is a resource needed for fast deploy. I see no inconvenient of using it always, but we can decide on that.

cmd/deploy.go
```diff
@@ -32,8 +32,9 @@ to interact with the subnet.
 
 Subsequent calls of deploy using the same subnet configuration will
 redeploy the subnet and reset the chain state to genesis.`,
-	RunE: deploySubnet,
-	Args: cobra.ExactArgs(1),
+	RunE:         deploySubnet,
+	Args:         cobra.ExactArgs(1),
+	SilenceUsage: true,
```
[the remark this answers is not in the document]
felipemadero: Without `SilenceUsage`, any error on deploy will cause a help text. But there are many possible errors that are unrelated to bad command usage, and it is a little anoying to the that help for all situations. Is there a way to only print help if the user forgets to give the required argument, but not for the other error situations?

cmd/start.go
felipemadero: maybe I can change the name to SetupLocalEnv, it is really that previous function (setupLocalEnv) with extra checks

cmd/start.go
felipemadero: Done both. https://ava-labs.atlassian.net/browse/AV-1749

cmd/start.go
felipemadero: Usually this and deploy prints `Local network node endpoints` before the blockchains list. As in this case there are no blockchains yet, only an empty network running, It seemed reasonable to me to print that. But we can:
1) change the message
2) avoid the situation of starting a network without blockchains (a little more work to have some state for the default snapshot contents but can be done)

pkg/subnet/local.go
felipemadero: you are right, the caller does kill the server, will fix this

pkg/subnet/local.go
felipemadero: Current released version has almost all of this, all beign done by ANR (same owner, same number of keys, not sure about subnet ID but it is pretty much repeated on tests). In this case we will just saving time on that same operations. I think maybe we can have this fast option as default, and give a way to deploy from scratch for the case we want to give the user more options (change the subnet owner, change the control key number, but anyway that requires new commands/arguments in the CLI and of course ANR).

pkg/subnet/local.go
felipemadero: The subnets IDs are in the snapshot but not the blockchains. The blockchain is going to be created here. We are selecting  (among the available ones in snapshot) a subnet ID to be used when creating the blockchain. 
It is pretty similar to version without snapshot but now the process of subnet id creation was done previously.
I don't understand on the deploy everything part of the comment

felipemadero: adding this comment:
// in order to make subnet deploy faster, a set of validated subnet IDs is preloaded                                 // in the bootstrap snapshot                                                                                        
// we select one to be used for creating the next blockchain, for that we use the                                    // number of currently created blockchains as the index to select the next subnet ID,                               // so we get incremental selection

pkg/subnet/local.go
felipemadero: if there is no set of preloaded subnet ids, uses the empty subnet id (and so a new subnet id will be created by ANR as before). in this case it is > 0 always with the bootstrap snapshot, except in unit tests.

pkg/subnet/local.go
felipemadero: The code has to wait twice, and use different info for different wait results. What do you sugest for this case?

pkg/subnet/local.go
felipemadero: Ok maybe WaitForHealthy can just return ClusterInfo, and the caller creates the other stuff

cmd/stop.go
felipemadero: yes, stop is supposed to overwritte the default snapshot if not given args. if the snapshot directory is deleted by some way (currently the tool does not do that), it will be recreated on init (or where that first copy will do). 
really default snapshot and bootstrap snapshot (the one in the tar) coincide when the tool starts, or after the hidden snapshots dir or default snapshot subdir is deleted.
after that, subsecuent stop will change the default snapshot, to save users work on top of the original one

cmd/root.go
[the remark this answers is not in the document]
felipemadero: Other suggestion related to installation. Maybe we can add a hidden command to the CLI to download deps, and make the install script to execute it (instead of duplicating download logic in the script). Just ideas.

pkg/subnet/local.go
felipemadero: Changed the return of WaitForHealthy to be more general and the callers to create the remaining info

pkg/subnet/local.go
```diff
@@ -123,54 +122,171 @@ func (d *Deployer) doDeploy(chain string, chainGenesis string) error {
...
-		client.WithGlobalNodeConfig("{\"log-level\":\"debug\", \"log-display-level\":\"debug\"}"),
-		client.WithRootDataDir(runDir),
+	ctx := binutils.GetAsyncContext()
+
+	// check for network and get VM info
+	needsStart := false
+	clusterInfo, err := d.WaitForHealthy(ctx, cli, d.healthCheckInterval)
+	if err != nil {
+		// TODO: use error type not string comparison
+		if strings.Contains(err.Error(), "not bootstrapped") {
```
holisticode, line 137: ?
Is `WaitForHealthy` ever returning this message?

pkg/subnet/local.go
```diff
@@ -123,54 +122,171 @@ func (d *Deployer) doDeploy(chain string, chainGenesis string) error {
...
-	if err := d.binaryDownloader.Download(vmID, pluginDir, binDir); err != nil {
+	if err := d.binaryDownloader.Download(toInstallVMIDs, pluginDir, binDir); err != nil {
 		return err
 	}
 
-	ctx := binutils.GetAsyncContext()
+	ux.Logger.PrintToUser("VM ready.")
+
+	// do start/restart as needed
+	if needsStart {
```
holisticode, line 173: This looks overly complex... can we not simplify it?

It's difficult to follow, and the function has become pretty large...

pkg/subnet/local.go
```diff
@@ -123,54 +122,171 @@ func (d *Deployer) doDeploy(chain string, chainGenesis string) error {
...
+			loadSnapshotOpts...,
+		)
+		if err != nil {
+			return fmt.Errorf("failed to start network :%s", err)
+		}
+		d.app.Log.Debug(loadSnapshotsInfo.String())
+	} else {
+		// we need to restart the network if the VM was not deployed previously
+		needsRestart := false
+		_, ok := deployedVMIDs[toDeployVMID.String()]
```
holisticode, line 192: Can't this be simplified by 

```
if _,ok := deployedVMIds[toDeployVMID.String()]; !ok {
     // just do all the restart stuff, no need for the `needsRestart`?
```

pkg/subnet/local.go
```diff
@@ -123,54 +122,171 @@ func (d *Deployer) doDeploy(chain string, chainGenesis string) error {
...
+		if !ok {
+			needsRestart = true
+		}
+
+		if needsRestart {
+			// make snapshot of current state and reload it again
+			tmpSnapshotName := fmt.Sprintf("restart-tmp-%d", time.Now().Unix())
```
holisticode, line 199: Do we really need a tmp snapshot? Can't we just override the snapshot and restart? It will be done anyways when we stop this network afterwards?

pkg/subnet/local.go
holisticode: OK. However, this only works for subnets where there are no controlKeys and thresholds to be observed...which might be ok for many situations but maybe not for all...

That's why I insisted that in the ANR PR we always allow to create subnets via normal path, and not via pre-creation always.

pkg/subnet/local.go
holisticode: Ok this is clear now. 

For the records, I still don't like very much that we pre-create subnets. I think the user should always be in control. They should create a subnet, and if they save that to **their** snapshot, then they can reuse it. I think this is overcomplicating things, as we can see in fact here.

pkg/subnet/local.go
```diff
@@ -123,54 +122,171 @@ func (d *Deployer) doDeploy(chain string, chainGenesis string) error {
...
+	sort.Strings(subnetIDs)
+	subnetID := ""
+	// in unit tests, there are no preloaded subnets IDs
+	// also, for the case the network does not contain subnet IDs, empty subnet ID
+	// will make ANR to create one when creating the blockchain
+	if len(subnetIDs) > 0 {
+		subnetID = subnetIDs[numBlockchains%len(subnetIDs)]
+	}
+
+	blockchainSpecs := []*rpcpb.BlockchainSpec{
```
holisticode, line 252: Add documentation here why you need to do this now against how it was before. 

Please also add some general flow description to the start of the function about what is happening during a complete deploy, there's a lot going on and I needed quite some time to understand all pieces. It is only clear to us because of the way we changed the ANR, otherwise I think a reader will have a hard time.

pkg/subnet/local.go
```diff
@@ -358,3 +500,89 @@ func getGenesis(genesisFile string) (core.Genesis, error) {
...
+	}
+	return nil
+}
+
+func GetIndexer(uri string) (indexer.Client, error) {
+	return indexer.NewClient(uri), nil
+}
+
+// Get list of latest blockchain for each vm, by using indexer API
+func GetLatestBlockchains(ctx context.Context, cli client.Client, getIdxFunc getIndexerFunc) (map[ids.ID]ids.ID, error) {
```
holisticode, line 533: I don't understand why we are doing this and why it is necessary

cmd/deploy.go
```diff
@@ -32,9 +32,8 @@ to interact with the subnet.
 
 Subsequent calls of deploy using the same subnet configuration will
 redeploy the subnet and reset the chain state to genesis.`,
-	RunE:         deploySubnet,
-	Args:         cobra.ExactArgs(1),
-	SilenceUsage: true,
```
dasconnor: I think we want to keep `SilenceUsage`

cmd/mocks/client.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.10.4. DO NOT EDIT.
+// Code generated by mockery v2.10.6. DO NOT EDIT.
```
dasconnor, line 1: How do we generate this?

cmd/start.go
```diff
@@ -56,16 +58,26 @@ func startNetwork(cmd *cobra.Command, args []string) error {
...
+		client.WithRootDataDir(outputDir),
 	}
 	_, err = cli.LoadSnapshot(
 		ctx,
 		snapshotName,
 		loadSnapshotOpts...,
 	)
+
 	if err != nil {
+		// TODO: use error type not string comparison
```
dasconnor, line 80: Do we have a ticket for this TODO?

dasconnor (commented):
I think this is basically ready, couple minor comments

dasconnor:
Are all of the network runner PRs that this depends on already merged?

cmd/deploy.go
felipemadero: https://ava-labs.atlassian.net/browse/AV-1841

cmd/mocks/client.go
felipemadero: Basically `mockery --all` is executed in the directory where the interface resides. In this case, ANR client directory. The file is then renamed and moved to destiny. Args of mockery can be used to avoid renaming. A script in ANR does the automatic generation for mocks that ANR needs. Maybe we can add also this kind of script to CLI

felipemadero: https://ava-labs.atlassian.net/browse/AV-1842

cmd/start.go
felipemadero: https://ava-labs.atlassian.net/browse/AV-1749

cmd/mocks/client.go
dasconnor: perfect, I was modding those files by hand, glad to know.

cmd/start.go
```diff
@@ -52,23 +65,44 @@ func startNetwork(cmd *cobra.Command, args []string) error {
...
 		return fmt.Errorf("failed waiting for network to become healthy: %s", err)
 	}
 
+	endpoints := subnet.GetEndpoints(clusterInfo)
+
 	fmt.Println()
-	ux.Logger.PrintToUser("Network ready to use. Local network node endpoints:")
-	for _, u := range endpoints {
-		ux.Logger.PrintToUser(u)
+	if len(endpoints) == 0 {
```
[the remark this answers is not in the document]
felipemadero: fixed

pkg/binutils/binaries.go
```diff
@@ -309,7 +309,9 @@ func cleanupPluginDir(vmID, pluginDir string) error {
 
 	pluginWhiteList := map[string]struct{}{
 		"evm": {},
-		vmID:  {},
+	}
+	for vmID := range vmIDs {
```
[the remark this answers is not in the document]
felipemadero: I think it is not good practice to modify an input param. It cames from Download() then intro cleanupDir(). Modifying it may introduce problems. Better not touching it.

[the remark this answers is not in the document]
holisticode: Ah you are very right here, I thought for a moment that it was a copy of the object, but golang does indeed change the origin object when it's a map.

dasconnor (approved):

PR #44, opened 2022-06-06
Add Go back and Capture Token Name

dasconnor:
I restructured the creation flow to allow all commands to go back. I also added chainId collection and added it to the sidecar. This PR is starting to get large, so I'll update the chainId output in a future PR.

It may be easier to review this PR by looking at the first commit and then at the rest. I moved a few functions into their own files. The actual edits to those functions are pretty small. All the copy/paste is in the first commit. The second and third commits have the meat.


cmd/create.go
```diff
@@ -103,15 +103,16 @@ func createGenesis(cmd *cobra.Command, args []string) error {
...
 		switch subnetType {
 		case subnetEvm:
-			genesisBytes, err = vm.CreateEvmGenesis(subnetName, log)
+			genesisBytes, tokenName, err = vm.CreateEvmGenesis(subnetName, log)
 			if err != nil {
 				return err
 			}
 
-			err = createSidecar(subnetName, models.SubnetEvm)
+			err = createSidecar(subnetName, models.SubnetEvm, tokenName)
```
holisticode, line 115: Reminder for the 
```
if err := ....; err != nil {
```
pattern for 1-value assignments from funcs

cmd/utils_test.go
```diff
@@ -78,11 +78,12 @@ func Test_copyGenesisFile_failure(t *testing.T) {
 
 func Test_createSidecar_success(t *testing.T) {
 	subnetName := "TEST_subnet"
+	tokenName := "TEST"
 	sidecarFile := subnetName + sidecar_suffix
 	const vm = models.SubnetEvm
 
 	// Write sidecar
-	err := createSidecar(subnetName, vm)
+	err := createSidecar(subnetName, vm, tokenName)
```
holisticode, line 86: We might want to add a small test case when `tokenName` is `""`

pkg/vm/createEvm.go
```diff
@@ -5,43 +5,48 @@ package vm
...
-	if err != nil {
-		return []byte{}, err
-	}
-
-	allocation, err := getAllocation()
-	if err != nil {
-		return []byte{}, err
+	stage := startStage
+
+	var chainId *big.Int
```
holisticode, line 25:
```
var (
)
```

pkg/vm/createEvm.go
```diff
@@ -5,43 +5,48 @@ package vm
...
+		case startStage:
+			stage = descriptorStage
+		case descriptorStage:
+			chainId, tokenName, stage, err = getDescriptors()
+			fmt.Println("Creating token", tokenName)
+		case feeStage:
+			*conf, stage, err = getFeeConfig(*conf)
+		case airdropStage:
+			allocation, stage, err = getAllocation()
+		case precompileStage:
```
holisticode, line 41: Any reason why there isn't a `default` case? Should we add one (e.g. "invalid stage")?

pkg/vm/createEvm.go
```diff
@@ -5,43 +5,48 @@ package vm
...
+	var chainId *big.Int
+	var tokenName string
+	var allocation core.GenesisAlloc
+	var err error
+
+	for stage != doneStage {
+		switch stage {
+		case startStage:
+			stage = descriptorStage
+		case descriptorStage:
```
holisticode, line 34: This looks like a state machine:
I think it's clearer if the next stage would not come back from the function (unless, the next stage is dependent on the function's result, which doesn't seem to be the case), but is clearly set after the function is evaluated.

This allows to keep the stage flow in one place, helping to understand it more clearly.

pkg/vm/createEvm.go
```diff
@@ -5,43 +5,48 @@ package vm
...
+			chainId, tokenName, stage, err = getDescriptors()
+			fmt.Println("Creating token", tokenName)
```
holisticode, line 36: Is this a case for `ux.Logger.PrintToUser` ?
Also, I usually think that it's a good idea to log each stage change to the logs.

It might be a bit overkill here so leaving it to your judgment. But recommended.

pkg/vm/airdrop.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"errors"
+	"math/big"
+
+	"github.com/ava-labs/avalanche-cli/cmd/prompts"
+	"github.com/ava-labs/subnet-evm/core"
+)
+
+const stageAfterAirdrop = precompileStage
```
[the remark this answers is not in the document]
holisticode: As all stages are defined in `evm_settings.go`, it seems clearer to me to add these special stages there too, in order to avoid scattering these all over.

I understand your thinking, you're wanting to see them here in this file because these special ones are (probably) used just here. 

Nevertheless I vote for a central place.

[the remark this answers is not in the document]
holisticode: Finally, even if it seems a good idea to have "labels" on existing stages to help understand the current stage, I think it can quickly become confusing, and make refactoring complex and error prone if needed.

If possible, I would like to suggest to avoid labeling stages and just use the same stage name where the stage applied. Maybe different stage names could help.

Don't get me wrong, I love state machines :D

holisticode (commented):
This is cool.

And I love state machines.
But I suggest not to scatter state names into different files and also to not create labels on them. One name per stage.

cmd/create.go
dasconnor: good catch, addressed in a few places

pkg/vm/createEvm.go
dasconnor: added `default`

pkg/vm/createEvm.go
dasconnor: you know, stages are just ints, so maybe we could control stage ordering with addition, hmmmm

pkg/vm/descriptors.go
```diff
@@ -0,0 +1,45 @@
...
+	"math/big"
+
+	"github.com/ava-labs/avalanche-cli/cmd/prompts"
+)
+
+const stageAfterDescriptors = feeStage
+
+func getChainId() (*big.Int, error) {
+	// TODO check against known chain ids and provide warning
+	fmt.Println("Enter your subnet's ChainId. It can be any positive integer.")
```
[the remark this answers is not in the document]
dasconnor: done

pkg/vm/descriptors.go
```diff
@@ -0,0 +1,45 @@
...
+	chainId, err := prompts.CapturePositiveBigInt("ChainId")
+	if err != nil {
+		return nil, err
+	}
+
+	return chainId, nil
+}
+
+func getTokenName() (string, error) {
+	fmt.Println("Select a symbol for your subnet's native token")
```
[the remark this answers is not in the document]
dasconnor: done

pkg/vm/evm_settings.go
```diff
@@ -12,6 +12,19 @@ const (
 	GasLimit = 8_000_000
 
 	defaultAirdropAmount = "1000000000000000000000000"
+	goBackMsg            = "Go back to previous step"
+)
+
+type creationStage int64
```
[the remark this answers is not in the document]
dasconnor: I liked `wizardState`, thanks for the suggestions

pkg/vm/fees.go
```diff
@@ -0,0 +1,113 @@
...
+		feeConfigOptions,
+	)
+	if err != nil {
+		return config, errored, err
+	}
+
+	switch feeDefault {
+	case useFast:
+		StarterFeeConfig.TargetGas = fastTarget
+		config.FeeConfig = StarterFeeConfig
```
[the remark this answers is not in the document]
dasconnor: how would you do that? since we're assigning by value and not reference, I think we need to do it each time

pkg/vm/precompiles.go
```diff
@@ -167,67 +170,70 @@ func getPrecompiles(config params.ChainConfig) (params.ChainConfig, error) {
...
+			mintConfig, cancelled, err := configureMinterList()
 			if err != nil {
-				return config, err
+				return config, errored, err
 			}
-
-			switch precompileDecision {
-			case nativeMint:
-				mintConfig, cancelled, err := configureMinterList()
+			if !cancelled {
```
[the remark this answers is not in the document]
dasconnor: If the precompile is not cancelled, you apply its results and remove it from the list of precompiles to set. If it is cancelled, you discard the results and allow the user to select it again when they go to configure precompiles

pkg/vm/precompiles.go
```diff
@@ -167,67 +170,70 @@ func getPrecompiles(config params.ChainConfig) (params.ChainConfig, error) {
...
-			if len(remainingPrecompiles) == 1 {
-				return config, nil
 			}
+		case cancel:
+			return config, stageAfterPrecompile, nil
+		}
 
-		} else {
-			return config, nil
+		if len(remainingPrecompiles) == 1 {
```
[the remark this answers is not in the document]
dasconnor: I added a comment, but that's the state when all precompiles have been configured. There's a `cancel` option in the list, so that's the one element that's still in there.

cmd/create.go
```diff
@@ -153,8 +151,7 @@ func createGenesis(cmd *cobra.Command, args []string) error {
 			}
 			subnetType = models.VmTypeFromString(subnetTypeStr)
 		}
-		err = createSidecar(subnetName, subnetType)
-		if err != nil {
+		if err = createSidecar(subnetName, subnetType, ""); err != nil {
```
felipemadero, line 154: why not ask for token name for the case the genesis is being given?

pkg/vm/airdrop.go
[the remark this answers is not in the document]
dasconnor: I removed these labels and replaced them with a state direction/incrementing strategy

pkg/vm/createEvm.go
dasconnor: I implemented this new state incrementing/decrementing strategy

pkg/vm/createEvm.go
```diff
@@ -6,42 +6,64 @@ import (
...
+		err        error
+	)
 
-	allocation, err := getAllocation()
-	if err != nil {
-		return []byte{}, err
+	for stage != doneStage {
+		switch stage {
+		case startStage:
+			direction = forward
```
felipemadero, line 48: LGTM as far as you want to restart asking if the user says want to come back from descriptorStage

pkg/vm/createEvm.go
```diff
@@ -6,42 +6,64 @@ import (
...
+		tokenName  string
+		allocation core.GenesisAlloc
+		direction  stateDirection
+		err        error
+	)
 
-	allocation, err := getAllocation()
-	if err != nil {
-		return []byte{}, err
+	for stage != doneStage {
```
felipemadero, line 45: I shall put the stages definitions on this file, as it is the only one using them

pkg/vm/evm_settings.go
```diff
@@ -12,6 +12,27 @@ const (
...
+	doneStage
+	errored
+)
+
+type stateDirection int64
+
+const (
+	forward stateDirection = iota
+	backward
+	kill
```
felipemadero, line 35: stop may be more suitable?

felipemadero (commented):
LGTM. Just nits

cmd/utils_test.go
dasconnor: Added that test, also changed the behavior slightly when tokenName is "". Might change it when I integrate tokenName's more fully in a follow-up PR.

cmd/create.go
dasconnor: good idea, I'll address when I make my follow-up PR focusing on the tokenName

pkg/vm/evm_settings.go
dasconnor: great suggestion, changed

pkg/vm/createEvm.go
dasconnor: The interface is hard to go back from descriptorStage, so I made that one first haha

felipemadero (approved):
LGTM

pkg/vm/fees.go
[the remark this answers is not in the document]
holisticode: I mean `config` is passed as param to the function. Therefore 
`config.FeeConfig = StarterFeeConfig` can be set just before the switch (or anywhere earlier). 
Unless I am missing something.

holisticode (approved):

pkg/vm/fees.go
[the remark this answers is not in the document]
dasconnor: ah, makes sense, fixed

dasconnor:
This is ready to go, going to wait to merge this till we cut the version stabilizing release

PR #51, opened 2022-06-09
generate error for invalid status code of http get

felipemadero:


holisticode (approved):
Basically ok for me, but any reason why we are using `%v` to format the status code? It's just an `int`

dasconnor (approved):

PR #53, opened 2022-06-10
Do not allow duplicate chain IDs

holisticode:
This PR prevents a new chain with an existing chain ID to be created.
It also refactors global variables.

closes https://ava-labs.atlassian.net/browse/AV-1648 and https://ava-labs.atlassian.net/browse/AV-1716


cmd/delete.go
```diff
@@ -22,15 +23,15 @@ var deleteCmd = &cobra.Command{
 
 func deleteGenesis(cmd *cobra.Command, args []string) error {
 	// TODO sanitize this input
-	genesis := filepath.Join(baseDir, args[0]+genesis_suffix)
-	sidecar := filepath.Join(baseDir, args[0]+sidecar_suffix)
+	genesis := filepath.Join(app.GetBaseDir(), args[0]+constants.Genesis_suffix)
```
dasconnor, line 26: Can't we use the functions in `app` for this?

pkg/app/app.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (app *Avalanche) GetGenesisPath(subnetName string) string {
+	return filepath.Join(app.baseDir, subnetName+constants.Genesis_suffix)
+}
+
+func (app *Avalanche) GetSidecarPath(subnetName string) string {
+	return filepath.Join(app.baseDir, subnetName+constants.Sidecar_suffix)
+}
+
+func (app *Avalanche) WriteGenesisFile(subnetName string, genesisBytes []byte) error {
```
dasconnor, line 48: I'm glad we moved this functionality out of `cmd`. I was running into problems because I couldn't import these tools due to a cyclical import.

pkg/app/app_test.go
```diff
@@ -91,41 +96,46 @@ func Test_createSidecar_success(t *testing.T) {
...
 		{
 			name:              "no token name",
 			subnetName:        "TEST_subnet",
 			tokenName:         "",
 			expectedTokenName: "TEST",
+			chainID:           "888",
 		},
 	}
 
+	ap := newTestApp(t)
```
dasconnor, line 110: Should we add an AfterAll to this suite to clean up the created dir or just ignore since it is already tmp?

pkg/models/sidecar.go
```diff
@@ -7,4 +7,5 @@ type Sidecar struct {
 	Vm        VmType
```
dasconnor, line 7: Why don't we also add a version for future compatibility matching

pkg/vm/descriptors.go
```diff
@@ -4,19 +4,29 @@ import (
...
 	"github.com/ava-labs/avalanche-cli/ux"
 )
 
-func getChainId() (*big.Int, error) {
+func getChainId(app *app.Avalanche) (*big.Int, error) {
 	// TODO check against known chain ids and provide warning
 	ux.Logger.PrintToUser("Enter your subnet's ChainId. It can be any positive integer.")
 
-	chainId, err := prompts.CapturePositiveBigInt("ChainId")
+	chainID, err := prompts.CapturePositiveBigInt("ChainId")
```
dasconnor, line 15: I prefer `ChainId` over `ChainID`. We should adopt a universal standard there.

dasconnor (commented):
A couple things to consider. We probably want to print the chainId in the `subnet list` command. We also might want to think about the behavior if you provide the `--force` flag. We can add a check in the genesis if no chainId exists in the sidecar. Lastly, what happens if a user already has multiple subnets with the same chainId.

pkg/vm/descriptors.go
dasconnor: I believe it is `chainId` in the actual Eth-RPC spec: https://eips.ethereum.org/EIPS/eip-695

pkg/app/app_test.go
holisticode: Because we are creating the testdir with `t.Tempdir()`:
```
// TempDir returns a temporary directory for the test to use.
// The directory is automatically removed by Cleanup when the test and
// all its subtests complete.
```

pkg/vm/createCustom.go
```diff
@@ -6,18 +6,26 @@ import (
...
 
 	"github.com/ava-labs/avalanche-cli/cmd/prompts"
+	"github.com/ava-labs/avalanche-cli/pkg/app"
+	"github.com/ava-labs/avalanche-cli/pkg/models"
 	"github.com/ava-labs/avalanche-cli/ux"
-	"github.com/ava-labs/avalanchego/utils/logging"
 )
 
-func CreateCustomGenesis(name string, log logging.Logger) ([]byte, error) {
+func CreateCustomGenesis(name string, app *app.Avalanche) ([]byte, *models.Sidecar, error) {
```
felipemadero, line 14: I think app is not used in the function

felipemadero (commented):
LGTM. Only one comment

cmd/delete.go
holisticode: I guess we can

pkg/models/sidecar.go
holisticode: OK, current strategy I adopted is to write the Version only on sidecar creation, with a different version than the app version (otherwise it's not clear when there needs to be a compatibility check).

Suggesting to apply sidecar version "1.0.0"

pkg/vm/descriptors.go
holisticode: I think it's a golang standard to use uppercase letters for such cases:
https://github.com/golang/go/wiki/CodeReviewComments#initialisms mentions ID explicitly.

I think we have applied this scheme in all other our golang repos.

pkg/vm/createCustom.go
holisticode: That is correct. But all the other similar functions take the app as a parameter, as it holds all shared information. It looked consistent to me.

Having said that, I can remove it if you insist.

holisticode:
> We probably want to print the chainId in the subnet list command
Added

> We also might want to think about the behavior if you provide the --force flag
Not sure I understand what you are aiming at. If the force flag is provided, the user first goes through the wizard the same way as for a new one, if I understand correctly. Therefore it would catch if the user is wanting to add an existing chain ID?

> Lastly, what happens if a user already has multiple subnets with the same chainId.
We don't have multiple subnets support yet, but we'll soon have. In any case, a subnet is supposed to be addressing different blockchains, and having the same chainIDs across different subnets _should_ not be a problem. If this is incorrect, we would need some global ID registry the same way as ethereum does, I guess, and probably not defined in this repo.
Nevertheless I am probably not the correct person to say anything conclusive/authoritative on this matter.

cmd/list.go
```diff
@@ -68,17 +69,29 @@ func listGenesis(cmd *cobra.Command, args []string) error {
...
 			if err != nil {
 				return err
 			}
 
+			chainID := sc.ChainID
+			// for older sidecars, check in genesis if sidecar has
+			// no chainID set
+			if chainID == "" {
+				sc, err := app.LoadEvmGenesis(carName)
+				if err != nil {
```
dasconnor, line 84: on error, we probably just want to use `""` for the chainId instead of failing the command

dasconnor (commented):
one small thing, then looks good

pkg/models/sidecar.go
dasconnor: yep, agreed it's a separate version

pkg/vm/descriptors.go
dasconnor: I don't love it, but I can live with it, I'll open a ticket to convert the other references.

pkg/vm/createCustom.go
holisticode: Fixed

cmd/list.go
holisticode: Applied

dasconnor (approved):
LGTM

PR #55, opened 2022-06-14
[AV-1736] Move network runner logs and configs from tmp to .avalanche-cli

dasconnor:
Place network runner logs in the `.avalanche-cli/runs` directory.

Requires a network runner update before merging: https://github.com/ava-labs/avalanche-network-runner/pull/167

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -8,6 +8,7 @@ import (
 	"github.com/ava-labs/avalanche-cli/pkg/binutils"
 	"github.com/ava-labs/avalanche-cli/pkg/subnet"
 	"github.com/ava-labs/avalanche-cli/ux"
+	"github.com/ava-labs/avalanche-network-runner/client"
 	"github.com/spf13/cobra"
 )
 
@@ -42,7 +43,7 @@ func startNetwork(cmd *cobra.Command, args []string) error {
 	ctx := binutils.GetAsyncContext()
 
 	ux.Logger.PrintToUser(startMsg)
-	_, err = cli.LoadSnapshot(ctx, snapshotName, app.GetRunDir())
+	_, err = cli.LoadSnapshot(ctx, snapshotName, client.WithRootDataDir(app.GetRunDir()))
 	if err != nil {
 		return fmt.Errorf("failed to start network with the persisted snapshot: %s", err)
 	}
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -4,6 +4,7 @@ package cmd
 
 import (
 	"fmt"
+	"path"
 
 	"github.com/ava-labs/avalanche-cli/pkg/binutils"
 	"github.com/ava-labs/avalanche-cli/pkg/subnet"
@@ -15,7 +16,7 @@ var startCmd = &cobra.Command{
 	Use:   "start [snapshotName]",
 	Short: "Starts a stopped local network",
 	Long: `The network start command starts a local, multi-node Avalanche network
-on your machine. If "snapshotName" is provided, that snapshot will be used for starting the network 
+on your machine. If "snapshotName" is provided, that snapshot will be used for starting the network
 if it can be found. Otherwise, the last saved unnamed (default) snapshot will be used. The command may fail if the local network
 is already running or if no subnets have been deployed.`,
 
@@ -42,7 +43,8 @@ func startNetwork(cmd *cobra.Command, args []string) error {
 	ctx := binutils.GetAsyncContext()
 
 	ux.Logger.PrintToUser(startMsg)
-	_, err = cli.LoadSnapshot(ctx, snapshotName)
+	rootDataDir := path.Join(app.GetBaseDir(), "runs")
+	_, err = cli.LoadSnapshot(ctx, snapshotName, rootDataDir)
 	if err != nil {
 		return fmt.Errorf("failed to start network with the persisted snapshot: %s", err)
 	}
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -4,11 +4,13 @@ package cmd
 
 import (
 	"fmt"
+	"path"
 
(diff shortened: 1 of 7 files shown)


pkg/app/app.go
```diff
@@ -17,6 +18,8 @@ import (
 
 const (
 	WriteReadReadPerms = 0o644
+
+	runDir = "runs"
```
felipemadero, line 22: probably shall go to constants.go (same as bin, and any other dirname for the tree under .avalanche-cli)

pkg/binutils/processes.go
```diff
@@ -111,12 +114,18 @@ func GetServerPID() (int, error) {
...
-func StartServerProcess(log logging.Logger) error {
+func StartServerProcess(app app.Avalanche) error {
 	thisBin := reexec.Self()
 
 	args := []string{"backend", "start"}
 	cmd := exec.Command(thisBin, args...)
-	outputFile, err := os.CreateTemp("", "avalanche-cli-backend*")
+
+	outputFilePrefix := path.Join(app.GetRunDir(), "server")
+	outputFilePath, err := utils.MkDirWithTimestamp(outputFilePrefix)
```
felipemadero, line 124: is this a file or a dir? names outputFilePrefix, etc... may be confusing

pkg/binutils/processes.go
```diff
@@ -111,12 +114,18 @@ func GetServerPID() (int, error) {
...
 // it just executes `avalanche-cli backend start`
-func StartServerProcess(log logging.Logger) error {
+func StartServerProcess(app app.Avalanche) error {
 	thisBin := reexec.Self()
 
 	args := []string{"backend", "start"}
 	cmd := exec.Command(thisBin, args...)
-	outputFile, err := os.CreateTemp("", "avalanche-cli-backend*")
+
+	outputFilePrefix := path.Join(app.GetRunDir(), "server")
```
felipemadero, line 123: why a dir with tmp for each execution? maybe only a dir like server-info, and inside it the different execution files with timestamps?

pkg/binutils/processes.go
```diff
@@ -141,7 +150,7 @@ func StartServerProcess(log logging.Logger) error {
 	}
 	err = os.WriteFile(constants.ServerRunFile, rfBytes, perms.ReadWrite)
```
felipemadero, line 151: probably we want to move ServerRunFile to .avalanche-cli?

pkg/app/app.go
dasconnor: done

pkg/binutils/processes.go
dasconnor: renamed

pkg/binutils/processes.go
dasconnor: good catch, fixed

cmd/start.go
```diff
@@ -42,7 +45,14 @@ func startNetwork(cmd *cobra.Command, args []string) error {
 	ctx := binutils.GetAsyncContext()
 
 	ux.Logger.PrintToUser(startMsg)
-	_, err = cli.LoadSnapshot(ctx, snapshotName)
+
+	outputDirPrefix := path.Join(app.GetRunDir(), "restart")
```
felipemadero, line 49: shall also use the current rundir of the server? as the deploy

pkg/binutils/processes.go
```diff
@@ -26,6 +29,12 @@ import (
 // gRPCTimeout is a common error message if the gRPC server can't be reached
 var gRPCTimeout = errors.New("Timed out trying to contact backend controller, it is most probably not running.")
 
+var latestRunDir string
+
+func GetLatestRunDir() string {
```
felipemadero, line 34: is it going to be available in a subsecuent deploy or start command? because the server will continue to be alive.
It seems the idea of putting the ANR server logs in the same directory as the ANR local network logs is ok for debugging purposes. 
But as the server lifetime exceeds that of a ANR local network (you can deploy, then stop, then deploy, stop, start, all using the same server).
The directory struct seems to be like this:
```
server directory
- ANR directory 1
- ...
- ANR directoryN
```
What do you think on this?

felipemadero: Anyway GetLatestRunDir maybe can be renamed to something related to backend or server rundir, and also, needs to be persistent in some way

cmd/start.go
dasconnor: since it's restarting, it gets a new folder

pkg/binutils/processes.go
dasconnor: Yeah, it works for the current structure but when we get subsequent deploys working, we might want to tweak things a little. It's hard to test right now though. We can actually grab the latest dir persistently by getting the latest timestamped directory.

pkg/subnet/local_test.go
```diff
@@ -164,7 +169,7 @@ func TestGetLatestAvagoVersion(t *testing.T) {
 func getTestClientFunc() (client.Client, error) {
 	c := &mocks.Client{}
 	fakeStartResponse := &rpcpb.StartResponse{}
-	c.On("Start", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(fakeStartResponse, nil)
+	c.On("Start", mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything).Return(fakeStartResponse, nil)
```
holisticode, line 172: Huh, why is this needed?

dasconnor: A test was failing, not sure why it didn't break the pipeline on main. This was required for it to pass.

holisticode (approved):

pkg/subnet/local_test.go
dasconnor: Looks like I added a new argument to start, h/t @holisticode

PR #60, opened 2022-06-16
[AV-1732] Print Token Name During Deploy

dasconnor:
Small PR, depends on https://github.com/ava-labs/avalanche-cli/pull/55 being deployed first.

### DIFF
--- a/pkg/subnet/local.go
+++ b/pkg/subnet/local.go
@@ -179,14 +179,7 @@ func (d *Deployer) doDeploy(chain string, chainGenesis string) error {
 	fmt.Println()
 	firstURL := endpoints[0]
 
-	tokenName := "TEST"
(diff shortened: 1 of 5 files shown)


pkg/subnet/local.go
```diff
@@ -172,6 +175,15 @@ func (d *SubnetDeployer) doDeploy(chain string, chain_genesis string) error {
 	fmt.Println()
 	firstURL := endpoints[0]
 
+	tokenName := "TEST"
+	sidecar, err := d.app.LoadSidecar(chain)
```
holisticode, line 179: Does this need to be applied to all `LoadSidecar` calls? If so, it might be nicer to just put this code inside `LoadSidecar`, to avoid every caller to have to do it.

Otherwise ignore this comment.

holisticode (approved):

PR #64, opened 2022-06-24
Attach server run file to app

holisticode:
This PR fixes a bug where the run file could never be found for removal nor PID check


tests/e2e/e2e_test.go
```diff
@@ -14,6 +15,9 @@ import (
 )
 
 func TestE2e(t *testing.T) {
+	if os.Getenv("RUN_E2E") == "" {
```
dasconnor, line 18: When/Why would we ever run this without this option?

dasconnor (approved):

PR #66, opened 2022-06-27
AV-1799 Refactor package structure

dasconnor:
I moved the `ux` package from the top level into `pkg` and split up the `cmd` package into a few subpackages. I only made a few substantive changes.

1. I moved a few constants around.
2. Instead of setting up all subcommands in the root command's init, I wrote a function that sets up the commands in each subcommand's root, so `subnet` is setup in `subnet.go` and not `root.go`. 
3. To accomplish 2., I had to do one weird thing, because the `app` variable isn't setup when `init()` runs, I needed to pass in `app` as a double pointer instead of a single.

This should not be merged until E2E has been merged and this branch has been run with those tests.


dasconnor:
This is finally ready for review :sweat_smile:

.github/workflows/lint-and-tests.yml
```diff
@@ -63,7 +63,11 @@ jobs:
       - name: Run e2e tests
         shell: bash
         run: scripts/run.e2e.sh
+      - name: Check Snapshots
```
felipemadero, line 66: probably we can remove this

.github/workflows/lint-and-tests.yml
```diff
@@ -63,7 +63,11 @@ jobs:
...
+        if: always()
+        run: ls /home/runner/.avalanche-cli/snapshots
       - name: 'Upload Artifact'
+        if: always()
```
felipemadero, line 70: why this if?

cmd/backendcmd/spawn_server.go
```diff
@@ -1,15 +1,18 @@
 // Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
-package cmd
+package backendcmd
```
felipemadero, line 3: I'm not sure about this cmd postfix. It is already inside cmd directory. Probably you can move other stuff to pkg (prompts, mocks), and only keep backend, network, subnet.

cmd/backendcmd/spawn_server.go
```diff
@@ -1,15 +1,18 @@
...
 import (
 	"context"
 	"fmt"
 
+	this "github.com/ava-labs/avalanche-cli/pkg/app"
 	"github.com/ava-labs/avalanche-cli/pkg/binutils"
 	"github.com/spf13/cobra"
 )
 
+var app **this.Avalanche
```
felipemadero, line 14: why don't we create app on init? I am not sure which is best, the double pointer is clever but still is a double pointer so generates questions.

cmd/constants.go
```diff
@@ -2,18 +2,8 @@
 // See the file LICENSE for licensing terms.
 package cmd
```
felipemadero, line 3: should be better to put this constants into root.go? seems to be used only there

cmd/networkcmd/network.go
```diff
@@ -1,13 +1,16 @@
 // Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
 // See the file LICENSE for licensing terms.
-package cmd
+package networkcmd
 
 import (
 	"fmt"
 
+	this "github.com/ava-labs/avalanche-cli/pkg/app"
```
felipemadero, line 8: I am not sure of using a package renamed to this for the app package. Probably will be best to decide on different names for the package and for the variable (not both app)

felipemadero (commented):
LGTM regarding file and constants movement. Left some comments.
Not sure about app double pointer should prefer instanciate app on init.
Made suggestions on file naming and package alias naming.

cmd/networkcmd/start.go
```diff
@@ -59,7 +59,7 @@ func startNetwork(cmd *cobra.Command, args []string) error {
 
 	ux.Logger.PrintToUser(startMsg)
 
-	outputDirPrefix := path.Join(app.GetRunDir(), "restart")
+	outputDirPrefix := path.Join((*app).GetRunDir(), "restart")
```
holisticode, line 62: This does not come from this PR but we should use `filepath.Join`. `path` targets mostly URLs, and can't handle all filesystems correctly.

cmd/backendcmd/spawn_server.go
dasconnor: It's probably a good idea to move those packages, but I do think we should keep the `-cmd` suffix. We have several packages with similar names (ex: `pkg/subnet` and `cmd/subnet`). I think we should name them distinctly to avoid confusion.

.github/workflows/lint-and-tests.yml
dasconnor: removed

.github/workflows/lint-and-tests.yml
dasconnor: It makes sure that artifacts still upload if the previous command fails. Otherwise, this step will be skipped and we will only get logs if the tests pass.

cmd/backendcmd/spawn_server.go
dasconnor: I changed this to eliminate the double pointer, the structure is largely unchanged.

cmd/constants.go
dasconnor: I ended up moving them to the constants package.

cmd/networkcmd/network.go
dasconnor: I renamed to `application`

cmd/networkcmd/start.go
dasconnor: Made a ticket for this: https://ava-labs.atlassian.net/browse/AV-1856

cmd/root.go
```diff
@@ -44,52 +47,18 @@ in with avalanche subnet create myNewSubnet.`,
...
-	// subnet describe
-	subnetCmd.AddCommand(readCmd)
-	readCmd.Flags().BoolVarP(
-		&printGenesisOnly,
-		"genesis",
-		"g",
-		false,
-		"Print the genesis to the console directly instead of the summary",
-	)
+	// We need to pass double pointers to app into these functions since app
```
felipemadero, line 51: probably this comment is somewhat obsolete

cmd/root.go
```diff
@@ -44,52 +47,18 @@ in with avalanche subnet create myNewSubnet.`,
...
-	// network stop
-	networkCmd.AddCommand(stopCmd)
-
-	// network clean
-	networkCmd.AddCommand(cleanCmd)
-
-	// network status
-	networkCmd.AddCommand(statusCmd)
+	// add hidden backend command
+	backend := backendcmd.SetupBackendCmd(app)
```
felipemadero, line 60: probably it will be nest to create and setup the command instance in one function, and do eg 
`rootCmd.AddCommand(backendcmd.NewCmd(app))`

felipemadero: same for other cobra.Command objects, try to give to each one its own NewCmd() function (if there are many in the same package, call it NewFooCmd, NewBarCmd)

cmd/networkcmd/start.go
```diff
@@ -30,7 +30,7 @@ is already running or if no subnets have been deployed.`,
 }
 
 func startNetwork(cmd *cobra.Command, args []string) error {
-	sd := subnet.NewLocalDeployer(app)
+	sd := subnet.NewLocalDeployer(*app)
```
holisticode: No action to be take for this PR, but I created a ticket https://ava-labs.atlassian.net/browse/AV-1857. Network commands should not have a need to use functionality of a subnet deployer object.

cmd/root.go
dasconnor: This was the way we originally developed it, but we decided to take a more centralized approach. I think this is a decent compromise.

cmd/root.go
dasconnor: removed

cmd/networkcmd/start.go
dasconnor: that's a good call

felipemadero (approved):

PR #76, opened 2022-07-04
remove network restart operation

felipemadero:
- avoid network restart when deploying a subnet over a started network, by using new ANR branch that calls admin.LoadVMs


dasconnor (approved):
LGTM

PR #78, opened 2022-07-05
Create key

dasconnor:
Add a command for creating a cryptographic key to use on Fuji.

I still plan to add some tests.


cmd/keycmd/create.go
```diff
@@ -0,0 +1,76 @@
...
+		Args:         cobra.ExactArgs(1),
+		RunE:         createKey,
+		SilenceUsage: true,
+	}
+
+	cmd.Flags().StringVar(
+		&filename,
+		"filename",
+		"",
+		"file path of genesis to use instead of the wizard",
```
holisticode, line 66: Probably not "genesis...wizard"

cmd/keycmd/create.go
```diff
@@ -0,0 +1,76 @@
...
+	)
+	cmd.Flags().BoolVarP(
+		&forceCreate,
+		forceFlag,
+		"f",
+		false,
+		"overwrite the existing configuration if one exists",
```
holisticode, line 73: Should the text here change as well?

cmd/keycmd/create.go
```diff
@@ -0,0 +1,76 @@
...
+		Long:         `Generates a new private key to use for creating and controlling test subnets.`,
+		Args:         cobra.ExactArgs(1),
+		RunE:         createKey,
+		SilenceUsage: true,
+	}
+
+	cmd.Flags().StringVar(
+		&filename,
+		"filename",
+		"",
```
holisticode, line 65: I wonder if we should actually suggest a default path name?

pkg/key/key.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+var (
+	ErrInvalidType = errors.New("invalid type")
+	ErrCantSpend   = errors.New("can't spend")
+)
+
+// Key defines methods for key manager interface.
+type Key interface {
+	// P returns all formatted P-Chain addresses.
```
holisticode, line 24: Huh? The interface is completely empty?

pkg/key/soft_key.go
```diff
@@ -0,0 +1,353 @@
...
+
+	keyChain *secp256k1fx.Keychain
+}
+
+const (
+	privKeyEncPfx = "PrivateKey-"
+	privKeySize   = 64
+
+	rawEwoqPk      = "ewoqjP7PxY4yr3iLTpLisriqt94hdyDFNgchSxGGztUrTXtNN"
+	EwoqPrivateKey = "PrivateKey-" + rawEwoqPk
```
holisticode, line 49: There is a Prefix const somewhere, I think it's in avalanchego.
If not, you just created it above anyways

pkg/key/soft_key.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// Returns the private key encoded in CB58 and "PrivateKey-" prefix.
+func (m *SoftKey) Encode() string {
+	return m.privKeyEncoded
+}
+
+// Saves the private key to disk with hex encoding.
+func (m *SoftKey) Save(p string) error {
+	k := hex.EncodeToString(m.privKeyRaw)
```
holisticode, line 264: This will save the hex encoded version of the file. I think there is no agreement and what should be used, but the ewoq key is also in the `PrivateKey-...` format, I personally prefer to write the `privKeyEncoded` version instead. There was some discussion about this in the `subnet-cli` repo. Maybe @aaronbuchwald wants to add some comments?

cmd/keycmd/create.go
dasconnor: fixed

cmd/keycmd/create.go
dasconnor: fixed

cmd/keycmd/create.go
dasconnor: That's not a suggestion, that's the value that `filename` gets if there's no flag provided I believe.

pkg/key/key.go
dasconnor: whoops, I accidentally deleted, restored

pkg/key/soft_key.go
dasconnor: fixed

pkg/key/soft_key.go
dasconnor: I don't really have strong feelings one way or the other.

dasconnor:
I updated the base branch to `fujiDeploy` which we can use for staging our v2 changes until they're ready for main.

cmd/keycmd/create.go
holisticode: I understood this is the filename where the key will be saved, not from where it would be imported. Maybe this should be clarified in the docs.

pkg/key/soft_key.go
holisticode: OK let's leave it at this, this is how subnet-cli has it too.

PR #79, opened 2022-07-05
[AV-1779] Deploy to fuji

holisticode:
Closes https://ava-labs.atlassian.net/browse/AV-1779


pkg/subnet/local.go
```diff
@@ -47,8 +47,8 @@ type Deployer struct {
 	setDefaultSnapshot  setDefaultSnapshotFunc
 }
 
-func NewLocalDeployer(app *application.Avalanche) *Deployer {
-	return &Deployer{
+func NewLocalSubnetDeployer(app *application.Avalanche) *LocalSubnetDeployer {
```
dasconnor, line 50: We decided a couple weeks ago to eliminate the duplicate `Subnet` in the name since that's the package name. I think we want to use the names `LocalDeployer` and `PublicDeployer`

cmd/subnetcmd/deploy.go
```diff
@@ -104,26 +110,161 @@ func deploySubnet(cmd *cobra.Command, args []string) error {
...
+		addCtrlKey = "Add control key"
+		doneMsg    = "Done"
+		cancelMsg  = "Cancel"
+	)
+
+	var controlKeys []string
+
+	for {
+		listDecision, err := prompts.CaptureList(
+			controlKeysPrompt, []string{addCtrlKey, doneMsg, cancelMsg},
```
dasconnor, line 222: maybe remove the `done` option if the controlKeys list has length 0

cmd/subnetcmd/deploy.go
```diff
@@ -104,26 +110,161 @@ func deploySubnet(cmd *cobra.Command, args []string) error {
...
+		)
+		if err != nil {
+			return nil, false, err
+		}
+
+		switch listDecision {
+		case addCtrlKey:
+			controlKey, err := prompts.CapturePChainAddress(
+				"Enter the addresses which control who can add validators to this subnet",
```
dasconnor, line 231: maybe specify this must be a p-chain address

pkg/models/sidecar.go
```diff
@@ -2,11 +2,15 @@
...
-	TokenName string
-	ChainID   string
-	Version   string
+	Name         string
+	VM           VMType
+	Subnet       string
+	TokenName    string
+	ChainID      string
+	Version      string
+	SubnetID     ids.ID
```
dasconnor, line 14: we might want to update `subnetcmd/describe.go`

pkg/subnet/public.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	sf, err := key.LoadSoft(networkID, d.privKeyPath)
+	if err != nil {
+		return nil, "", err
+	}
+
+	kc := sf.KeyChain()
+
+	wallet, err := primary.NewWalletFromURI(ctx, api, kc)
```
dasconnor, line 90: I don't believe this will work for a ledger, you can't export the keychain the same way, but I'm not 1000% sure

dasconnor (commented):
Nice job!

pkg/subnet/public.go
```diff
@@ -0,0 +1,121 @@
...
+	baseDir     string
+	privKeyPath string
+	network     models.Network
+	log         logging.Logger
+}
+
+func NewPublicSubnetDeployer(app *application.Avalanche, privKeyPath string, network models.Network) *PublicSubnetDeployer {
+	return &PublicSubnetDeployer{
+		LocalSubnetDeployer: *NewLocalSubnetDeployer(app),
+		baseDir:             app.GetBaseDir(),
```
felipemadero, line 33: why just don't keep the app instead of the app members?

cmd/subnetcmd/deploy.go
```diff
@@ -104,26 +110,161 @@ func deploySubnet(cmd *cobra.Command, args []string) error {
...
 	case models.Local:
 		app.Log.Debug("Deploy local")
-		// TODO: Add signal management here. If we Ctrl-C this guy it can leave
-		// the gRPC server is a weird state. Should kill that too
-		deployer := subnet.NewLocalDeployer(app)
-		chain := chains[0]
-		chainGenesis := filepath.Join(app.GetBaseDir(), fmt.Sprintf("%s_genesis.json", chain))
-		err := deployer.DeployToLocalNetwork(chain, chainGenesis)
-		if err != nil {
+		deployer := subnet.NewLocalSubnetDeployer(app)
```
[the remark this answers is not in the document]
holisticode: Because it refers to an old version, unless I am confusing something. If we do Ctrl-C now it **will** kill the server _unless_ started from a different run (`BackendStartedHere`).

cmd/subnetcmd/deploy.go
```diff
@@ -104,26 +110,161 @@ func deploySubnet(cmd *cobra.Command, args []string) error {
...
+		if err := deployer.DeployToLocalNetwork(chain, chainGenesis); err != nil {
 			if deployer.BackendStartedHere() {
 				if innerErr := binutils.KillgRPCServerProcess(app); innerErr != nil {
 					app.Log.Warn("tried to kill the gRPC server process but it failed: %w", innerErr)
 				}
 			}
 		}
 		return err
+	case models.Fuji: // just make the switch pass
+		if keyName == "" {
```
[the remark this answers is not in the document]
holisticode: ok

cmd/subnetcmd/deploy.go
```diff
@@ -104,26 +110,161 @@ func deploySubnet(cmd *cobra.Command, args []string) error {
...
+			return errors.New("this command requires the name of the private key")
+		}
+
+	case models.Mainnet: // just make the switch pass, fuij/main implementation is the same (for now)
```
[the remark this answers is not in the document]
holisticode: Now that we will be prompting it's not possible to provide an empty one

cmd/subnetcmd/deploy.go
```diff
@@ -104,26 +110,161 @@ func deploySubnet(cmd *cobra.Command, args []string) error {
...
+	}
+	sidecar.SubnetID = subnetID
+	sidecar.BlockchainID = blockchainID
+	return app.UpdateSidecar(&sidecar)
+}
+
+func getControlKeys() ([]string, bool, error) {
+	controlKeysPrompt := "Configure which addresses allow to add new validators"
+	// TODO: is this text ok?
+	noControlKeysPrompt := "You did not add any control key. This means anyone can add validators to your subnet. " +
```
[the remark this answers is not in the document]
holisticode: OK we discussed that we want to promote "good" behavior so I am removing this, making it mandatory to have at least one key

cmd/subnetcmd/deploy.go
holisticode: New approach is to keep asking for keys

pkg/models/sidecar.go
holisticode: Added to `describe` command

pkg/prompts/prompts.go
```diff
@@ -61,6 +92,26 @@ func CapturePositiveBigInt(promptStr string) (*big.Int, error) {
 	return amountInt, nil
 }
 
+func validatePChainAddress(input string) error {
+	_, _, _, err := address.Parse(input)
```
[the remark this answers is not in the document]
holisticode: Not sure I understand. This just would have every valid address parse, regardless of which network it is in?

pkg/prompts/prompts.go
```diff
@@ -61,6 +92,26 @@ func CapturePositiveBigInt(promptStr string) (*big.Int, error) {
...
+
+	return err
+}
+
+func CapturePChainAddress(promptStr string) (string, error) {
```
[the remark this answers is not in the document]
holisticode: I am not aware of an actual complete `Address` representation. We use the Bech32 addresses, so `address.Parse` for example returns a ChainID part ("P"), a HRP part ("-custom". "-fuji") and the actual bytes. We are later populating the `OutputOwners` field, which just takes `ShortID` - that'd be the bytes representation part of the address. Thus, if we'd just pass that part, we loose all ChainID and HRP information.

Thus it seems better to move around the string until the actual parts are needed. Unless there is a better way that  is.

pkg/subnet/public.go
holisticode: I have had other PRs rejected because I was passing the whole object instead of the individual parameters...I think you yourself have made that objection in the past...

I don't know if there is any guideline here...
¯\_(ツ)_/¯

pkg/subnet/public.go
```diff
@@ -0,0 +1,121 @@
...
+	options := []common.Option{}
+	fxIDs := make([]ids.ID, 0)
+	return wallet.P().IssueCreateChainTx(subnetID, genesis, vmID, fxIDs, chainName, options...)
+}
+
+func (d *PublicSubnetDeployer) createSubnetTx(controlKeys []string, threshold uint32, wallet primary.Wallet) (ids.ID, error) {
+	var err error
+
+	addrs := make([]ids.ShortID, len(controlKeys))
+	for i, c := range controlKeys {
```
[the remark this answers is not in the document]
holisticode: Nice one

tests/e2e/hardhat/yarn.lock
```diff
@@ -3318,12 +3318,12 @@ create-require@^1.1.0:
   integrity sha512-dcKFX3jn0MpIaXjisoRvexIJVEKzaq7z2rZKxf+MSr9TkdmHmsU4m2lcLojrj/FHl8mk5VxMmYA+ftRkP/3oKQ==
```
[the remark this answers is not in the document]
holisticode: Because I first merged from main to this branch, then realized I actually had to do this PR to `fujiDeploy`....

pkg/subnet/public.go
dasconnor: alright, we agreed to defer this until we actually implement ledger integration

pkg/prompts/prompts.go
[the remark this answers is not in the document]
dasconnor: Looks like I was looking at an outdated diff, I think this is resolved.

pkg/prompts/prompts.go
```diff
@@ -109,6 +117,26 @@ func CapturePChainAddress(promptStr string) (string, error) {
 		return "", err
 	}
 
+	_, hrp, _, err := address.Parse(addressStr)
+	if err != nil {
+		return "", err
+	}
+	switch network {
+	case models.Fuji:
```
dasconnor, line 125: We should push these checks into the validate function rather than capture. All of our prompt functions return err, but they never should, even when the user has bad input. (I actually don't know what can trigger that error)

dasconnor: It might be a little ugly, you might need to make multiple validation functions and pick between them, but I think that's a better approach.

dasconnor (commented):
Looks very close to me!

pkg/prompts/prompts.go
holisticode: Fixed by providing a `getValidationFunc` based on the `network` param

felipemadero (approved):
LGTM

PR #80, opened 2022-07-06
[AV-1834, AV-1810] Add key list

dasconnor:
Adds `key list` command. I also made an improvement so that the `subnet list` command won't get tripped up by temporary files.

### DIFF
--- a/cmd/keycmd/list.go
+++ b/cmd/keycmd/list.go
@@ -39,7 +39,7 @@ func listKeys(cmd *cobra.Command, args []string) error {
 	for _, f := range files {
 		if strings.HasSuffix(f.Name(), constants.KeySuffix) {
 			filename := f.Name()
-			rows = append(rows, []string{filename[:len(filename)-len(constants.KeySuffix)]})
+			rows = append(rows, []string{strings.TrimSuffix(filename, constants.KeySuffix)})
 		}
 	}
 	for _, row := range rows {
--- a/cmd/keycmd/list.go
+++ b/cmd/keycmd/list.go
@@ -1,0 +1,50 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package keycmd
+
+import (
+	"os"
+	"strings"
+
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+	"github.com/olekukonko/tablewriter"
+	"github.com/spf13/cobra"
+)
+
+// avalanche subnet list
+func newListCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "list",
+		Short: "List all created signing keys",
+		Long: `The key list command prints the names of all created signing
+keys.`,
+		RunE:         listKeys,
+		SilenceUsage: true,
+	}
+}
+
+func listKeys(cmd *cobra.Command, args []string) error {
+	header := []string{"Key Name"}
+	table := tablewriter.NewWriter(os.Stdout)
+	table.SetHeader(header)
+	table.SetRowLine(true)
+
+	files, err := os.ReadDir(app.GetKeyDir())
+	if err != nil {
+		return err
+	}
+
+	rows := [][]string{}
+
+	for _, f := range files {
+		if strings.HasSuffix(f.Name(), constants.KeySuffix) {
+			filename := f.Name()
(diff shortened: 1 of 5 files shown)


tests/e2e/key/suite.go
```diff
@@ -117,4 +117,36 @@ var _ = ginkgo.Describe("[Key]", func() {
...
+		}
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		// The matcher for this test is a little weird. Instead of matching an exact
+		// string, we check that it contains all of a few necessary substrings. This
+		// is to facilitate running the test locally. If you have other keys in your
+		// key directory, they will be printed as well. It's impossible to check the
+		// list output for exact equality without removing pre-existing user keys.
+		// Hence, the matcher here.
+		gomega.Expect(output).Should(gomega.ContainSubstring(expectedLine1))
```
felipemadero, line 148: probably you may use a regex to include all 3 things in one check. also a regex will avoid problems related to key length (if there is any)

tests/e2e/key/suite.go
```diff
@@ -117,4 +117,36 @@ var _ = ginkgo.Describe("[Key]", func() {
 		_, err = commands.CreateKey(keyName)
 		gomega.Expect(err).Should(gomega.HaveOccurred())
 	})
+
+	ginkgo.It("can list a created key", func() {
+		expectedLine1 := "+----------+"
+		expectedLine2 := "| KEY NAME |"
+		expectedLine3 := "| e2eKey   |"
```
felipemadero, line 124: are you going to have problems with different key lengths?

tests/e2e/key/suite.go
dasconnor: updated, I check a regex and that the expected key is in there

cmd/keycmd/list.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return err
+	}
+
+	rows := [][]string{}
+
+	for _, f := range files {
+		if strings.HasSuffix(f.Name(), constants.KeySuffix) {
+			filename := f.Name()
+			rows = append(rows, []string{filename[:len(filename)-len(constants.KeySuffix)]})
```
holisticode, line 42: Note: there is the `strings.TrimSuffix` method which makes it a bit more readable, but, on the other hand, slower, so up to you.

cmd/keycmd/list.go
```diff
@@ -0,0 +1,50 @@
...
+	files, err := os.ReadDir(app.GetKeyDir())
+	if err != nil {
+		return err
+	}
+
+	rows := [][]string{}
+
+	for _, f := range files {
+		if strings.HasSuffix(f.Name(), constants.KeySuffix) {
+			filename := f.Name()
```
holisticode, line 41: We could also do some validation here to make sure it's actually a private key and not something (veeeery accidentally....) with the suffix. Leaving to your judgement too.

tests/e2e/key/suite.go
```diff
@@ -117,4 +117,33 @@ var _ = ginkgo.Describe("[Key]", func() {
...
+			utils.PrintStdErr(err)
+		}
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		// The matcher for this test is a little weird. Instead of matching an exact
+		// string, we check that it matches a regex and contains created keyName. This
+		// is to facilitate running the test locally. If you have other keys in your
+		// key directory, they will be printed as well. It's impossible to check the
+		// list output for exact equality without removing pre-existing user keys.
+		// Hence, the matcher here.
```
holisticode, line 145: Note: if you were able to have the keys created in a tmp test dir, like we do with unit tests, it _would_ be possible. However, this being e2e tests, I understand that this probably amounts to unnecessary overhead. But worth having if possible.

holisticode (commented):
Basically all good for me. I left some notes which do not require intervention, unless you decide to.

cmd/keycmd/list.go
dasconnor: Eh, I don't think this is really worth it. I don't care so much about users trying to fuzz test us with bad inputs. There are a few other places where we make similar assumptions.

cmd/keycmd/list.go
dasconnor: updated

holisticode (approved):

PR #82, opened 2022-07-07
[AV-1837] Added export command

dasconnor:
Added the `key export` command. Also renamed the `DeleteKeys` function to `DeleteKey` since it only deletes one key.

### DIFF
--- a/cmd/keycmd/export.go
+++ b/cmd/keycmd/export.go
@@ -1,0 +1,49 @@
+package keycmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/ava-labs/avalanche-cli/pkg/application"
+	"github.com/spf13/cobra"
+)
+
+func newExportCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "export [keyName]",
+		Short: "Exports a signing key",
+		Long: `Exports a created signing key. By default, the tool writes the hex encoded key to
+stdout. If the --output flag is provided, the key will be written to a file of
+your choosing.`,
+		Args:         cobra.ExactArgs(1),
+		RunE:         exportKey,
+		SilenceUsage: true,
+	}
+
+	cmd.Flags().StringVarP(
+		&filename,
+		"output",
+		"o",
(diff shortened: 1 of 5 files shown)


cmd/keycmd/export.go
```diff
@@ -0,0 +1,49 @@
...
+	keyName := args[0]
+
+	keyPath := app.GetKeyPath(keyName)
+	keyBytes, err := os.ReadFile(keyPath)
+	if err != nil {
+		return err
+	}
+
+	if filename == "" {
+		fmt.Println(string(keyBytes))
```
holisticode, line 44: Note:This is not necessarily printing the key in hex encoded format, it just prints the content of the file. As we are writing the hex-encoded version of the key on `Save`, this implicitly holds. Nevertheless, wanted to have mentioned it.

tests/e2e/key/suite.go
```diff
@@ -117,4 +121,61 @@ var _ = ginkgo.Describe("[Key]", func() {
...
+		gomega.Expect(err).Should(gomega.BeNil())
+
+		// Check two keys are equal
+		equal, err := utils.CheckKeyEquality(testKey, outputKey)
+		if err != nil {
+			fmt.Println(output)
+			utils.PrintStdErr(err)
+		}
+		gomega.Expect(err).Should(gomega.BeNil())
+		gomega.Expect(equal).Should(gomega.BeTrue())
```
holisticode, line 179: Shouldn't we remove the test output file afterwards?

holisticode (commented):
Basically LGTM, a couple of nits I guess

felipemadero (approved):
LGTM

tests/e2e/key/suite.go
dasconnor: it gets cleaned up by the `AfterEach`

cmd/keycmd/export.go
dasconnor: I think this is fine for now. If we want to be a little more flexible, we can add more support in the future.

PR #88, opened 2022-07-12
Add Validator Join Command

### LINKED ISSUE #98: Add Validator Join Command

holisticode:
Closes https://ava-labs.atlassian.net/browse/AV-1840

Spec https://ava-labs.atlassian.net/wiki/spaces/AP/pages/2087747585/Command+Tree#Join-Existing-Subnet

Closes #98


cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,138 @@
...
+		"Choose a network to deploy on (this command only supports public networks)",
+		[]string{models.Fuji.String(), models.Mainnet.String()},
+	)
+	if err != nil {
+		return err
+	}
+	network = models.NetworkFromString(networkStr)
+	networkLower := strings.ToLower(network.String())
+
+	if avagoConfigPath == "" {
```
dasconnor, line 70: Maybe we should make this a prompt instead of just setting via flags.

How would you like update the avalanchego config?
1. Automatic
2. Manual

For manual, we just print instructions to the screen. For automatic, we ask for the path to the avalanchego config

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,138 @@
...
+	} else {
+		if err := editConfigFile(subnetIDstr, networkLower); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func editConfigFile(subnetID string, networkID string) error {
+	warn := "WARNING: This will overwrite your existing config file if there is any, are you sure?"
```
dasconnor, line 81: Can we say edit instead of overwrite?

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,138 @@
...
+		return fmt.Errorf("failed to load avalanchego config file %s: %w", avagoConfigPath, err)
+	}
+	if fileBytes == nil {
+		fileBytes = []byte("{}")
+	}
+	var avagoConfig map[string]interface{}
+	if err := json.Unmarshal(fileBytes, &avagoConfig); err != nil {
+		return fmt.Errorf("failed to unpack the config file %s to JSON: %w", avagoConfigPath, err)
+	}
+	avagoConfig["whitelisted-subnets"] = subnetID
```
dasconnor, line 101: I think we want these to be arrays right? Shouldn't we check for existing values and append so we don't overwrite what's there?

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,138 @@
...
+	avagoConfig["network-id"] = networkID
+
+	writeBytes, err := json.MarshalIndent(avagoConfig, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to pack JSON to bytes for the config file: %w", err)
+	}
+	if err := os.WriteFile(avagoConfigPath, writeBytes, constants.DefaultPerms755); err != nil {
+		return fmt.Errorf("failed to write JSON config file, check permissions? %w", err)
+	}
+	msg := `The config file has been edited. To use it, make sure to start the node with the '--config-file' option, e.g.
```
dasconnor, line 111: We should also say you need to restart your node

dasconnor (commented):
Overall looks good, I think we can fine tune the implementation a little bit.

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,138 @@
...
+(consider correct permissions).`,
+		RunE: joinCmd,
+		Args: cobra.ExactArgs(1),
+	}
+	cmd.Flags().StringVar(&avagoConfigPath, "avalanchego-config", "", "file path of the avalanchego config file")
+	return cmd
+}
+
+func joinCmd(cmd *cobra.Command, args []string) error {
+	/*
```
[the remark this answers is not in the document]
holisticode: Because all that is implemented in the addValidator PR (unfortunate PR setup on my end, that part could probably have been an independent PR). As soon as that is merged, I need to uncomment this, and that's when this PR goes from draft to public.

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,138 @@
...
+		if err != nil {
+			return err
+		}
+
+		subnetID := sc.SubnetID
+		if subnetID == ids.Empty {
+			return errNoSubnetID
+		}
+	*/
+	subnetIDstr := "29Zd5yhP7Yb2cTebBbUVKUjHHNviHzgj1y9kKJsvMn2dyWnkpG"
```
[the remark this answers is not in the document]
holisticode: This is just hardcoded for testing rn and part of the draft issue above. It will go away as soon as we enable the above code.

cmd/subnetcmd/join.go
holisticode: Will do, although I thought it important to tell the user that we are actually indeed overwriting it.

cmd/subnetcmd/join.go
holisticode: Makes sense, but we are only adding to existing subnet IDs _in the config file_ correct? NOT by checking the sidecars etc.

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	if err := editConfigFile(subnetIDstr, networkLower); err != nil {
+		return err
+	}
+	return nil
+}
+
+func checkIsWhitelisted(subnetID ids.ID, network models.Network) (bool, error) {
+	/*
+		promptStr := "Please enter your node's ID (NodeID-...)"
```
holisticode, line 143: This is also just temp commented until we move this out of draft

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,256 @@
...
+	network = models.NetworkFromString(networkStr)
+	networkLower := strings.ToLower(network.String())
+
+	ask := "Would you like to check if your node is already whitelisted to join this subnet?"
+	yes, err := prompts.CaptureYesNo(ask)
+	if err != nil {
+		return err
+	}
+	if yes {
+		isWhitelisted, err := checkIsWhitelisted(ids.GenerateTestID(), network)
```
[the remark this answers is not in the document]
felipemadero: I think you can already pass a subnet id to it, instead of GenerasteTestID

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,256 @@
...
+		if err != nil {
+			return err
+		}
+		if !isWhitelisted {
+			ux.Logger.PrintToUser(`The node seems to not have been whitelisted.
```
felipemadero, line 92: I think the msg is misleading, the node is not a validator maybe?

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,256 @@
...
+			if err != nil {
+				return err
+			}
+			if !y {
+				return nil
+			}
+		}
+	}
+
+	if print {
```
felipemadero, line 105: it would be cleaner to me if you have a booleans on what to do (manual or automatic), fill them it on cmdline flags and user responses, and at the end of the function use it to either printJoinCmd or editConfigFile

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,256 @@
...
+	"github.com/ava-labs/avalanche-cli/pkg/prompts"
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/vms/platformvm"
+	"github.com/spf13/cobra"
+)
+
+var (
+	avagoConfigPath string
+	print           bool
```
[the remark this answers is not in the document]
holisticode: Used `printManual` now

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,256 @@
...
+)
+
+// avalanche subnet deploy
+func newJoinCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "join [subnetName]",
+		Short: "Instruct a validator node to begin validating a new subnet",
```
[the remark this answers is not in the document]
holisticode: This comes from here...: https://ava-labs.atlassian.net/wiki/spaces/AP/pages/2087747585/Command+Tree#Join-Existing-Subnet

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,256 @@
...
+		Long: `Instruct a validator node to begin validating a new subnet.
+The NodeID of that validator node must have been whitelisted by one of the 
+subnet's control keys for this to work.
+
+The node also needs to be restarted.
```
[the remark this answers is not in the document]
holisticode: I added a bit more info without diverging too much from the original spec in Confluence

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,256 @@
...
+If --avalanchego-config is provided, this command tries to edit the config file at that path
+(consider correct permissions).`,
```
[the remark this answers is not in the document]
holisticode: Updated the description.

cmd/subnetcmd/join.go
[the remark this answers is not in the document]
holisticode: Yes, but because the actual subnetID is commented out above, I am just using a test ID until the code can be enabled (it's in the addValidator PR which itself can't be merged yet...)

cmd/subnetcmd/join.go
holisticode: The whole command has been designed/specified to be about whitelisting...I understand where you are coming from, but I think @dasconnor should have a say here?

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	ctx := context.Background()
+	nodeIDs := []ids.NodeID{nodeID}
+
+	pClient := platformvm.NewClient(api)
+	vals, err := pClient.GetCurrentValidators(ctx, subnetID, nodeIDs)
+	if err != nil {
+		return false, err
+	}
+	for _, v := range vals {
```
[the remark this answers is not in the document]
holisticode: That is correct. It's not needed, but I thought to do double-checking. If you insist I will remove.

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,256 @@
...
+		case choiceAutomatic:
+			avagoConfigPath, err = prompts.CaptureString("Path to your existing config file (or where it will be generated)")
+			if err != nil {
+				return err
+			}
+		}
+		// if choice is automatic, we just pass through this block,
+		// so we don't need another else if the the config path is not set
+	}
+	if err := editConfigFile(subnetIDstr, networkLower); err != nil {
```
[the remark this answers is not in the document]
holisticode: It's also used in `printJoincmd`

cmd/subnetcmd/join.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	for _, v := range vals {
+		if v.NodeID == nodeID {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func editConfigFile(subnetID string, networkID string) error {
```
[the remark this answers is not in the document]
holisticode: OK. I thought it's odd to pass around a global...

cmd/subnetcmd/join.go
holisticode: The node is probably a validator on the primary network, but not yet on this subnet. I will change this to "The node is not whitelisted to validate this subnet"

cmd/subnetcmd/join.go
holisticode: I thought of this too. But I decided against this because it should never happen that both are executed. Also, if `print` (now `printManual`) is set, we should return right away, with the boolean case we have to wait until the end. It becomes confusing in a different way. I vote to keep it like this.

cmd/subnetcmd/join.go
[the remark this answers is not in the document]
felipemadero: Will not insist ha

cmd/subnetcmd/join.go
dasconnor: To me, overwrite implies discarding everything that's there already, editing preserves any values that aren't relevant to what's being changed. Right now, we only have the global node config, but in the future we will probably allow other configs that should be preserved.

cmd/subnetcmd/join.go
dasconnor: Yep. I'm just thinking of the case where I'm validating one subnet already and I want to validate a second. Those values need to be merged, the second shouldn't overwrite the first.

cmd/subnetcmd/join.go
```diff
@@ -67,18 +78,101 @@ func joinCmd(cmd *cobra.Command, args []string) error {
...
+func checkIsWhitelisted(subnetID ids.ID, network models.Network) (bool, error) {
+	/*
+		promptStr := "Please enter your node's ID (NodeID-...)"
+		nodeID, err := prompts.CaptureNodeID(promptStr)
+		if err != nil {
+			return false, err		}
+	*/
+	nodeID := ids.GenerateTestNodeID()
+	var api string
+	switch network {
```
dasconnor, line 150: do we want to move this to its own function? I imagine we will need to use this logic a bunch

cmd/subnetcmd/join.go
```diff
@@ -98,7 +192,31 @@ func editConfigFile(subnetID string, networkID string) error {
 	if err := json.Unmarshal(fileBytes, &avagoConfig); err != nil {
 		return fmt.Errorf("failed to unpack the config file %s to JSON: %w", avagoConfigPath, err)
 	}
-	avagoConfig["whitelisted-subnets"] = subnetID
+
+	// check the old entries in the config file for whitelisted subnets
+	oldVal := avagoConfig["whitelisted-subnets"]
+	newVal := ""
+	if oldVal != nil {
+		// if an entry already exists, we check if the subnetID already is part
```
dasconnor, line 200: really nice documentation here

dasconnor (commented):
nice work

cmd/subnetcmd/join.go
holisticode: That is a good idea but I suggest to move this down the line a bit: Many functions are currently very network-dependent, meaning some are fuji only, some local only, and thus what the `switch` is allowed to contain is not (yet?) consistent.

cmd/subnetcmd/join.go
holisticode: Thanks

cmd/subnetcmd/join.go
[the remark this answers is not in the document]
holisticode: Updated to clean code now

cmd/subnetcmd/join.go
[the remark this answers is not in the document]
holisticode: Updated to clean code now

cmd/subnetcmd/join.go
holisticode: done

dasconnor (approved):

PR #89, opened 2022-07-13
[AV-1888] Add global node config and bump Avago version

dasconnor:


cmd/root.go
```diff
@@ -44,6 +46,7 @@ in with avalanche subnet create myNewSubnet.`,
 	// Disable printing the completion command
 	rootCmd.CompletionOptions.HiddenDefaultCmd = true
 
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.avalanche-cli.json)")
```
felipemadero, line 49: why not inside of .avalanche-cli ?

cmd/root.go
```diff
@@ -128,6 +132,30 @@ func setupLogging(baseDir string) (logging.Logger, error) {
...
+	if cfgFile != "" {
+		// Use config file from the flag.
+		viper.SetConfigFile(cfgFile)
+	} else {
+		// Search for default config.
+		home, err := os.UserHomeDir()
+		cobra.CheckErr(err)
+		viper.AddConfigPath(home)
+		viper.SetConfigType("json")
+		viper.SetConfigName(".avalanche-cli")
```
felipemadero, line 146: probably can just also use viper.SetConfigFile with the default config path

pkg/application/app.go
```diff
@@ -180,3 +181,15 @@ func (app *Avalanche) ChainIDExists(chainID string) (bool, error) {
 
 	return false, nil
 }
+
+func (app *Avalanche) LoadNodeConfig() (string, error) {
```
felipemadero, line 185: I would probably make a config object with one field for each needed config item, create it in initConfig, and them pass it to app, which will expose it, eg app.Conf.GlobalNodeConf

pkg/application/app_test.go
```diff
@@ -419,10 +421,59 @@ func Test_genesisExists(t *testing.T) {
...
 		baseDir: tempDir,
 		Log:     logging.NoLog{},
 	}
 }
+
+func useViper(configName string) error {
+	viper.Reset()
+	viper.SetConfigName(configName)
+	viper.SetConfigType("json")
+	viper.AddConfigPath("./../../tests/assets/")
```
felipemadero, line 476: this config path should be constructed with path join

felipemadero: anyway, I think you can generate the files on the fly

cmd/networkcmd/start.go
```diff
@@ -72,6 +72,16 @@ func startNetwork(cmd *cobra.Command, args []string) error {
 		client.WithExecPath(avalancheGoBinPath),
 		client.WithRootDataDir(outputDir),
 	}
+
+	// load global node configs if they exist
+	configStr, err := app.LoadNodeConfig()
```
felipemadero, line 77: I would have preferred something like: app.Config.NodeConf

cmd/root.go
dasconnor: We don't want users modifying that directory

cmd/root.go
dasconnor: moved to constants

pkg/application/app.go
dasconnor: I made the config object as you suggested

pkg/application/app_test.go
dasconnor: I think we should keep these as files. I'm not sure how you're suggesting I use path join. Can you let me know what you think the line should be?

cmd/networkcmd/start.go
dasconnor: fixed

pkg/application/app_test.go
felipemadero: I was suggesting to avoid slash in paths, filepath.join("..", "..", "tests", "assets")

pkg/config/config.go
```diff
@@ -0,0 +1,28 @@
...
+
+package config
+
+import (
+	"encoding/json"
+
+	"github.com/spf13/viper"
+)
+
+type Config struct{}
```
felipemadero, line 12: really I was thinking in having field string NodeConfig inside this struct, fill it up in New, returning algo error there.
avoid the accesor function LoadNodeConfig and the error return for it.

dasconnor: I don't think we want to do that. Viper actually has the capability to refresh configs if the file changes. In the future, we may want to have long running commands. If we do, this approach will allow us to always use the latest values.

PR #93, opened 2022-07-15
[AV-1886] add e2e test for multiple deploy

felipemadero:


dasconnor (commented):
tests look great!

tests/e2e/subnet/suite.go
```diff
@@ -65,10 +65,10 @@ var _ = ginkgo.Describe("[Subnet]", func() {
 
 		deployOutput = commands.DeploySubnetLocally(subnetName)
 		rpcs, err = utils.ParseRPCsFromOutput(deployOutput)
-		if err != nil {
+		if err == nil {
 			fmt.Println(deployOutput)
 		}
-		gomega.Expect(err).Should(gomega.BeNil())
+		gomega.Expect(err).Should(gomega.HaveOccurred())
```
dasconnor, line 71: I don't understand why the pipeline passed before and passed now. Does the error occur or does it not?

dasconnor:
Also, Fabio's PR is to `fujiDeploy`, can we set this to that branch so that we can integrate your changes?

tests/e2e/subnet/suite.go
felipemadero: I changed ParseRPCsFromOutput following a previous PR comment, to return error if subnet was already deployed

dasconnor (approved):
LGTM

PR #102, opened 2022-07-20
Bumped subnet-evm version and added new precompile

### LINKED ISSUE #96: Update Subnet-EVM to latest version and add fee config precompile support
We need to bump the version of our subnet-evm support to v0.2.5 (or later). This will enable us to add support for fee config precompiles. We need to add the new menu option in the create process.

### LINKED ISSUE #141: Fix flaky E2E test
The key list test seems flaky. Maybe print ordering isn't consistent?

https://github.com/ava-labs/avalanche-cli/runs/7542652944?check_suite_focus=true

dasconnor:
Bumps subnet-evm to the latest version in both our go mod and our deployment. I also added an additional precompile to the creation process.

Closes #96 

Also fixed flaky E2E test. Closes #141


pkg/constants/constants.go
```diff
@@ -9,8 +9,8 @@ const (
 
 	BaseDirName = ".avalanche-cli"
 
-	SubnetEVMReleaseVersion   = "v0.2.3"
-	AvalancheGoReleaseVersion = "v1.7.13"
+	SubnetEVMReleaseVersion   = "v0.2.5"
+	AvalancheGoReleaseVersion = "v1.7.14"
```
holisticode, line 13: Should we actually bump this to v1.7.16 already?

pkg/vm/precompiles.go
```diff
@@ -154,12 +174,13 @@ func getPrecompiles(config params.ChainConfig, app *application.Avalanche) (para
 		nativeMint        = "Native Minting"
 		contractAllowList = "Contract deployment whitelist"
 		txAllowList       = "Transaction allow list"
+		feeManager        = "Manage Fee Settings"
```
holisticode, line 177: This list looks a bit odd, there is a mix of all first case uppercase and lowercase

holisticode (commented):
Just a couple of nits, otherwise lgtm

pkg/constants/constants.go
dasconnor: fixed

pkg/vm/precompiles.go
dasconnor: good catch, cleaned up

felipemadero (approved):

PR #138, opened 2022-07-26
Help user install plugins

### LINKED ISSUE #133: Print instructions or automatically move binary to plugin dir
When we do a validator join, we should copy the VM to the plugin dir or print instructions to do so.

dasconnor:
Added the ability to print instructions for installing the VM binary and complemented that with functionality to do it automatically.

Closes #133

### DIFF
--- a/pkg/application/app.go
+++ b/pkg/application/app.go
@@ -68,7 +68,7 @@ func (app *Avalanche) GetKeyDir() string {
 	return filepath.Join(app.baseDir, constants.KeyDir)
 }
 
-func (app *Avalanche) GetDefaultPluginDir() string {
+func (app *Avalanche) GetPluginDir() string {
 	return filepath.Join(app.baseDir, constants.DefaultPluginDir)
 }
 
--- a/pkg/application/app.go
+++ b/pkg/application/app.go
@@ -68,6 +68,10 @@ func (app *Avalanche) GetKeyDir() string {
 	return filepath.Join(app.baseDir, constants.KeyDir)
 }
 
+func (app *Avalanche) GetDefaultPluginDir() string {
+	return filepath.Join(app.baseDir, constants.DefaultPluginDir)
+}
+
 func (app *Avalanche) GetKeyPath(keyName string) string {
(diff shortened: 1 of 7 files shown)


pkg/constants/constants.go
```diff
@@ -53,4 +53,6 @@ const (
 
 	DefaultConfigFileName = ".avalanche-cli"
 	DefaultConfigFileType = "json"
+
+	DefaultPluginDir = "plugins"
```
felipemadero, line 57: really this is not going to be the plugin dir used by avalanchego, probably the name DefaultPluginDir is misleaving

pkg/binutils/binaries.go
```diff
@@ -29,6 +29,7 @@ var (
 
 type PluginBinaryDownloader interface {
 	Download(vmIDs map[string]struct{}, pluginDir, binDir string) error
+	DownloadVM(vmID string, pluginDir, binDir string) error
```
felipemadero, line 32: why not use Download?

pkg/application/app.go
```diff
@@ -68,6 +68,10 @@ func (app *Avalanche) GetKeyDir() string {
 	return filepath.Join(app.baseDir, constants.KeyDir)
 }
 
+func (app *Avalanche) GetPluginDir() string {
```
felipemadero, line 71: GetPluginDir may be misleaving as avalanchego is not going to use this directory

pkg/constants/constants.go
dasconnor: this is a good point, I moved the referenced plugin dir to `/tmp`

pkg/binutils/binaries.go
dasconnor: Download has a lot of additional complexity. Honestly, I had a lot of confusion trying to figure out how to use it. DownloadVM was much simpler and already public, just needed to be added to the interface.

pkg/application/app.go
dasconnor: renamed

felipemadero (approved):
LGTM

PR #139, opened 2022-07-26
Add Subnet Import/Export

### LINKED ISSUE #120: Create a path for users to import an existing fuji subnet
Users will need to be able to recover a deployed Fuji subnet on a new CLI instance. If we do a Fuji deploy and then delete our subnet config, how can we get it back?

dasconnor:
Add commands to export subnet configurations to a file and import them in other CLI instances. With this, users can deploy a subnet with one validator and spin up additional validators on other nodes.

Closes #120

### DIFF
--- a/cmd/subnetcmd/export.go
+++ b/cmd/subnetcmd/export.go
@@ -1,0 +1,69 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package subnetcmd
+
+import (
+	"encoding/json"
+	"os"
+
+	"github.com/ava-labs/avalanche-cli/pkg/application"
+	"github.com/ava-labs/avalanche-cli/pkg/models"
+	"github.com/spf13/cobra"
+)
+
+var exportOutput string
+
+// avalanche subnet list
+func newExportCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "export [subnetName]",
+		Short:        "Export deployment details",
+		Long:         `The subnet export command prints the details of an existing subnet deploy.`,
+		RunE:         exportSubnet,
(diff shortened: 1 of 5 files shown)


cmd/subnetcmd/export.go
```diff
@@ -0,0 +1,72 @@
...
+	exportData := models.Exportable{
+		Sidecar: sc,
+		Genesis: gen,
+	}
+
+	exportBytes, err := json.Marshal(exportData)
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(exportOutput, exportBytes, application.WriteReadReadPerms)
```
felipemadero, line 71: do we want to have an app location for subnet export files? (and so have a app.LoadSubnet)

cmd/subnetcmd/export.go
```diff
@@ -0,0 +1,72 @@
...
+	)
+
+	return cmd
+}
+
+func exportSubnet(cmd *cobra.Command, args []string) error {
+	var err error
+	if exportOutput == "" {
+		pathPrompt := "Enter file path to write export data to"
+		exportOutput, err = app.Prompt.CaptureString(pathPrompt)
```
felipemadero, line 45: do we need to check for valid path?

cmd/subnetcmd/export.go
dasconnor: I don't think so, we want it to be user provided, not in any of our app directories.

cmd/subnetcmd/export.go
dasconnor: I think this is probably fine, if the user messes up, they can re-enter easily. If we want to in the future, we can add a function that checks the permissions of the directory, but I think that's a little much for right now. I'd want to test that more than we have time for right now.

felipemadero (approved):

PR #150, opened 2022-07-27
Sanitize Paths and Fix Genesis Filepath Issue

### LINKED ISSUE #149: Allow tilde in join paths

### LINKED ISSUE #151: Bug: Passing genesis filename instead of genesis bytes
H/T to @felipemadero for finding the issue.

dasconnor:
Sanitize paths to allow for using `~/` in your path.

Also fixes an issue where we are passing the genesis filepath in the create blockchain tx instead of the actual genesis bytes.

Closes #149 Closes #151


felipemadero (approved):
LGTM

PR #153, opened 2022-07-28
Added custom VM create flow

### LINKED ISSUE #154: Add Custom VM to create flow

dasconnor:
I refactored our `subnet create` flow to accommodate custom VMs. Custom VMs accept a genesis and copy the VM binary to our `~/.avalanche-cli/bin/vms` directory.

Closes #154

### DIFF
--- a/pkg/application/app.go
+++ b/pkg/application/app.go
@@ -5,6 +5,7 @@ package application
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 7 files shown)


pkg/application/app.go
```diff
@@ -141,7 +159,7 @@ func (app *Avalanche) CreateSidecar(sc *models.Sidecar) error {
 	// but better safe than sorry
 	exists, err := app.ChainIDExists(sc.ChainID)
 	if err != nil {
-		return err
+		return fmt.Errorf("unable to determine if chainID is unique: %s", err)
```
felipemadero, line 162: it is best practice to use %w to wrap errors, instead of %s, when using Errorf

dasconnor: good catch, fixed

felipemadero (approved):

PR #170, opened 2022-07-30
Prettify json output for subnet genesis description

lieferbrando:
Prettifies the JSON output of the `subnet describe --genesis <subnetName>` command. 
- Detects whether the output terminal has a light or dark color-profile to adjust syntax highlighting accordingly
- Detects whether the output is written to a file (e.g. `> output.json`) and omits styling in that case to not write ansi escapes into the file

`PrintPrettyCode` works on a variety of languages, see [here](https://github.com/alecthomas/chroma#supported-languages) for an extensive list.

Here's an example for light/dark styling in action:
Screenshot 2022-07-30 at 16 43 44


holisticode:
Thanks for your submission! Looks nice, would you still consider providing a `--no-colors` flag if other users maybe prefer?

Issue #121, opened 2022-07-25
Add --fuji deploy flag


dasconnor:
The `avalanche subnet deploy` command should accept a `--fuji` flag to skip the select network prompt. This is a mutually exclusive flag with `--local`.

PR #179, opened 2022-08-08
Add flag for fuji

holisticode:
Closes https://github.com/ava-labs/avalanche-cli/issues/121

### DIFF
--- a/cmd/subnetcmd/deploy.go
+++ b/cmd/subnetcmd/deploy.go
@@ -19,8 +19,12 @@ import (
 )
 
 var (
-	deployNetwork string
+	deployLocal   bool
+	deployTestnet bool
+	deployMainnet bool
 	keyName       string
+
+	errMutuallyExlusive = errors.New("--local, --fuji (resp. --testnet) and --mainnet are mutually exclusive")
 )
 
 // avalanche subnet deploy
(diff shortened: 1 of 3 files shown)


cmd/subnetcmd/deploy.go
```diff
@@ -90,8 +90,15 @@ func deploySubnet(cmd *cobra.Command, args []string) error {
 
 	// get the network to deploy to
 	var network models.Network
-	if deployLocal {
-		network = models.Local
+	if deployNetwork != "" {
+		switch deployNetwork {
+		case strings.ToLower(models.Fuji.String()):
+			network = models.Fuji
+		case "local":
```
felipemadero, line 97: there is also a modes.Local.String()

holisticode: Yes but that `String()` function prints `Local Network`. I think we just want people to enter `local`.

cmd/subnetcmd/deploy.go
```diff
@@ -46,7 +46,7 @@ subnet and deploy it on Fuji or Mainnet.`,
 		RunE:         deploySubnet,
 		Args:         cobra.ExactArgs(1),
 	}
-	cmd.Flags().BoolVarP(&deployLocal, "local", "l", false, "deploy to a local network")
+	cmd.Flags().StringVarP(&deployNetwork, "network", "n", "", "network to which to deploy to (one of 'local' or 'fuji')")
```
dasconnor, line 49: Oh interesting. I was actually thinking we would do `--local`, `--fuji`, and `--mainnet` and then just check that only one flag was selected. We need to use this pattern a few places, so we should make a global decision. Does anyone have a strong opinion? @holisticode @felipemadero

cmd/subnetcmd/deploy.go
dasconnor: If we take the approach I mentioned above, we can synchronize this with the model string, which I think would be preferable.

cmd/subnetcmd/deploy.go
holisticode: I am suggesting here to use the `--network` flag so I stand to that suggestion. However, no strong opinion.

holisticode: Ok refactored to use dedicated flags

cmd/subnetcmd/deploy.go
holisticode: With bool flags it's not even needed

cmd/subnetcmd/deploy.go
felipemadero: No strong opinion

felipemadero (commented):
LGTM

dasconnor (approved):
LGTM

Issue #129, opened 2022-07-26
Add list to control key capture flow


dasconnor:
The control key flow should have list and delete functions, same as precompile configuration.

PR #180, opened 2022-08-09
List and delete control keys

holisticode:
Closes https://github.com/ava-labs/avalanche-cli/issues/129 

Also regenerated all mocks (maybe due to go1.18 and/or mockery update it was required for all interfaces)


cmd/subnetcmd/deploy.go
```diff
@@ -258,6 +260,25 @@ func controlKeysLoop(controlKeysPrompt string, network models.Network) ([]string
...
+			}
+			index, err := app.Prompt.CaptureIndex("Choose key to remove:", ifaceK)
+			if err != nil {
+				return nil, false, err
+			}
+			controlKeys = append(controlKeys[:index], controlKeys[index+1:]...)
+		case preview:
+			fmt.Println("Control keys:")
+			for i, k := range controlKeys {
+				fmt.Printf("%d. %s\n", i, k)
```
felipemadero, line 280: probably use ux here

cmd/subnetcmd/deploy.go
```diff
@@ -258,6 +260,25 @@ func controlKeysLoop(controlKeysPrompt string, network models.Network) ([]string
...
+			for i, v := range controlKeys {
+				ifaceK[i] = v
+			}
+			index, err := app.Prompt.CaptureIndex("Choose key to remove:", ifaceK)
+			if err != nil {
+				return nil, false, err
+			}
+			controlKeys = append(controlKeys[:index], controlKeys[index+1:]...)
+		case preview:
+			fmt.Println("Control keys:")
```
felipemadero, line 278: probably use ux here

cmd/subnetcmd/deploy.go
```diff
@@ -258,6 +260,25 @@ func controlKeysLoop(controlKeysPrompt string, network models.Network) ([]string
 				continue
 			}
 			controlKeys = append(controlKeys, controlKey)
+		case delCtrlKey:
+			if len(controlKeys) == 0 {
+				fmt.Println("no keys added yet")
```
felipemadero, line 265: probably use ux here

pkg/vm/precompiles.go
```diff
@@ -57,7 +57,15 @@ func getAdminList(initialPrompt string, info string, app *application.Avalanche)
 			}
 			admins = append(admins, adminAddr)
 		case removeAdmin:
-			index, err := app.Prompt.CaptureIndex("Choose address to remove:", admins)
+			if len(admins) == 0 {
+				fmt.Println("no admins added yet")
```
felipemadero, line 61: probably use ux here

cmd/subnetcmd/deploy.go
```diff
@@ -258,6 +260,25 @@ func controlKeysLoop(controlKeysPrompt string, network models.Network) ([]string
 				continue
 			}
 			controlKeys = append(controlKeys, controlKey)
+		case delCtrlKey:
```
dasconnor, line 263: Could you modify the deploy function to just use the same code you modified in `precompiles.go`? You might need to accept an extra arg or two, but that way there would only be one code path.

cmd/subnetcmd/deploy.go
holisticode: I actually disagree, ux prints both to log and stdout. Logs such as the above will make no sense in the log as there is no context, so I think it is ok to just print to stdout directly as it is.

Applies to all the other occurrences as well.

cmd/subnetcmd/deploy.go
holisticode: I just pushed such an attempt.

pkg/prompts/prompts.go
```diff
@@ -23,23 +23,39 @@ import (
...
 	CapturePositiveBigInt(promptStr string) (*big.Int, error)
-	CaptureAddress(promptStr string) (common.Address, error)
+	CaptureAddress(promptStr string, arg any) (any, error)
 	CaptureExistingFilepath(promptStr string) (string, error)
 	CaptureYesNo(promptStr string) (bool, error)
 	CaptureNoYes(promptStr string) (bool, error)
 	CaptureList(promptStr string, options []string) (string, error)
 	CaptureString(promptStr string) (string, error)
-	CaptureIndex(promptStr string, options []common.Address) (int, error)
+	CaptureIndex(promptStr string, options []interface{}) (int, error)
```
felipemadero, line 43: probably can use `any` here also

pkg/prompts/prompts.go
```diff
@@ -23,23 +23,39 @@ import (
...
 	CaptureDuration(promptStr string) (time.Duration, error)
 	CaptureDate(promptStr string) (time.Time, error)
 	CaptureNodeID(promptStr string) (ids.NodeID, error)
 	CaptureWeight(promptStr string) (uint64, error)
 	CaptureUint64(promptStr string) (uint64, error)
-	CapturePChainAddress(promptStr string, network models.Network) (string, error)
+	CapturePChainAddress(promptStr string, network any) (any, error)
+	CaptureListDecision(
```
felipemadero, line 50: more descriptive names for params, and comments

pkg/prompts/prompts.go
```diff
@@ -127,6 +143,65 @@ func validateBiggerThanZero(input string) error {
 	return nil
 }
 
+func (r *realPrompter) CaptureListDecision(
+	prompter Prompter, // we need this in order to be able to run mock tests
```
felipemadero, line 147: same, need more descriptive names and comments

felipemadero, line 147: think the prompter can be avoided

pkg/prompts/prompts.go
```diff
@@ -127,6 +143,65 @@ func validateBiggerThanZero(input string) error {
...
+	prompt string,
+	capture func(prompt string, args any) (any, error),
+	capturePrompt string,
+	label string,
+	info string,
+	arg any,
```
felipemadero, line 153: shall avoid the arg, and instead give a capture function with the param already included

pkg/prompts/prompts.go
```diff
@@ -127,6 +143,65 @@ func validateBiggerThanZero(input string) error {
...
+) ([]any, bool, error) {
```
felipemadero, line 154: maybe... this is a great opportunity to use go generics, and so avoid the extra for loop after calling this function to convert types

pkg/prompts/prompts.go
```diff
@@ -380,3 +456,12 @@ func (*realPrompter) CaptureIndex(promptStr string, options []common.Address) (i
 	}
 	return listIndex, nil
 }
+
+func contains(list []any, element any) bool {
```
felipemadero, line 460: great opportunity to use go generics?

cmd/subnetcmd/deploy.go
```diff
@@ -228,44 +228,32 @@ func getControlKeys(network models.Network) ([]string, bool, error) {
...
-		case cancelMsg:
-			return nil, true, nil
-		default:
-			return nil, false, errors.New("unexpected option")
+	ctrlKeys := make([]string, len(list))
+	var (
+		key string
+		ok  bool
+	)
+	for i, k := range list {
```
felipemadero, line 250: probably with go generics we can avoid this loop

pkg/prompts/prompts.go
```diff
@@ -286,7 +362,7 @@ func (*realPrompter) CapturePChainAddress(promptStr string, network models.Netwo
 	return prompt.Run()
 }
 
-func (*realPrompter) CaptureAddress(promptStr string) (common.Address, error) {
+func (*realPrompter) CaptureAddress(promptStr string, arg any) (any, error) {
```
felipemadero, line 365: the arg is not needed, should not be included

pkg/prompts/prompts.go
```diff
@@ -23,23 +23,39 @@ import (
...
+	Preview  = "Preview"
+	MoreInfo = "More Info"
+	Done     = "Done"
+	Cancel   = "Cancel"
 )
 
 type Prompter interface {
 	CapturePositiveBigInt(promptStr string) (*big.Int, error)
-	CaptureAddress(promptStr string) (common.Address, error)
+	CaptureAddress(promptStr string, arg any) (any, error)
```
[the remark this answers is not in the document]
felipemadero: anyway, this can be avoided if `CaptureListDecision` where generic

pkg/prompts/prompts.go
```diff
@@ -23,23 +23,39 @@ import (
...
 	CaptureString(promptStr string) (string, error)
-	CaptureIndex(promptStr string, options []common.Address) (int, error)
+	CaptureIndex(promptStr string, options []interface{}) (int, error)
 	CaptureDuration(promptStr string) (time.Duration, error)
 	CaptureDate(promptStr string) (time.Time, error)
 	CaptureNodeID(promptStr string) (ids.NodeID, error)
 	CaptureWeight(promptStr string) (uint64, error)
 	CaptureUint64(promptStr string) (uint64, error)
-	CapturePChainAddress(promptStr string, network models.Network) (string, error)
+	CapturePChainAddress(promptStr string, network any) (any, error)
```
felipemadero, line 49: the any can be avoided if `CaptureListDecision` is made generic

pkg/prompts/prompts.go
felipemadero: probably can be generic

pkg/vm/precompiles.go
```diff
@@ -15,69 +15,32 @@ import (
...
-		addAdmin    = "Add admin"
-		removeAdmin = "Remove admin"
-		preview     = "Preview"
-		moreInfo    = "More info"
-		doneMsg     = "Done"
-		cancelMsg   = "Cancel"
+	label := "Address"
+
+	list, canceled, err := app.Prompt.CaptureListDecision(
+		app.Prompt,
```
felipemadero, line 22: improve params description, or naming

cmd/subnetcmd/deploy.go
```diff
@@ -228,44 +228,32 @@ func getControlKeys(network models.Network) ([]string, bool, error) {
...
-	const (
-		addCtrlKey = "Add control key"
-		doneMsg    = "Done"
-		cancelMsg  = "Cancel"
+	label := "Control key"
+	info := "Control keys are P-Chain addresses which have admin rights on the subnet.\n" +
+		"Only private keys which control such addresses are allowed to make changes on the subnet"
+	arg := network
+	list, canceled, err := app.Prompt.CaptureListDecision(
+		app.Prompt, // we need this to be able to mock test
```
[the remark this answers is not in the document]
holisticode: The unit test exists in this PR and is at `pkg/prompts/capturetests/list_decision_test.go`. The intent of the parameter is if we use the mock, then the test is useless. We want the whole flow to actually take place, not just fake a return from the mocker, which then would not actually run the code. So we use a `realPrompter` and then pass the `mockPrompter` as an argument here in order to test.

cmd/subnetcmd/deploy.go
```diff
@@ -228,44 +228,32 @@ func getControlKeys(network models.Network) ([]string, bool, error) {
...
+		controlKeysPrompt,
+		app.Prompt.CapturePChainAddress,
+		"Enter P-Chain address (Example: P-...)",
```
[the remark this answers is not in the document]
holisticode: Well ok as I am using variables for the other stuff it makes sense

cmd/subnetcmd/deploy.go
```diff
@@ -228,44 +228,32 @@ func getControlKeys(network models.Network) ([]string, bool, error) {
...
-		addCtrlKey = "Add control key"
-		doneMsg    = "Done"
-		cancelMsg  = "Cancel"
+	label := "Control key"
+	info := "Control keys are P-Chain addresses which have admin rights on the subnet.\n" +
+		"Only private keys which control such addresses are allowed to make changes on the subnet"
+	arg := network
+	list, canceled, err := app.Prompt.CaptureListDecision(
+		app.Prompt, // we need this to be able to mock test
+		controlKeysPrompt,
```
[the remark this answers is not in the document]
holisticode: added comments

cmd/subnetcmd/deploy.go
```diff
@@ -228,44 +228,32 @@ func getControlKeys(network models.Network) ([]string, bool, error) {
...
+		app.Prompt.CapturePChainAddress,
+		"Enter P-Chain address (Example: P-...)",
+		label,
+		info,
+		arg,
```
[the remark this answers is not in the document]
holisticode: That doesn't work as far as I can see as we want a common interface for the prompter function. `CapturePChainAddress` needs a `network` param, `CaptureAddress` doesn't, therefore `arg` is just an optional parameter, just as `args` in `ux.Logger.PrintToUser` or `fmt.Printf`

pkg/prompts/capturetests/list_decision_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/stretchr/testify/mock"
+)
+
+/*
+	These tests need to be in an independent module
+	because otherwise they create a circular dependency
+	between the mocks and the prompts modules
+*/
+
+func TestListDecision(t *testing.T) {
```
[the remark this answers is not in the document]
holisticode: This _IS_ testing `CaptureListDecision`

pkg/prompts/prompts.go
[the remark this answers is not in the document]
holisticode: I don't think that is needed here. This works perfectly.

pkg/prompts/prompts.go
holisticode: `any` is just an alias to `interface{}` but yes, I will change it here because I used `any` elsewhere. Why should I use generics, this works perfectly. Using generics is not necessarily better.

pkg/prompts/prompts.go
holisticode: Added comments just as the other place.

pkg/prompts/prompts.go
holisticode: Comments yes. Done. The comments make clear why the names are such.

holisticode: Comments yes. Done. The comments make clear why the names are such.

holisticode: I am not sure it can though. See, the implementation is on `realPrompter` but to mock it we need to pass the mock for the other prompts required in this function.

pkg/prompts/prompts.go
holisticode: I am not sure in this design the `arg` can be avoided. It is _not_ representing the capture function, but an **optional** argument for the actual capture function. It can not be the capture function because the two capture functions supported do not share the same signature (`CapturePChainAddress` has an extra arg `network`, which is passed through this `arg` here).

pkg/prompts/prompts.go
holisticode: I am afraid I don't understand this suggestion. In any case, generics help where the _signature_ is the same but the _types_ are different. This is exactly not the case here. Also, it took me a lot of time to get to this design, and trying to change it to generics will take more time again, which I don't think is necessary.

pkg/prompts/prompts.go
holisticode: Already commented.

cmd/subnetcmd/deploy.go
holisticode: Already commented.

pkg/prompts/prompts.go
holisticode: I think  it is, as already commented above, because the signatures are different. Like this, it becomes an optional param.

pkg/prompts/prompts.go
holisticode: Already commented.

pkg/vm/precompiles.go
holisticode: Done 3 times in interface spec, in implementation and first use. I think that's sufficient.

cmd/subnetcmd/deploy.go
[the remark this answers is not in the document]
felipemadero: I should insist on `CaptureListDecision` taking param `capture func(prompt string) (any, error)` and removing the `arg any` param.
So for `app.Prompt.CapturePChainAddress`, you use a lambda function in `deploy.go`.
And for `app.Prompt.CaptureAddress`, you avoid the useless arg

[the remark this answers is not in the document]
felipemadero: Reason for that, is that there is no good reason for CaptureListDecision to do that work of passing around args.

pkg/prompts/prompts.go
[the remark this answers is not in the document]
felipemadero: You can keep the expected type for the function logic, `common.Address`, by using generics

pkg/prompts/prompts.go
felipemadero: Using generics you keep the most appropiate type for the domain of each function, and avoid boilerplate code to copy data.

pkg/prompts/prompts.go
felipemadero: As I mentioned somewhere else, I have preference for avoiding this extra work of passing around this arg to the capture function, forcing some signatures.

pkg/prompts/prompts.go
felipemadero: I don't believe it shall take much time. Here you have the same interface, and will call it with different captured types.

pkg/prompts/prompts.go
felipemadero: this is another function, really contains will be much generic in that way

pkg/prompts/prompts.go
felipemadero: You can keep the expected type for the function logic, models.Network

dasconnor (commented):
I think the code here looks pretty good, I'd be fine merging as is and moving to generics in a later ticket if we want

holisticode:
That would be my suggestion too. @felipemadero has already done some work with generics in https://github.com/ava-labs/avalanche-cli/pull/202 so we can bring that in in the next release.

felipemadero (approved):

Issue #130, opened 2022-07-26
Add range print text for number of control keys


dasconnor:
When setting the number of control keys in `deploy`, we should add range parameters in the prompt.

PR #181, opened 2022-08-09
Select threshold from list

holisticode:
Closes https://github.com/ava-labs/avalanche-cli/issues/130

Instead of "When setting the number of control keys in deploy, we should add range parameters in the prompt", in this PR I am suggesting just showing a list to pick the threshold from.

### DIFF
--- a/cmd/subnetcmd/deploy.go
+++ b/cmd/subnetcmd/deploy.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/ava-labs/avalanche-cli/pkg/binutils"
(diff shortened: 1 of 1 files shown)


cmd/subnetcmd/deploy.go
```diff
@@ -171,7 +172,7 @@ func deploySubnet(cmd *cobra.Command, args []string) error {
 	var threshold uint32
 
 	if len(controlKeys) > 0 {
-		threshold, err = getThreshold(uint64(len(controlKeys)))
+		threshold, err = getThreshold(len(controlKeys))
```
felipemadero, line 175: I am not completelly sure that len(controlKeys) should be the max. I think a max of somewhat 3 was refered to in some meeting.

holisticode: I don't know of any such restriction, can you find the source please?

dasconnor (approved):

dasconnor:
I think this is good for now. Later we might want to do a branching path. If there are a lot of control keys, it might break the interface. If `numControlKeys > 10` or something like that, we might want to capture via text rather than list.

cmd/subnetcmd/deploy.go
holisticode: Resolved as per https://github.com/ava-labs/avalanche-cli/pull/181#issuecomment-1216984871

Issue #128, opened 2022-07-26
Use test key as control key


dasconnor:
In the Fuji deploy process, we should be able to use our generated keys as control keys.

PR #184, opened 2022-08-09
Use generated keys as control keys

holisticode:
Closes https://github.com/ava-labs/avalanche-cli/issues/128

* Preselects all P-Chain addresses (usually just one) of the selected deploy key as control key 
* Allows to alternatively pick from other local key's P-Chain addresses
* Finally also allows alternatively to enter custom P-Chain addresses

Do not merge before https://github.com/ava-labs/avalanche-cli/pull/180 is merged as it changes the same `CaptureIndex` prompt func, while the other PR also adds the required new mocks. This PR will fail unit tests until https://github.com/ava-labs/avalanche-cli/pull/180 is merged to main and into this PR.


cmd/subnetcmd/deploy.go
```diff
@@ -256,13 +258,131 @@ func deploySubnet(cmd *cobra.Command, args []string) error {
...
 }
 
-func getControlKeys(network models.Network) ([]string, bool, error) {
+func getControlKeys(network models.Network, keyName string) ([]string, bool, error) {
 	controlKeysInitialPrompt := "Configure which addresses may add new validators to the subnet.\n" +
 		"These addresses are known as your control keys. You will also\n" +
-		"set how many control keys are required to add a validator."
-	controlKeysPrompt := "Set control keys"
+		"set how many control keys are required to add a validator (the threshold)."
+	moreKeysPrompt := "P-Chain addresses from your private key will be added by default to the control key set. Would you like to add more?"
```
dasconnor, line 265: I think we should change this prompt. Instead of saying "We're adding all keys, add more? No/Yes" we should do "How would you like to set your control keys?" 1. Use all stored keys 2. Use creation key only 3. Custom list (that does not automatically include the stored keys)

dasconnor (commented):
Code looks great. I think we should reconsider one of the prompts

cmd/subnetcmd/deploy.go
holisticode: Updated according to this request.

dasconnor (approved):
LGTM

PR #187, opened 2022-08-10
Run network clean --hard after install

### LINKED ISSUE #177: Provide a clean update path for binaries
When a user updates the `avalanche-cli` binary via install script, only the `avalanche` binary itself will be updated, but not other required binaries (avalanchego and vms) https://github.com/ava-labs/avalanche-cli/issues/176.

We need to provide a clean update path.

One suggestion could be to optionally prompt the user to run `avalanche clean --hard` after installation.

holisticode:
Closes #177 

This is only a suggestion for how to approach the problem. Other and better solutions may be possible.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -33,6 +33,32 @@ parse_args() {
   shift $((OPTIND - 1))
   TAG=$1
 }
+
+clean() {
+  # this would prob fail on windows
+  # only ask to delete if the bin folder already exists
+  # which means this is probably an update and not a first install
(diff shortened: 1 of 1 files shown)


dasconnor:
I'm not sure we need this. I think we should just add it in the release notes and announce on Twitter. As soon as package management PR goes in, this won't be a problem going forward.

holisticode:
According to @dasconnor this will be addressed by https://github.com/ava-labs/avalanche-cli/pull/173 and won't be a problem anymore

PR #188, opened 2022-08-10
Add flags so that no prompting is possible

holisticode:


cmd/subnetcmd/join.go
```diff
@@ -58,7 +64,10 @@ This command currently only supports subnets deployed on the Fuji testnet.`,
 	}
 	cmd.Flags().StringVar(&avagoConfigPath, "avalanchego-config", "", "file path of the avalanchego config file")
 	cmd.Flags().StringVar(&pluginDir, "plugin-dir", "", "file path of avalanchego's plugin directory")
+	cmd.Flags().StringVar(&networkStr, "network", "", "avalanche network where the validator will be joining (either `fuji` or `mainnet`)")
```
dasconnor, line 67: Same global decision I mentioned in the other PR

cmd/subnetcmd/join.go
```diff
@@ -133,7 +153,10 @@ but until the node is whitelisted, it will not be able to validate this subnet.`
 		return nil
 	}
 
+	// if **both** flags were set, nothing special needs to be done
```
dasconnor, line 156: oof, these comments are showing me how confusing this is. We don't need to fix now, but we probably want to think about refactoring in the future to make simpler/more clear.

dasconnor (commented):
I think I can approve this after we finalize the decision around network flags

cmd/subnetcmd/join.go
holisticode: Fixed.

dasconnor (approved):

Issue #126, opened 2022-07-26
Add Balance to key list


dasconnor:
The table generated in `key list` should have the balances of the addresses.

PR #191, opened 2022-08-16
add balance to key list

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/126

### DIFF
--- a/cmd/keycmd/list.go
+++ b/cmd/keycmd/list.go
@@ -91,26 +91,26 @@ func printAddresses(keyPaths []string) error {
 	}
 
 	for _, keyPath := range keyPaths {
-		cAdded := false
 		keyName := strings.TrimSuffix(filepath.Base(keyPath), constants.KeySuffix)
 		for net, id := range supportedNetworks {
 			sk, err := key.LoadSoft(id, keyPath)
 			if err != nil {
 				return err
 			}
-			if !cAdded {
-				strC := sk.C()
-				balanceStr, err := getCChainBalanceStr(ctx, fujiCClient, strC)
+
+			strC := sk.C()
+			balanceStr := ""
+			if net == models.Fuji.String() {
+				balanceStr, err = getCChainBalanceStr(ctx, fujiCClient, strC)
 				if err != nil {
 					return err
 				}
-				table.Append([]string{keyName, "C-Chain (Ethereum hex format)", strC, balanceStr, "All"})
 			}
-			cAdded = true
+			table.Append([]string{keyName, "C-Chain (Ethereum hex format)", strC, balanceStr, net})
 
 			strP := sk.P()
 			for _, p := range strP {
-				balanceStr := "0"
+				balanceStr := ""
 				if net == models.Fuji.String() {
 					var err error
 					balanceStr, err = getPChainBalanceStr(ctx, fujiPClient, p)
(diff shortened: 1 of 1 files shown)


cmd/keycmd/list.go
```diff
@@ -68,17 +77,39 @@ func printAddresses(keyPaths []string) error {
 			}
 			if !cAdded {
 				strC := sk.C()
-				table.Append([]string{keyName, "C-Chain (Ethereum hex format)", strC, "All"})
+				table.Append([]string{keyName, "C-Chain (Ethereum hex format)", strC, "0", "All"})
```
dasconnor, line 80: we should check the c-chain balance as well and not just mark it as zero

cmd/keycmd/list.go
```diff
@@ -68,17 +77,39 @@ func printAddresses(keyPaths []string) error {
...
+	if err != nil {
+		return "", err
+	}
+	ctx, cancel := context.WithTimeout(ctx, constants.RequestTimeout)
+	resp, err := pClient.GetBalance(ctx, []ids.ShortID{pID})
+	cancel()
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%d", uint64(resp.Balance)), nil
```
dasconnor, line 114: Do we want to display in whole avax units instead of wei?

dasconnor (commented):
Can you add a screenshot?

felipemadero:
> Can you add a screenshot?

scrKeyList

holisticode:
The screenshot is looking somewhat odd - unless I am misunderstanding something: It's unlikely that an address has the same balance on "All" networks.

cmd/keycmd/list.go
```diff
@@ -83,17 +100,61 @@ func printAddresses(keyPaths []string) error {
...
+	ctx, cancel := context.WithTimeout(ctx, constants.RequestTimeout)
+	balance, err := cClient.BalanceAt(ctx, addr, nil)
+	cancel()
+	if err != nil {
+		return "", err
+	}
+	if balance.Cmp(big.NewInt(0)) == 0 {
+		return "0", nil
+	}
+	balance = balance.Div(balance, big.NewInt(int64(units.Avax)))
```
holisticode, line 141: Nit: Could do this just in one calculation (possibly just uint64)

cmd/keycmd/list.go
```diff
@@ -83,17 +100,61 @@ func printAddresses(keyPaths []string) error {
 			}
 			if !cAdded {
 				strC := sk.C()
-				table.Append([]string{keyName, "C-Chain (Ethereum hex format)", strC, "All"})
+				balanceStr, err := getCChainBalanceStr(ctx, fujiCClient, strC)
```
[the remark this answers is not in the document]
felipemadero: It is doing that for the first supported network. Fixing to do it for all networks except local.

cmd/keycmd/list.go
```diff
@@ -83,17 +100,61 @@ func printAddresses(keyPaths []string) error {
...
 			}
 			cAdded = true
 
 			strP := sk.P()
 			for _, p := range strP {
-				table.Append([]string{keyName, "P-Chain (Bech32 format)", p, net})
+				balanceStr := "0"
+				if net == models.Fuji.String() {
+					var err error
+					balanceStr, err = getPChainBalanceStr(ctx, fujiPClient, p)
```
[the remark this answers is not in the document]
felipemadero: This is fuji only.

cmd/keycmd/list.go
felipemadero: I'm not sure to understand the suggestion. Balances got from c-chain have an extra units.Avax multiplier. I split the division into this first extra c-chain multiplier part (to obtain nAvax), ignoring the decimal part.
Afterthat I convert nAvax to avax, keeping record of the multiplier.

holisticode: It's really just a nit. We could just divide only once instead of twice. On the other hand, this makes it clearer to understand why. So up to you,

holisticode (approved):

Issue #164, opened 2022-07-29
Add E2E Test for creating and deleting a custom subnet


dasconnor:
We need to make sure the custom VM is deleted.

PR #193, opened 2022-08-16
164 custom subnet create delete e2e

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/164


dasconnor (commented):
This looks good. We can approve this as is, but should we also add a test that deploys the custom subnet?

Issue #159, opened 2022-07-28
Try to avoid attempting blockchain creation if the control key has no permission


holisticode:
Currently on a `subnet deploy` for fuji, if the control key has no permissions on the subnet, the subnet will actually be created, but the blockchain creation fails.

Try to  recover from this error state or prevent it in the first place.

PR #194, opened 2022-08-16
add check for wallet address having rights to create fuji blockchain

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/159

### DIFF
--- a/pkg/subnet/public.go
+++ b/pkg/subnet/public.go
@@ -139,6 +139,10 @@ func (d *PublicDeployer) createBlockchainTx(chainName string, vmID, subnetID ids
 }
 
 func (d *PublicDeployer) createSubnetTx(controlKeys []string, threshold uint32, wallet primary.Wallet) (ids.ID, error) {
+	err := d.validateWalletIsSubnetOwner(controlKeys, threshold)
+	if err != nil {
+		return ids.Empty, err
+	}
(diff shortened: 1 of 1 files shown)


pkg/subnet/public.go
```diff
@@ -151,3 +155,40 @@ func (d *PublicDeployer) createSubnetTx(controlKeys []string, threshold uint32,
...
+	switch d.network {
+	case models.Fuji:
+		networkID = avago_constants.FujiID
+	case models.Mainnet:
+		networkID = avago_constants.MainnetID
+	default:
+		return fmt.Errorf("unsupported public network")
+	}
+
+	sf, err := key.LoadSoft(networkID, d.privKeyPath)
```
holisticode, line 171: This should only be applied for `fuji` (on mainnet the subnet would have most probably had been created with ledger anyway)

felipemadero: wallet is currently covering both fuji and mainnet, but should be fuji only. probably fix all of this for mainnet deploy.

dasconnor (approved):

pkg/subnet/public.go
holisticode: It's a bit odd though that just a couple of lines above we are supporting Mainnet. If you don't want to change this, then we should maybe remove the Mainnet case above.

holisticode (approved):

PR #202, opened 2022-08-19
129 list del ctrl key fm

felipemadero:


pkg/prompts/prompts.go
```diff
@@ -155,23 +139,21 @@ func validateBiggerThanZero(input string) error {
 // and `CaptureAddress` is supported) until the user cancels or
 // chooses `Done`. It does also offer an optional `info` to print
 // (if provided) and a preview. Items can also be removed.
-func (r *realPrompter) CaptureListDecision(
+func CaptureListDecision[T comparable](
```
dasconnor, line 142: Why did you take this out of the interface?

dasconnor (approved):
I think this is good. Just one minor question. Can we also add a ticket for this?

pkg/prompts/prompts.go
felipemadero: https://github.com/ava-labs/avalanche-cli/issues/258

Issue #198, opened 2022-08-18
Run E2E tests on Mac


dasconnor:
Edit the CI to support tests running in both envs.

PR #213, opened 2022-08-23
add macos to unit tests and e2e

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/198

In relation of debugging some issues here:
- Made some cleanup of fmt.Errorf formats
- Improved output for E2E failed commands
- Add github token usage to avoid rate limit problems at CI

### DIFF
--- a/pkg/binutils/latest.go
+++ b/pkg/binutils/latest.go
@@ -25,10 +25,8 @@ func GetLatestReleaseVersion(releaseURL string) (string, error) {
 		return "", fmt.Errorf("failed to get latest version from %s: %w", releaseURL, err)
 	}
 	if resp.StatusCode != http.StatusOK {
-		fmt.Println(resp.Header)
 		return "", fmt.Errorf("failed to get latest version from %s: unexpected http status code: %d", releaseURL, resp.StatusCode)
 	}
-	fmt.Println(resp.Header)
 	defer resp.Body.Close()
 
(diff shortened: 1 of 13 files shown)


pkg/binutils/latest.go
```diff
@@ -20,7 +21,16 @@ import (
 func GetLatestReleaseVersion(releaseURL string) (string, error) {
 	// TODO: Question if there is a less error prone (= simpler) way to install latest avalanchego
 	// Maybe the binary package manager should also allow the actual avalanchego binary for download
-	resp, err := http.Get(releaseURL)
+	request, err := http.NewRequest("GET", releaseURL, nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to create request for latest version from %s: %w", releaseURL, err)
+	}
+	token := os.Getenv(constants.GithubAPITokenEnvVarName)
```
holisticode, line 28: 1. I guess you already thought of this, but is there no way to keep this out of the code and only in the CI context?
2. On the other hand, could rate limiting be applied to normal operations, affecting cli users?

dasconnor (approved):
LGTM

pkg/binutils/latest.go
felipemadero: Regarding 2, no issue for an user. Rate limiting applies currently only to subnet create operations, the user may need to create 60 subnets within an hour to reach so limit, which is improbable.
Regarding 1, env var seems a good compromise.

PR #221, opened 2022-08-25
216 subnet publish

holisticode:


cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+func newPublishCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "publish [subnetName]",
+		Short:        "Publish the subnet's VM to a repository",
+		Long:         ``,
+		SilenceUsage: true,
+		RunE:         publish,
+		Args:         cobra.ExactArgs(1),
+	}
+	cmd.Flags().StringVar(&alias, "alias", "", "An alias for the repo to publish to")
```
dasconnor, line 39: Reading the help string alone, I'm not sure what this is supposed to be. Maybe we need an example value

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+			// there is a repo already for this alias, let's try to figure out the remote URL from there
+			conf, err := repo.Config()
+			if err != nil {
+				// TODO Would we really want to abort here?
+				return err
+			}
+			remotes := make([]string, len(conf.Remotes))
+			i := 0
+			for _, r := range conf.Remotes {
+				// TODO: Usually there's only one URL for a remote, but more are supported - should we too?
```
dasconnor, line 205: I think one remote is fine

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+				remotes[i] = r.URLs[0]
+				i++
+			}
+			repoURL, err = app.Prompt.CaptureList("Which is the remote URL for this repo?", remotes)
+			if err != nil {
+				// TODO Would we really want to abort here?
```
dasconnor, line 211: I don't think this error can ever really hit. It would require some deep, messed up stuff in the prompt library. Don't worry about it.

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+	return yaml.Unmarshal(fileBytes, &defType)
+}
+
+func getSubnetInfo(sc *models.Sidecar) (*types.Subnet, error) {
+	homepage, err := app.Prompt.CaptureEmpty("What is the homepage of the Subnet project?", nil)
+	if err != nil {
+		return nil, err
+	}
+
+	desc, err := app.Prompt.CaptureEmpty("Please provide a free-text description of the Subnet", nil)
```
dasconnor, line 240: I would remove all of the `please`s

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	maintrs, canceled, err := app.Prompt.CaptureListDecision(
```
dasconnor, line 245: Glad we're reusing this!

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+		return nil, errors.New("canceled by user")
+	}
+
+	strMaintrs := make([]string, len(maintrs))
+	for i, m := range maintrs {
+		strMaintrs[i] = m.(string)
+	}
+
+	// TODO: In this version, we are publishing 1 subnet and exactly 1 VM.
+	// 1. Will the VM list has to be editale in the future?
```
dasconnor, line 268: No, it should be a property of the subnet config. At this point, the subnet should be "locked-in" and deployed to fuji or prod.

dasconnor: Well, actually I could see a situation where another chain is added later. Theoretically possible, but probably not worth worrying about right now.

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+	// 2. If yes, need to think about the whole process
+	// 3. Can VMs later be added independently?
+	/*
+		vms, canceled, err := app.Prompt.CaptureListDecision(
```
dasconnor, line 272: Definitely shouldn't be a "decision." It should be taken from the config.

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+		Maintainers: strMaintrs,
+		// VMs:         strVMs,
+		VMs: nil,
+	}
+
+	return subnet, nil
+}
+
+func getVMInfo(sc *models.Sidecar) (*types.VM, error) {
+	vmID, err := app.Prompt.CaptureEmpty("What is the ID of this VM?", nil)
```
dasconnor, line 309: We probably only want to ask this if it is a custom vm. Otherwise use subnet-evm or spaces.

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	desc, err := app.Prompt.CaptureEmpty("Provide a description for this VM please", nil)
```
dasconnor, line 314: We should probably have a stock description for subnet-evm and spaces.

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	maintrs, canceled, err := app.Prompt.CaptureListDecision(
```
dasconnor, line 319: I'm not 100% sure of this, but do we maybe only want to capture one set of maintainers and use them in both places?

dasconnor: Alternatively, should we add ava labs if using spaces or subnet-evm?

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+		ux.Logger.PrintToUser("Publishing aborted")
+		return nil, errors.New("canceled by user")
+	}
+
+	strMaintrs := make([]string, len(maintrs))
+	for i, m := range maintrs {
+		strMaintrs[i] = m.(string)
+	}
+
+	scr, err := app.Prompt.CaptureEmpty("What scripts needs to run to install this VM?", nil)
```
dasconnor, line 341: This needs to be an executable command to build the vm. Probably need to make that more clear.

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	bin, err := app.Prompt.CaptureEmpty("What is the binary path?", nil)
```
dasconnor, line 346: This should be the output of the build command.

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	url, err := app.Prompt.CaptureEmpty("Tell us the URL to download the binary please", nil)
```
dasconnor, line 351: Might want to add needs to be a fixed version, not latest. And we need to download source, not a bin.

pkg/prompts/prompts.go
```diff
@@ -155,6 +165,25 @@ func validateBiggerThanZero(input string) error {
 	return nil
 }
 
+func validateSemanticVersion(input string) error {
+	_, err := version.Parse(input)
+	return err
+}
+
+func (r *realPrompter) CaptureSemanticVersion(promptStr string) (*version.Semantic, error) {
```
dasconnor, line 173: Is this different than `CaptureVersion`?

pkg/prompts/prompts.go
```diff
@@ -500,6 +529,33 @@ func (r *realPrompter) CaptureAnyList(promptStr string, options any) (any, error
...
+
+	str, err := prompt.Run()
+	if err != nil {
+		return "", err
+	}
+
+	return str, nil
+}
+
+func (*realPrompter) CaptureEmpty(promptStr string, arg any) (any, error) {
```
dasconnor, line 546: Not sure what `CaptureEmpty` means?

pkg/subnet/publisher.go
```diff
@@ -0,0 +1,110 @@
...
+	subnetName, vmName string,
+	subnetYAML []byte,
+	vmYAML []byte,
+) error {
+	wt, err := repo.Worktree()
+	if err != nil {
+		return err
+	}
+	// TODO: This might not always be the right path!
+	// TODO: Use constants
```
dasconnor, line 63: probably want to implement this todo

scripts/regenerate_mocks.sh
```diff
@@ -7,8 +7,9 @@ fi
 
 go install github.com/vektra/mockery/v2@latest
 
-mockery -r --output ./internal/mocks --name BinaryChecker --filename binaryChecker.go
```
dasconnor, line 10: Undo this

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+			conf, err := repo.Config()
+			if err != nil {
+				// TODO Would we really want to abort here?
+				return err
+			}
+			remotes := make([]string, len(conf.Remotes))
+			i := 0
+			for _, r := range conf.Remotes {
+				// TODO: Usually there's only one URL for a remote, but more are supported - should we too?
+				remotes[i] = r.URLs[0]
```
felipemadero, line 206: you can use append instead of an index, if using capacity in make. in that way you avoid introducing the index element as a extra posible cause of bugs (eg aliasing)

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+func loadYAMLFile[T types.Definition](path string, defType T) error {
+	fileBytes, err := os.ReadFile(path)
+	if err != nil {
+		return err
+	}
+	return yaml.Unmarshal(fileBytes, &defType)
+}
+
+func getSubnetInfo(sc *models.Sidecar) (*types.Subnet, error) {
+	homepage, err := app.Prompt.CaptureEmpty("What is the homepage of the Subnet project?", nil)
```
felipemadero, line 235: CaptureString. Also for the rest

holisticode: CaptureString requires a valid string. But according to @dasconnor we want to allow people to be able to just hit enter everywhere - which means allow empty values.

pkg/prompts/prompts.go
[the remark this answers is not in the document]
holisticode: CaptureString requires a valid string. But according to @dasconnor we want to allow people to be able to just hit enter everywhere - which means allow empty values.

cmd/subnetcmd/publish.go
holisticode: Trying this time with `We publish to a remote repo, but store identify the repo locally under a user-provided alias (e.g. myrepo).`

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+	if err = getRepoURL(reposDir); err != nil {
+		return err
+	}
+
+	var (
+		tsubnet *types.Subnet
+		vm      *types.VM
+	)
+	// TODO: Do we need to check if it has already been published?
+	// If yes, do we query the local repo, the remote, do we write into the sidecar, ...?
```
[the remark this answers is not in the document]
holisticode: The issue is not so much about what to do if it already has been published, but rather how to identify that it already _has_ been published. 

The answer is not straightforward, as publishing could basically be done on multiple repos.

Now applying a simplistic approach which just checks if _any locally cloned repo_ appears to have the subnet published (by checking if the repo's `subnets` folder contains the subnet)

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+		tsubnet, err = getSubnetInfo(sc)
+	} else {
+		tsubnet = new(types.Subnet)
+		err = loadYAMLFile(subnetDescPath, tsubnet)
+	}
+	if err != nil {
+		return err
+	}
+
+	ux.Logger.PrintToUser("Nice! We got the subnet info. Let's now get the VM details")
```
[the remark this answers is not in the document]
holisticode: What exactly do you mean here? We don't collect things like homepage, description and maintainers in the sidecar (yet). We have to ask the user?

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+		Short:        "Publish the subnet's VM to a repository",
+		Long:         ``,
+		SilenceUsage: true,
+		RunE:         publish,
+		Args:         cobra.ExactArgs(1),
+	}
+	cmd.Flags().StringVar(&alias, "alias", "", "An alias for the repo to publish to")
+	cmd.Flags().StringVar(&repoURL, "repo-url", "", "The URL of the repo where we are publishing")
+	cmd.Flags().StringVar(&vmDescPath, "vm-file-path", "", "Path to the VM description file. If not given, a prompting sequence will be initiated.")
+	cmd.Flags().StringVar(&subnetDescPath, "subnet-file-path", "", "Path to the Subnet description file. If not given, a prompting sequence will be initiated.")
```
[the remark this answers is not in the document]
holisticode: `local-only`? `no-push`? 

`offline-only` doesn't tell me much, personally. Using `local-only` for now, feel free to suggest others.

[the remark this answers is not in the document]
holisticode: Note that this creates problems with the new  `isAlreadyPublished` check, which would return published because it would find the subnetName in the local repo structure.

TBH I am not so sure this is a good idea to have this.

cmd/subnetcmd/publish.go
holisticode: OK. Removing all this and setting the `sidecar.Name` for VM

cmd/subnetcmd/publish.go
holisticode: I thought of these questions too and couldn't get to a definitive answer either. 

Like - I don't think we can just provide the same maintainers. 
However, I don't like to have to ask nearly the same question twice either.

Maybe we should do as above - if CustomVM, ask twice, otherwise just use spaces and subnet-evm. Probably what you had in mind too.

cmd/subnetcmd/publish.go
holisticode: Clarified a bit. But I think we would want to point to some docs (I guess the APM docs), which I don't think exist yet?

cmd/subnetcmd/publish.go
holisticode: Same here.

cmd/subnetcmd/publish.go
holisticode: Same

pkg/prompts/prompts.go
holisticode: No! Missed that, will remove.

holisticode: Actually, my version returned an actual `*version.Semantic` object instead of a string. Removed nonetheless.

pkg/prompts/prompts.go
holisticode: You required that for these data, it is possible to just hit enter, which means it needs to be possible to allow empty values (empty strings). `CaptureString` does not allow empty strings.

pkg/subnet/publisher.go
holisticode: Fixed

pkg/subnet/publisher.go
```diff
@@ -0,0 +1,110 @@
...
+	repoPath string
+}
+
+var _ Publisher = &publisherImpl{}
+
+func NewPublisher(repoDir, repoURL, alias string) Publisher {
+	repoPath := filepath.Join(repoDir, alias)
+	return &publisherImpl{
+		alias:    alias,
+		repoDir:  repoDir,
```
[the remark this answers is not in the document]
holisticode: Yep, thanks. Was required for an earlier iteration, not anymore.

pkg/subnet/publisher.go
```diff
@@ -0,0 +1,110 @@
...
+) error {
+	wt, err := repo.Worktree()
+	if err != nil {
+		return err
+	}
+	// TODO: This might not always be the right path!
+	// TODO: Use constants
+	subnetPath := filepath.Join(p.repoPath, "subnets", subnetName)
+	vmPath := filepath.Join(p.repoPath, "vms", vmName)
+	err = os.WriteFile(subnetPath, subnetYAML, constants.DefaultPerms755)
```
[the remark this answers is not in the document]
holisticode: Good point thanks. Probably cloned my test repo from remote, where it was already existing.

cmd/subnetcmd/publish.go
holisticode: Hehe, we have talked about this in the other PR. I still think the index one is the pattern I want to use.

cmd/subnetcmd/publish.go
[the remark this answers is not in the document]
dasconnor: Instead of creating the local directory in the internal structure, should we just write it to a directory of the user's choice or `/tmp`?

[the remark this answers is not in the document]
dasconnor: The process I'm thinking of here is that the user has a weird git setup and doesn't want to manage pushes through our tool. They just want the files to be generated with the right format and they'll setup the git repo themselves.

cmd/subnetcmd/publish.go
dasconnor: I'm good with indexing too

pkg/constants/constants.go
```diff
@@ -84,5 +84,10 @@ const (
 	// #nosec G101
 	GithubAPITokenEnvVarName = "GITHUB_TOKEN"
 
-	ReposDir = "repos"
+	ReposDir  = "repos"
+	SubnetDir = "subnets"
+	VMDir     = "vms"
+
+	GitRepoCommitName  = "Avalanche-CLI"
+	GitRepoCommitEmail = "info@avax.network"
```
dasconnor, line 92: Is this a real email? I'm not sure we want to override the global user and email. Especially when it comes to commit signing and stuff like that.

pkg/subnet/publisher.go
```diff
@@ -91,17 +90,16 @@ func (p *publisherImpl) Publish(
 	// TODO review these options
 	commit, err := wt.Commit(commitStr, &git.CommitOptions{
 		Author: &object.Signature{
-			Name:  "Avalanche-CLI",
-			Email: "info@avax.network", // this obviously would have to change
+			Name:  constants.GitRepoCommitName,
```
dasconnor, line 93: Do we need to set this? Can we use the global value?

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,380 @@
...
+// ask for a new alias
+func getNewAlias() (string, error) {
+	return app.Prompt.CaptureString("Please provide an alias for the repository we are going to use")
+}
+
+func getRepoURL(reposDir string) error {
+	if repoURL == "" {
+		path := filepath.Join(reposDir, alias)
+		repo, err := git.PlainOpen(path)
+		if err != nil {
```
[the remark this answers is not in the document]
felipemadero: Is not a problem for the tool if there is a directory (ErrRepositoryNotExists not happens) for that repo which has some type of inconsistency and cannot be readed?

dasconnor (commented):
Nice improvements. I think the prompt flow is a bit better. I do have some questions about the commit metadata tho.

cmd/subnetcmd/publish.go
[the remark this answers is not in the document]
dasconnor: Yeah, if the local directory that you created already exists, you made it before.

cmd/subnetcmd/publish.go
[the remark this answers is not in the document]
dasconnor: We should check that the sidecar has a subnetID and blockchainID, meaning that the subnet has been pushed to fuji or mainnet.

cmd/subnetcmd/publish.go
dasconnor: that sounds great!

cmd/subnetcmd/publish.go
dasconnor: agreed

pkg/subnet/publisher.go
```diff
@@ -0,0 +1,108 @@
...
+	return &publisherImpl{
+		alias:    alias,
+		repoURL:  repoURL,
+		repoPath: repoPath,
+	}
+}
+
+func (p *publisherImpl) GetRepo() (repo *git.Repository, err error) {
+	// path exists
+	if _, err = os.Stat(p.repoPath); err == nil {
```
felipemadero, line 41: for err pattern, you probably want to use `:=` instead of `=`, and keep that variable in the if score, as you don't need and probably don't want to modify a previous err value

felipemadero: applies to other occurrences of the end pattern

pkg/prompts/prompts.go
dasconnor: Why are you returning an `any` type. Shouldn't this be `CaptureStringAllowEmpty()` and return `string`?

pkg/subnet/publisher.go
```diff
@@ -0,0 +1,108 @@
...
+	subnetYAML []byte,
+	vmYAML []byte,
+) error {
+	wt, err := repo.Worktree()
+	if err != nil {
+		return err
+	}
+	// TODO: This might not always be the right path!
+	subnetPath := filepath.Join(p.repoPath, constants.SubnetDir, subnetName)
+	if err := os.MkdirAll(filepath.Join(p.repoPath, constants.SubnetDir), constants.DefaultPerms755); err != nil {
```
felipemadero, line 62: I should use filepath.Dir(subnetPath)

pkg/subnet/publisher.go
```diff
@@ -0,0 +1,108 @@
...
+		return err
+	}
+	vmPath := filepath.Join(p.repoPath, constants.VMDir, vmName)
+	if err := os.MkdirAll(filepath.Join(p.repoPath, constants.VMDir), constants.DefaultPerms755); err != nil {
```
felipemadero, line 66: I should use filepath.Dir(vmPath)

pkg/prompts/prompts.go
```diff
@@ -510,6 +518,33 @@ func (r *realPrompter) CaptureAnyList(promptStr string, options any) (any, error
...
```
felipemadero, line 535: probably call it CaptureString and the other, CaptureNonEmptyString ? Or something else, but would like to see a description of the captured value after Capture prefix, just Empty does not follow the other names convention

felipemadero, line 535: I think these is not need to add a `arg any` param

cmd/subnetcmd/publish.go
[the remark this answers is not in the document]
holisticode: Ah understood.

pkg/constants/constants.go
holisticode: OK so this library will clone a repo locally. When it wants to publish, it first basically does a git-add and a git-commit locally before a git-push to the (given) remote.

Any git commit action requires a name and an email for the repo config.

I don't think this library overrides the global user and email, nor do I think that it can easily get the global user and email. I guess what it can do is try executing `git config` via a os.Exec and get it from there, but that does not guarantee we get a value (scenario: new machine setup on EC2, git just installed but not configured yet).

I guess what we could do is try the above, then if we get a value, use that, otherwise use a generic config as above.

The other approach would be to leave as-is, then in whatever repo it gets pushed, it clearly can be identified as pushed by the tool (CLI). This could also be interesting to be able to discern.

cmd/subnetcmd/publish.go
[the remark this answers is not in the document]
holisticode: Added a new `isReadyToPublish` function

cmd/subnetcmd/publish.go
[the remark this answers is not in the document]
holisticode: I feel this might create more confusion. Also, nobody actually needs to know this is using git underneath. Everybody _thinks_ to know git, if we hand out control we indeed loose control of what's going on altogether and can't manage any state properly I am afraid.

For the process you are thinking of, it might be best to still do `local-only` (no push), then have the user grab the files and put/push them wherever/however they want.

pkg/prompts/prompts.go
holisticode: Yes that is probably a better suggestion. I used the `any` pattern because initially I had a scenario which was querying through `CaptureListDecision`.

We don't have that anymore so returning a string is way better now.

cmd/subnetcmd/publish.go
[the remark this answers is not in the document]
holisticode: Ah I see what you mean now. Good question.

There are a series of errors which could return here which are not `ErrRepositoryNotExists`. But the possible errors are of different kind, e.g. filesystem issues or internal git issues. Would have to do a proper scan and discriminate between all possible errors to be able to identify what to do.

I decided therefore to for now just leave a debug log which also logs the error in case this becomes an issue, but to still ignore it for now.

pkg/subnet/publisher.go
holisticode: Again, will try to get information from `git config`, and if not available, just use this one as a fallback.

pkg/subnet/publisher.go
holisticode: Agreed. Although it can always be assumed that the previous err value has been checked (the linter will complain if not).

pkg/prompts/prompts.go
holisticode: Yep has been reported by @dasconnor as well. Already addressed.

holisticode: Already fixed too

cmd/subnetcmd/publish.go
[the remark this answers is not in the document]
dasconnor: Yes, I think that's what I'm suggesting

pkg/constants/constants.go
dasconnor: I'm not really satisfied with this approach, though I don't immediately know the answer. I think there are some corner cases that might cause problems and I'm not sure it will meet the security level we want to provide. I think this is ok for beta version 1, but let's spend some time next week and come up with a really concrete plan.

dasconnor: I think you good a job, the design spec was just unclear. This part's on me.

dasconnor (commented):
We're getting really close!

felipemadero (commented):
LGTM. As far as other reviewer comments are addressed is ok.

cmd/subnetcmd/publish.go
```diff
@@ -146,6 +146,36 @@ func doPublish(sc *models.Sidecar, subnetName string, publisherCreateFunc newPub
...
 	}
 
+	if noRepoPath != "" {
+		ux.Logger.PrintToUser("Writing the file specs to the provided directory at: %s", noRepoPath)
+		// the directory does not exist
+		if _, err := os.Stat(noRepoPath); err != nil {
+			if err := os.MkdirAll(noRepoPath, constants.DefaultPerms755); err != nil {
+				return fmt.Errorf("attempted to create the given --no-repo-path directory at %s, but failed: %w", noRepoPath, err)
+			}
+			ux.Logger.PrintToUser("The given --no-repo-path at %s was not existing; created it with permissions %o", noRepoPath, constants.DefaultPerms755)
```
dasconnor, line 156: `was not existing` -> `did not exist`

cmd/subnetcmd/publish.go
```diff
@@ -146,6 +146,36 @@ func doPublish(sc *models.Sidecar, subnetName string, publisherCreateFunc newPub
...
+		if !forceWrite {
+			// do not automatically overwrite
+			if _, err := os.Stat(subnetFile); err == nil {
+				return fmt.Errorf("a file with the name %s already exists. If you wish to overwrite, provide the %s flag", subnetFile, forceFlag)
+			}
+			if _, err := os.Stat(vmFile); err == nil {
+				return fmt.Errorf("a file with the name %s already exists. If you wish to overwrite, provide the %s flag", vmFile, forceFlag)
+			}
+		}
+		if err := os.WriteFile(subnetFile, subnetYAML, constants.DefaultPerms755); err != nil {
```
dasconnor, line 169: These need to be in a `vms` and `subnets` directory

Screen Shot 2022-09-02 at 4 12 38 PM

dasconnor (commented):
One last thing to fix. We should also create a ticket to add an E2E test for publish.

cmd/subnetcmd/publish.go
holisticode: ok fixed

holisticode:
https://github.com/ava-labs/avalanche-cli/issues/255

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,462 @@
...
+	for i, m := range maintrs {
+		strMaintrs[i] = m
+	}
+
+	subnet := &types.Subnet{
+		ID:          sc.Networks[models.Fuji.String()].SubnetID.String(),
+		Alias:       sc.Name,
+		Homepage:    homepage,
+		Description: desc,
+		Maintainers: strMaintrs,
```
felipemadero, line 362: you can just use maintrs here, and avoid the copy

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,462 @@
...
+	)
+	if err != nil {
+		return nil, err
+	}
+	if canceled {
+		ux.Logger.PrintToUser("Publishing aborted")
+		return nil, errors.New("canceled by user")
+	}
+
+	strMaintrs := make([]string, len(maintrs))
```
felipemadero, line 352: you can avoid this copy now

cmd/subnetcmd/publish.go
```diff
@@ -0,0 +1,462 @@
...
+		)
+		if err != nil {
+			return nil, err
+		}
+		if canceled {
+			ux.Logger.PrintToUser("Publishing aborted")
+			return nil, errors.New("canceled by user")
+		}
+
+		strMaintrs = make([]string, len(maintrs))
```
felipemadero, line 402: you can avid this copy now

cmd/subnetcmd/publish.go
```diff
@@ -332,14 +333,13 @@ func getSubnetInfo(sc *models.Sidecar) (*types.Subnet, error) {
 		return nil, err
 	}
 
-	maintrs, canceled, err := app.Prompt.CaptureListDecision(
```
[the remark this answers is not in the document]
dasconnor: ohh, I see, we pass in the prompt as an arg, interesting...

dasconnor (approved):
Felipe had a good nit, but I think we're basically ready

felipemadero (approved):

PR #222, opened 2022-08-25
Add ability to import subnets via APM

### LINKED ISSUE #215: Add APM integration to Subnet Import command
- The join command should accept an APM-compatible repo instead of just the manual vm bin and genesis
- We should add the repo on the backend and run an install

dasconnor:
# Summary of Changes
This PR integrates the APM into the CLI and uses it in a revamped `subnet import` command. Imported subnets are NOT deployable and can only be joined with `subnet join`. In addition to integrating the APM, I also slightly refactored the `subnet join` command to move some of its functionality to a new `plugins` package that will be used to manage node configuration and vm binary copying.

Closes #215

# Details
- Created a new `apmintegration` package to manage the CLI's integration with the APM
- The root cmd now calls `SetupAPM()` when it initializes the app's state
  - This initialization command can only be run once at a time. When we run the hidden network runner backend command, we cannot run `SetupAPM()`, so that setup is disabled for that command.
- I updated the sidecar with a few new values for imported subnets. I store whether or not the subnet was imported and its imported vmid.
  - The boolean `ImportedFromAPM` is false by default, so this shouldn't affect any existing subnets.
  - In the `subnet create` command, I explicitly set `ImportedFromAPM` to false.
  - I bumped the sidecar version
- `subnet deploy` now checks that `ImportedFromAPM` is false. If it is true, the deploy is aborted.
- I overhauled the `subnet import command to be able to import subnets from a file or from a repo.
  - If you pass in the same command line command that used to work, it will behave the same.
  - If you don't pass in a command line argument, you will put through an interactive menu. You can select an already installed repo or provide a URL to download. After, you select which subnet in the repo you want to import. After that, binaries are downloaded and copied into an apm-specific plugin directory.
  - The `ava-labs/avalanche-plugins-core` repo is loaded by the APM by default
- I moved `editConfigFile()` and `createPlugin()` out of `subnet join` and moved them into a new `pkg/plugins` package. The functions are largely unmodified. Maybe a line or two different.
- `subnet list` now displays whether or not a subnet was imported
- The `apmintegration` package has several files
  - `apm.go` has functions that call the APM's main methods
  - `file.go` has functions that read the files imported by the apm
  - `helpers.go` has some utility functions
  - `setup.go` has functions to initialize the apm
  - I wrote tests for all of these files except `apm.go`
- I updated the customVM binutil to handled apm-imported binaries
- Added constants to support the apm


internal/testutils/setup.go
```diff
@@ -17,3 +18,11 @@ func SetupTest(t *testing.T) *assert.Assertions {
 	ux.NewUserLog(logging.NoLog{}, io.Discard)
 	return assert.New(t)
 }
+
+func SetupTempTestDir() (string, error) {
```
holisticode, line 22: Rather than doing this, I suggest using `t.Tempdir()`, which is automatically cleaned up by the `testing` package.

pkg/apmintegration/file_test.go
```diff
@@ -0,0 +1,322 @@
...
+			repo:        repo1,
+			subnetNames: []string{},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert := require.New(t)
+
+			testDir, err := testutils.SetupTempTestDir()
```
holisticode, line 171: As mentioned earlier, I suggest using `t.Tempdir` instead

pkg/apmintegration/helpers.go
```diff
@@ -0,0 +1,51 @@
...
+
+	if len(split) < 3 {
+		return "", errors.New("unable to find organization")
+	}
+
+	return split[len(split)-2], nil
+}
+
+func getGithubRepo(url string) (string, error) {
+	base := filepath.Base(url)
```
holisticode, line 24: As this is no filesystem path, but an URL path, the `path` package should be preferred instead of filepath.

pkg/apmintegration/helpers.go
```diff
@@ -0,0 +1,51 @@
...
+	if base[len(base)-4:] != ".git" {
```
holisticode, line 25: The `path` package has the `Ext` function. https://pkg.go.dev/path#Ext

pkg/apmintegration/helpers.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"path/filepath"
+	"strings"
+)
+
+func getGithubOrg(url string) (string, error) {
+	split := strings.Split(url, "/")
```
holisticode, line 14: As this is a URL path, also the `path` package seems more appropriate (also has a `Split` func)

pkg/apmintegration/helpers.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return "", fmt.Errorf("unable to create alias: %w", err)
+	}
+
+	repo, err := getGithubRepo(url)
+	if err != nil {
+		return "", fmt.Errorf("unable to create alias: %w", err)
+	}
+
+	return org + "/" + repo, nil
```
holisticode, line 42: `return makeAlias`?

pkg/apmintegration/helpers.go
```diff
@@ -0,0 +1,51 @@
...
+package apmintegration
+
+import (
+	"errors"
+	"fmt"
+	"path/filepath"
+	"strings"
+)
+
+func getGithubOrg(url string) (string, error) {
```
holisticode, line 13: Are we only supporting `github` repos?

pkg/apmintegration/helpers.go
```diff
@@ -0,0 +1,51 @@
...
+	split := strings.Split(url, "/")
+
+	if len(split) < 3 {
+		return "", errors.New("unable to find organization")
+	}
+
+	return split[len(split)-2], nil
```
holisticode, line 20: It may be useful to use index constants to refer to specific parts. We could even create a `RepoPart` type or something like that.

pkg/apmintegration/setup.go
```diff
@@ -0,0 +1,89 @@
...
+// If we need to use custom git credentials (say for private repos).
+// the zero value for credentials is safe to use.
+// Stolen from APM repo
+func initCredentials() (http.BasicAuth, error) {
+	result := http.BasicAuth{}
+
+	if viper.IsSet(credentialsFileKey) {
+		credentials := &config.Credential{}
+
+		bytes, err := os.ReadFile(viper.GetString(credentialsFileKey))
```
holisticode, line 76: Are we using unencrypted credential files? Is that no issue?

holisticode (commented):
Question: Can we also not deploy an imported VM to local?

pkg/apmintegration/setup.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return err
+	}
+	defer apmLog.Close()
+	os.Stdout = apmLog
+	fmt.Println("testing log print")
+	apmConfig := apm.Config{
+		Directory:        apmBaseDir,
+		Auth:             credentials,
+		AdminAPIEndpoint: viper.GetString(adminAPIEndpointKey),
```
felipemadero, line 52: do we want a default value if this is empty?

pkg/apmintegration/setup.go
```diff
@@ -0,0 +1,89 @@
...
+	result := http.BasicAuth{}
+
+	if viper.IsSet(credentialsFileKey) {
+		credentials := &config.Credential{}
+
+		bytes, err := os.ReadFile(viper.GetString(credentialsFileKey))
+		if err != nil {
+			return result, err
+		}
+		if err := yaml.Unmarshal(bytes, credentials); err != nil {
```
felipemadero, line 80: are we using yaml due to apm requirement?

pkg/apmintegration/setup.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// Note, you can only call this method once per run
+func SetupApm(app *application.Avalanche, apmBaseDir string) error {
+	credentials, err := initCredentials()
+	if err != nil {
+		return err
+	}
+
+	fs := afero.NewOsFs()
```
felipemadero, line 32: please add comment on why we are doing this

pkg/plugins/avagoConfig.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/ava-labs/avalanche-cli/pkg/application"
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+)
+
+func EditConfigFile(app *application.Avalanche, subnetID string, networkID string, configFile string, forceWrite bool) error {
```
felipemadero, line 18: maybe split the args once per line

felipemadero, line 18: probably add a comment to the function, also telling that the function creates a config file if it does not exists

pkg/binutils/custom.go
```diff
@@ -9,3 +9,8 @@ func SetupCustomBin(app *application.Avalanche, subnetName string) string {
 	// Just need to get the path of the vm
 	return app.GetCustomVMPath(subnetName)
 }
+
+func SetupAPMBin(app *application.Avalanche, vmid string) string {
```
felipemadero, line 13: why is this in custom? probably this can be named utils or something else to avoid confusion with custom vm stuff

cmd/subnetcmd/deploy.go
```diff
@@ -104,6 +104,17 @@ func deploySubnet(cmd *cobra.Command, args []string) error {
...
 	}
 
+	chain := chains[0]
+
+	sc, err := app.LoadSidecar(chain)
+	if err != nil {
+		return fmt.Errorf("failed to load sidecar for later update: %w", err)
+	}
+
+	if sc.ImportedFromAPM {
```
felipemadero, line 114: why is this restriction?

cmd/subnetcmd/import.go
```diff
@@ -68,3 +107,101 @@ func importSubnet(cmd *cobra.Command, args []string) error {
...
+
+	promptStr := "What repo would you like to import from"
+	repo, err := app.Prompt.CaptureList(promptStr, installedRepos)
+	if err != nil {
+		return err
+	}
+
+	if repo == customRepo {
+		promptStr = "Enter your repo URL"
+		repoURL, err := app.Prompt.CaptureString(promptStr)
```
felipemadero, line 128: repo url format validation?

cmd/subnetcmd/import.go
```diff
@@ -68,3 +107,101 @@ func importSubnet(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+
+	version := fmt.Sprintf("%d.%d.%d", vmDescr.Version.Major, vmDescr.Version.Minor, vmDescr.Version.Patch)
+
+	sidecar := models.Sidecar{
+		Name:            subnetDescr.Alias,
+		VM:              vmType,
+		VMVersion:       version,
+		Subnet:          subnetDescr.Alias,
```
felipemadero, line 186: I am not sure, asking instead of going on the code: are we having redundant fields sidecar.Name and sidecar.Subnet?

cmd/subnetcmd/list.go
```diff
@@ -37,7 +38,7 @@ func (c subnetMatrix) Swap(i, j int) { c[i], c[j] = c[j], c[i] }
 func (c subnetMatrix) Less(i, j int) bool { return strings.Compare(c[i][0], c[j][0]) == -1 }
 
 func listGenesis(cmd *cobra.Command, args []string) error {
```
felipemadero, line 40: maybe this function name is outdated

cmd/subnetcmd/list.go
```diff
@@ -51,7 +52,7 @@ func listGenesis(cmd *cobra.Command, args []string) error {
 
 	rows := subnetMatrix{}
 	// append a second "header" row for the networks
-	rows = append(rows, []string{"", "", "", "", "Local", "Fuji", "Mainnet"})
+	rows = append(rows, []string{"", "", "", "", "", "Local", "Fuji", "Mainnet"})
```
felipemadero, line 55: do we need to print which is the repo? version? anything like that?

internal/testutils/setup.go
```diff
@@ -17,3 +18,11 @@ func SetupTest(t *testing.T) *assert.Assertions {
...
+	return os.MkdirTemp(os.TempDir(), "cli-test")
+}
+
+func CleanTempTestDir(testDir string) {
```
felipemadero, line 26: I am not sure why we need this function, it is just calling another one

pkg/apmintegration/helpers.go
felipemadero: path is not very good with urls, specially due to the `//` part at the start of it

pkg/apmintegration/helpers.go
felipemadero: probably dont refer relative to the end

pkg/apmintegration/helpers.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func getGithubRepo(url string) (string, error) {
+	base := filepath.Base(url)
+	if base[len(base)-4:] != ".git" {
+		return "", errors.New("unable to find repo name")
+	}
+	return base[:len(base)-4], nil
```
felipemadero, line 28: avoid repeating the same index expresion over the function, probably do refer from start and not from the end, and use contants

pkg/apmintegration/helpers.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	return org + "/" + repo, nil
+}
+
+func makeAlias(org, repo string) string {
+	return org + "/" + repo
+}
+
+func makeKey(alias, subnet string) string {
```
felipemadero, line 49: this may be used also in import cmd

pkg/apmintegration/file.go
```diff
@@ -0,0 +1,110 @@
...
+	return output, nil
+}
+
+func GetSubnets(app *application.Avalanche, repoAlias string) ([]string, error) {
+	subnetDir := filepath.Join(app.ApmDir, "repositories", repoAlias, "subnets")
+	subnets, err := os.ReadDir(subnetDir)
+	if err != nil {
+		return []string{}, err
+	}
+	subnetOptions := make([]string, len(subnets))
```
felipemadero, line 46: why is this called subnetOptions?

pkg/apmintegration/file.go
```diff
@@ -0,0 +1,110 @@
...
+type SubnetWrapper struct {
+	Subnet types.Subnet `yaml:"subnet"`
+}
+
+type VMWrapper struct {
+	VM types.VM `yaml:"vm"`
+}
+
+func LoadSubnetFile(app *application.Avalanche, subnetKey string) (types.Subnet, error) {
+	splitSubnet := strings.Split(subnetKey, ":")
```
felipemadero, line 63: I think this repo:subnet creation and access will probably need to we encasulated in two functions and called elsewhere

pkg/apmintegration/apm.go
```diff
@@ -0,0 +1,59 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package apmintegration
```
joshua-kim, line 4: There'a few apm commands that are missing here (`upgrade`, `list-repositories`, etc). If it's not needed for v1 of this, then feel free to ignore.

pkg/apmintegration/helpers.go
joshua-kim: apm itself doesn't only support github repos, apm can support an arbitrary git server but has a concept of organization/repository-name (which sounds github-ish), but these are only used for its internal aliasing to short-hand/make human readable a long url (e.g org/repo (alias)  => https://www.github.com/org/repo.git (url))

pkg/apmintegration/apm.go
```diff
@@ -0,0 +1,59 @@
...
+		ux.Logger.PrintToUser("Avalanche Plugins Core already installed, skipping...")
+		return "", nil
+	}
+
+	fmt.Println("Installing repo")
+
+	return alias, app.Apm.AddRepository(alias, repoURL, branch)
+}
+
+func UpdateRepos(app *application.Avalanche) error {
```
joshua-kim, line 32: Kind of feels like we don't need this function if we're just calling `.Update()` and doing nothing else

pkg/apmintegration/file.go
```diff
@@ -0,0 +1,110 @@
...
+	return subnetOptions, nil
+}
+
+type SubnetWrapper struct {
+	Subnet types.Subnet `yaml:"subnet"`
+}
+
+type VMWrapper struct {
+	VM types.VM `yaml:"vm"`
+}
```
joshua-kim, line 60: if I understand correctly these weird wrapper things are needed because of the funny key in the apm definitions (e.g vm/subnet). I think we can probably just remove these from the definition spec (I can do this) to make this cleaner and make it so we don't need these wrappers

pkg/apmintegration/file.go
```diff
@@ -0,0 +1,110 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/ava-labs/apm/types"
+	"github.com/ava-labs/avalanche-cli/pkg/application"
+	"gopkg.in/yaml.v3"
+)
+
+func GetRepos(app *application.Avalanche) ([]string, error) {
```
joshua-kim, line 17: Note to myself but the majority of this file looks like good functionality that I should just end up exposing/refactoring this into the APM instead of having you to have to implement this with filesystem lookups upstream. Feels like avalanche-cli isn't the right place to have this code, I'll probably end up pulling this into apm at some point.

joshua-kim (commented):
Reviewed mostly just the `apmintegration` package, but mostly looks good. I'll probably need to pull in some of this code into the apm since i think it belongs there as opposed to in avalanche-cli.

pkg/apmintegration/file.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	subnetOptions := make([]string, len(subnets))
+	for i, subnet := range subnets {
+		subnetOptions[i] = subnet.Name()[:len(subnet.Name())-5]
+	}
+
+	return subnetOptions, nil
+}
+
+type SubnetWrapper struct {
```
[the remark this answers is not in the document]
dasconnor: Do you have a snippet you could send me? I think you do need it to match this example schema:
```
subnet:
  id: "<your base58 subnet id here>"
  alias: "<your unique alias here>"
  homepage: "<your website here>"
  description: |
    <your description here>
  maintainers:
    - "<your email here"
  installScript: "<your install script path here>"
  vms:
    - "<your vms here>"
  config:
    # This the default subnet config as of avalanchego v1.7.10
    - gossipAcceptedFrontierValidatorSize: 0
    - gossipAcceptedFrontierNonValidatorSize: 0
    - gossipAcceptedFrontierPeerSize: 35
    - gossipOnAcceptValidatorSize: 0
    - gossipOnAcceptNonValidatorSize: 0
    - gossipOnAcceptPeerSize: 20
    - appGossipValidatorSize: 10
    - appGossipNonValidatorSize: 0
    - appGossipPeerSize: 0
    - validatorOnly: false
    - consensusParameters:
      - k: 20
      - alpha: 15
      - betaVirtuous: 15
      - betaRogue: 20
      - concurrentRepolls: 4
      - optimalProcessing: 50
      - maxOutstandingItems: 1024
      - maxItemProcessingTime: 120_000_000_000
      - parents: 5
      - batchSize: 30

```

internal/testutils/setup.go
dasconnor: Great suggestion, didn't know about this!

cmd/subnetcmd/list.go
dasconnor: Let's add that info in describe instead of list. The table in list is already getting a bit big

cmd/subnetcmd/import.go
```diff
@@ -68,3 +107,101 @@ func importSubnet(cmd *cobra.Command, args []string) error {
...
+
+	// Populate the sidecar and create a genesis
+	subnetDescr, err := apmintegration.LoadSubnetFile(app, subnetKey)
+	if err != nil {
+		return err
+	}
+
+	var vmType models.VMType = models.CustomVM
+
+	vmDescr, err := apmintegration.LoadVMFile(app, repo, subnetDescr.VMs[0])
```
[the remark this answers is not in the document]
dasconnor: added a check

cmd/subnetcmd/import.go
```diff
@@ -68,3 +107,101 @@ func importSubnet(cmd *cobra.Command, args []string) error {
...
+		Subnet:          subnetDescr.Alias,
+		TokenName:       constants.DefaultTokenName,
+		Version:         constants.SidecarVersion,
+		ImportedFromAPM: true,
+		ImportedVMID:    vmDescr.ID,
+	}
+
+	ux.Logger.PrintToUser("Selected subnet, installing " + subnetKey)
+
+	err = apmintegration.InstallVM(app, subnetKey)
```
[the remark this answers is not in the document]
dasconnor: I wasn't sure if we did that if `err` was already defined and that didn't actually minimize scope, but I went ahead and did it anyway.

pkg/apmintegration/file.go
```diff
@@ -0,0 +1,110 @@
...
+
+func GetSubnets(app *application.Avalanche, repoAlias string) ([]string, error) {
+	subnetDir := filepath.Join(app.ApmDir, "repositories", repoAlias, "subnets")
+	subnets, err := os.ReadDir(subnetDir)
+	if err != nil {
+		return []string{}, err
+	}
+	subnetOptions := make([]string, len(subnets))
+	for i, subnet := range subnets {
+		subnetOptions[i] = subnet.Name()[:len(subnet.Name())-5]
```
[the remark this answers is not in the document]
dasconnor: fixed with a better method

pkg/apmintegration/file.go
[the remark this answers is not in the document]
dasconnor: Josh actually mentioned removing the top level `subnet:` and `vm:` tags on the APM side. I think that might be the best solution.

internal/testutils/setup.go
dasconnor: Implemented.

pkg/apmintegration/file_test.go
dasconnor: fixed

pkg/apmintegration/helpers.go
dasconnor: fixed

pkg/apmintegration/helpers.go
dasconnor: Right now, we're requiring every URL end in `.git`, but I might try to remove that.

pkg/apmintegration/helpers.go
dasconnor: The split function in `path` doesn't do exactly what we need. The `strings` split works better. It might be worth moving entirely to URL tho.

pkg/apmintegration/helpers.go
dasconnor: fixed

pkg/apmintegration/helpers.go
dasconnor: renamed, we should support any repo but it wouldn't totally surprise me if we hit a bug for a user using gitlab or something

pkg/apmintegration/helpers.go
dasconnor: I think indexing backwards is the best way to do this. In general, I define the `Org` to be the second to last element in the path and then the repo is the last part. Defining relative to the front is a lot harder since there can be a lot more variability in if you have `https://` or not. Using the `url` package is also a bit finnicky because it doesn't do the same error checking.

pkg/apmintegration/setup.go
dasconnor: We talked about this in stand-up. This is inherited from the APM, but I think it's fine for now. We can investigate more before going to production.

pkg/apmintegration/setup.go
```diff
@@ -0,0 +1,89 @@
...
+	// The New() function has a lot of prints we'd like to hide from the user,
+	// so going to divert stdout to the log temporarily
+	stdOutHolder := os.Stdout
+	apmLog, err := os.OpenFile(app.GetAPMLog(), os.O_APPEND|os.O_CREATE|os.O_WRONLY, constants.DefaultPerms755)
+	if err != nil {
+		return err
+	}
+	defer apmLog.Close()
+	os.Stdout = apmLog
+	fmt.Println("testing log print")
```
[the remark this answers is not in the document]
dasconnor: fixed

cmd/root.go
```diff
@@ -74,6 +75,18 @@ func createApp(cmd *cobra.Command, args []string) error {
...
 	cf := config.New()
 	app.Setup(baseDir, log, cf, prompts.NewPrompter())
+	// Setup APM, skip if running a hidden command
+	if !cmd.Hidden {
+		usr, err := user.Current()
+		if err != nil {
+			app.Log.Error("unable to get system user")
+			return err
+		}
+		apmBaseDir := filepath.Join(usr.HomeDir, constants.APMDir)
```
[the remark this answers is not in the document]
dasconnor: That is the apm's directory. If you use the apm outside of the cli, it should still have the same data and packages downloaded.

pkg/apmintegration/setup.go
```diff
@@ -0,0 +1,89 @@
...
+
+	fs := afero.NewOsFs()
+
+	err = os.MkdirAll(app.GetAPMPluginDir(), constants.DefaultPerms755)
+	if err != nil {
+		return err
+	}
+
+	// The New() function has a lot of prints we'd like to hide from the user,
+	// so going to divert stdout to the log temporarily
```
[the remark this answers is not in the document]
dasconnor: This would be a good suggestion, @joshua-kim

pkg/apmintegration/setup.go
dasconnor: The default value is the empty string. The apm handles that for us.

pkg/apmintegration/setup.go
dasconnor: Yes

pkg/apmintegration/setup.go
dasconnor: done

pkg/plugins/avagoConfig.go
dasconnor: done

dasconnor: done

pkg/binutils/custom.go
dasconnor: I thought it belonged there because it has the same structure as a custom. Instead of loading a binary from our "package management" directories, it just copies the VM over from a special folder.

cmd/subnetcmd/deploy.go
dasconnor: Because the genesis data is not included in the apm repo. It's possible we could retrieve it from the running network, but that would be an extension.

cmd/subnetcmd/import.go
dasconnor: added

cmd/subnetcmd/import.go
dasconnor: This is to support a future feature where we have multiple VMs per subnet.

cmd/subnetcmd/list.go
dasconnor: updated

internal/testutils/setup.go
dasconnor: removed

pkg/apmintegration/helpers.go
dasconnor: I cleaned this up

pkg/apmintegration/helpers.go
dasconnor: fixed

pkg/apmintegration/file.go
dasconnor: Because we're letting the user choose which subnet they want

pkg/apmintegration/apm.go
dasconnor: not needed for v1!

pkg/apmintegration/apm.go
dasconnor: I added it so that a caller only needs to call this package and doesn't need to call the apm directly.

pkg/apmintegration/file.go
dasconnor: Yes, I think that would be a good fix

pkg/apmintegration/file.go
dasconnor: Yep, I'd agree with that. We'll leave them here for now then remove when you update the apm.

pkg/apmintegration/helpers.go
holisticode: The best approach is to use `url.Parse` to get (and validate) a well-formed URL (including eventual password and username, query strings, etc.), then the `path` package can be used to manipulate the `url.Path` part of a URL.

`The path package should only be used for paths separated by forward slashes, such as the paths in URLs` https://pkg.go.dev/path#pkg-overview

pkg/apmintegration/helpers.go
holisticode: `url.Parse` will create objects which already split away the host part, the scheme part etc. away from the actual path and other elements. That is for sure the cleanest way to handle a URL. Going backwards is error-prone.

pkg/apmintegration/file.go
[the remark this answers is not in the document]
dasconnor: Let's leave this as is for now and update once we have a new APM version out

pkg/apmintegration/setup.go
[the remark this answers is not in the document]
felipemadero: https://github.com/ava-labs/apm/issues/23

cmd/subnetcmd/deploy.go
felipemadero: I think we may need a way to provide a genesis when importing the subnet, or maybe add it later, either by modifying the subnet config with a new command, or by adding an option to deploy.

felipemadero: maybe we want a new issue for that

cmd/subnetcmd/import.go
felipemadero: not sure to understand. Subnet field is used a the CLI subnet name, which is pretty 1-to-1 with a vm. in the future are we going to have multiple vms in a CLI subnet definition?

pkg/apmintegration/file.go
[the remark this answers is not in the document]
dasconnor: Tracked here https://github.com/ava-labs/apm/issues/22

cmd/subnetcmd/deploy.go
dasconnor: We can try addressing in the next release.

https://github.com/ava-labs/avalanche-cli/issues/238

cmd/subnetcmd/list.go
dasconnor: #239

pkg/apmintegration/file.go
dasconnor: https://github.com/ava-labs/apm/issues/24

felipemadero (commented):
LGTM. As far as the other reviewer points are addressed.

pkg/apmintegration/helpers.go
dasconnor: I removed the requirement to end in `.git`. The tool now accepts with or without the extension.

pkg/apmintegration/helpers.go
dasconnor: I switched to using the url package. I use `ParseRequestURI` instead of `Parse` because Parse accepts relative paths and we only want absolute paths.

pkg/apmintegration/helpers.go
dasconnor: Now we define the repo as the last element of the path and the org is everything after the TLD and before the repo.

joshua-kim (approved):

Issue #140, opened 2022-07-26
Add tests for subnet import/export


dasconnor:
We might want to wait until after our APM integration.

PR #237, opened 2022-08-31
Added export/import unit tests

holisticode:
Closes https://github.com/ava-labs/avalanche-cli/issues/140


dasconnor (approved):
It'd be nice to also make an E2E test, but that can be a follow up PR

PR #242, opened 2022-09-01
Refactor state machines

felipemadero:


pkg/statemachine/stateMachine.go
```diff
@@ -0,0 +1,53 @@
...
+
+func NewStateMachine(states []string) *StateMachine {
+	return &StateMachine{
+		states: states,
+	}
+}
+
+func (sm *StateMachine) CurrentState() string {
+	if sm.index < 0 || sm.index >= len(sm.states) {
+		return ""
```
holisticode, line 27: It already seems very improbable that this can happen at all, but I feel like in this case we might want to return an error instead? Or an error string constant at least?

pkg/statemachine/stateMachine.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func (sm *StateMachine) NextState(direction StateDirection) string {
+	switch direction {
+	case Forward:
+		sm.index++
+	case Backward:
+		sm.index--
+	default:
+		return ""
```
holisticode, line 39: ...then here we could also return the same error string constant...

pkg/statemachine/stateMachine.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	if sm.index == len(sm.states) {
```
holisticode, line 41: This looks like a very important invariant: as long as I do not misunderstand, for this to always be true, the last element of `StateDirection` needs always to be `Stop`, so we might want to document this clearly somewhere.

pkg/vm/createEvm.go
```diff
@@ -73,54 +54,48 @@ func CreateEvmSubnetConfig(app *application.Avalanche, subnetName string, genesi
...
 
-	for stage != doneStage {
-		switch stage {
-		case startStage:
-			direction = forward
-		case descriptorStage:
+	subnetEvmState := statemachine.NewStateMachine(
+		[]string{descriptorsState, feeState, airdropState, precompilesState},
+	)
+	for subnetEvmState.Running() {
```
holisticode, line 82: It looks like there is another invariant, which is that the state direction always gets automatically initialized to `Forward` because it takes the null value if the var is not initialized.

This is clear looking at the code, but for a user of the `StateMachine` it might be also worth documenting somewhere.

pkg/vm/createEvm.go
```diff
@@ -73,54 +54,48 @@ func CreateEvmSubnetConfig(app *application.Avalanche, subnetName string, genesi
...
+		case precompilesState:
 			*conf, direction, err = getPrecompiles(*conf, app)
 		default:
 			err = errors.New("invalid creation stage")
 		}
 		if err != nil {
 			return nil, nil, err
 		}
-		stage = nextStage(stage, direction)
+		subnetEvmState.NextState(direction)
```
holisticode, line 98: In fact we are not checking here for the return value and the possible return `""` (empty string), explicitly returning an error might make things clearer.

holisticode (commented):
LGTM, I only have a couple of suggestions, BUT:
* I think there should be some short documentation what the statemachine is/does and how to use it
* There should be at least a small unit test making sure the statemachine works as expected.

dasconnor (commented):
I don't really have anything to add beyond Fabio's comments, but I agree we should document this package pretty thoroughly.

pkg/statemachine/stateMachine.go
felipemadero: I preferred for CurrentState() to not return error, so the code has less boilerplate. The user should check for Running function to get this information (empty string means stopped).

pkg/statemachine/stateMachine.go
felipemadero: avoided for NextState to return anything so the only function to do that is CurrentState

pkg/statemachine/stateMachine.go
felipemadero: I think I don't understand the invariant you refer to. I am not seeing a proposition involving StateDirection here.

pkg/vm/createEvm.go
felipemadero: state direction is controlled by the user, and passed using NextState(). is initialized to whatever the module user wants.

pkg/vm/createEvm.go
felipemadero: I avoided this function to return anything

pkg/statemachine/stateMachine.go
holisticode: ?

`NextState` _is_ returning something. The implicit assumption that the empty string is an error isn't very intuitive and needs documentation. If you don't want to return an error, at least a string representing an error condition is much clearer to understand.

pkg/statemachine/stateMachine.go
holisticode: Your statemachine is implicitly assuming that the last index is always a stop. This is not necessarily the way a state machine works, where each state has its own label and the current state isn't identified by an index but by the actual label, to avoid having to interpret the index (e.g. `state == STATE_FORWARD`, `state == STATE_BACKWARD`, `state == INIT`, `state == PAUSE`, `state == EXIT`  etc.)

pkg/vm/createEvm.go
holisticode: Yes and in this case the user did not initialized by the user (at least I did not see it).

Which means there is an implicit assumption that it starts with `Forward` if nothing has been initialized?

pkg/vm/createEvm.go
holisticode: But `NextState` _does_ return a string and it is not being checked here (it could be the empty string?)

pkg/statemachine/stateMachine.go
```diff
@@ -10,38 +12,61 @@ const (
 	Stop
 )
 
+const notRunningState = ""
+
+// keeps track of a lineal state sequence given by the non empty slice [states], which can
+// be updated with steps forward and backward by using NextState() with suitable direction.
+// starts in the first elem of [states] and ends either when Stop() is called,
+// or when state bypasses last elem of [states]
```
dasconnor, line 20: `bypasses` -> `passes the`

pkg/vm/airdrop.go
```diff
@@ -63,10 +63,6 @@ func getAllocation(
 			return nil, statemachine.Stop, err
 		}
 
-		if addressHex, ok = addressAny.(common.Address); !ok {
```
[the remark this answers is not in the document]
felipemadero: ? I am not seeing this diff in my PR

pkg/statemachine/stateMachine.go
felipemadero: yes, last state is assumed as ending state. it is now documented in the comment for how the machine works in the module.
and, I avoided giving meaning to any state because wanted the user to has his own semantics and not depend on his states. I only have the end state assumption, the possibility to stop the machine, and the backward/forward state transitions.

pkg/statemachine/stateMachine.go
holisticode: Github was showing old code.

pkg/statemachine/stateMachine.go
holisticode: The `Running` function is in a way the same "boilerplate" as having a specific end state to query...therefore, being it basically the same I see no point in insisting for a change.

holisticode (approved):

PR #254, opened 2022-09-05
Don't ask sha256 for known vms, and get those download URLs ourselves

holisticode:
Closes https://github.com/ava-labs/avalanche-cli/issues/266


cmd/subnetcmd/publish.go
```diff
@@ -451,3 +457,38 @@ func getVMInfo(sc *models.Sidecar) (*types.VM, error) {
...
+	binPath := filepath.Join(vmBinDir, repoName+"-"+strVer, vmBin)
+	sha, err := getSHA256FromDisk(binPath)
+	if err != nil {
+		return nil, nil, "", "", err
+	}
+	ver, err := version.Parse(strVer)
+	if err != nil {
+		return nil, nil, "", "", err
+	}
+	dl := binutils.NewSubnetEVMDownloader()
```
dasconnor, line 472: does this work for spaces?

dasconnor (commented):
Good change, just want to make sure we test it fully

cmd/subnetcmd/publish.go
```diff
@@ -395,43 +399,45 @@ func getVMInfo(sc *models.Sidecar) (*types.VM, error) {
 			return nil, errors.New("canceled by user")
 		}
 
+		url, err = app.Prompt.CaptureStringAllowEmpty("Tell us the URL to download the source. Needs to be a fixed version, not `latest`.")
```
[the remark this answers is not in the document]
holisticode: OK. Applied some changes.

cmd/subnetcmd/publish.go
holisticode: Good catch! It works (no error) but it stores the wrong URL

cmd/subnetcmd/publish.go
felipemadero, line 402: can we check for fixed version in capture?

cmd/subnetcmd/publish.go
```diff
@@ -395,43 +399,45 @@ func getVMInfo(sc *models.Sidecar) (*types.VM, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		sha, err = app.Prompt.CaptureStringAllowEmpty("For integrity checks, provide the sha256 commit for the used version")
+		if err != nil {
+			return nil, err
+		}
+		strVer, err := app.Prompt.CaptureVersion("This is the last question! What is the version being used? Use semantic versioning (v1.2.3)")
```
felipemadero, line 411: can we get version from url?

cmd/subnetcmd/publish.go
```diff
@@ -451,3 +457,38 @@ func getVMInfo(sc *models.Sidecar) (*types.VM, error) {
...
+		return nil, nil, "", "", err
+	}
+	dl := binutils.NewSubnetEVMDownloader()
+	inst := binutils.NewInstaller()
+	url, _, err := dl.GetDownloadURL(strVer, inst)
+	if err != nil {
+		return nil, nil, "", "", err
+	}
+
+	return maintrs, ver, url, sha, err
```
felipemadero, line 479: err is nil here

felipemadero (commented):
LGTM. only a nit and a couple of questions

cmd/subnetcmd/publish.go
holisticode: I think that's a bit overkill if we're going to have to create new Captures for every single special case. Also, we are allowing Empty here so I think it's not worth it. 

Maybe later if we're eliminating the Empty option and also have identified what the proper design for all this will be.

cmd/subnetcmd/publish.go
holisticode: As this is a custom VM, I am not sure we can safely derive the version from the url - the URL structure will be designed by the user, not us.

cmd/subnetcmd/publish.go
felipemadero: https://github.com/ava-labs/avalanche-cli/issues/267

felipemadero (approved):
LGTM

dasconnor (approved):
LGTM

PR #271, opened 2022-09-09
Find config and plugin dir

### LINKED ISSUE #158: Attempt to find config file and plugin dir
Instead of asking the user to input their config file and plugin dir in `join`, we should check a few well known locations to see if we can find them and give the user a chance to confirm.

holisticode:
Closes #158


cmd/subnetcmd/join.go
```diff
@@ -197,9 +242,12 @@ but until the node is whitelisted, it will not be able to validate this subnet.`
 	// or, pluginDir was set but not avagoConfigPath
 	// if **both** flags were set, this will be skipped...
 	if avagoConfigPath == "" {
-		avagoConfigPath, err = app.Prompt.CaptureString("Path to your existing config file (or where it will be generated)")
-		if err != nil {
-			return err
+		avagoConfigPath = findAvagoConfigPath()
+		if avagoConfigPath == "" {
```
dasconnor, line 246: Should we provide an option to override or confirm the found path? Or maybe that would be to just use flags?

dasconnor (commented):
Code looks great, just have a question about user confirmation.

cmd/subnetcmd/join.go
holisticode: Hmm. 

I think you're right. We shouldn't just take whatever it found. We should probably give the user a chance to confirm.

holisticode: Updated with a fix

cmd/subnetcmd/join.go
```diff
@@ -73,6 +74,8 @@ func init() {
 		wd,
 		home,
 		filepath.Join(home, constants.AvalancheGoRepoName),
+		filepath.Join(home, "go", "src", "github.com",
```
dasconnor, line 77: Can we make these constants?

cmd/subnetcmd/join.go
```diff
@@ -243,6 +246,18 @@ but until the node is whitelisted, it will not be able to validate this subnet.`
...
+		if avagoConfigPath != "" {
+			ux.Logger.PrintToUser(logging.Bold.Wrap(logging.Green.Wrap("Found a config file at %s")), avagoConfigPath)
+			yes, err := app.Prompt.CaptureYesNo("Is this the file we should update?")
+			if err != nil {
+				return err
+			}
+			if yes {
+				ux.Logger.PrintToUser("Will use file at path %s to update the configuration", avagoConfigPath)
+			} else {
+				avagoConfigPath = ""
```
dasconnor, line 258: This is a great way to handle this control logic!

dasconnor (commented):
One small suggestion, then ready to go

cmd/subnetcmd/join.go
holisticode: We could make constants for each part ("go", "src", etc.) individually. Which seems cumbersome.

But we can't make it a constant due to the filepath.Join() construct. 
I made it a `var` local to this file (and package) for now

Issue #274, opened 2022-09-13
Add `subnet stats` command


holisticode:
Should print a table that shows each validator, how much stake they have, how much time they have remaining, and what version VM they are running.

PR #287, opened 2022-09-14
Subnet stats

holisticode:
Show stats for a subnet

Closes https://github.com/ava-labs/avalanche-cli/issues/274

```
✔ Fuji
+------------------------------------------+-----------+---------------+--------+--------------------------------+-----------+
|                  NODEID                  | CONNECTED | STAKE-AMOUNT  | WEIGHT |           REMAINING            | VMVERSION |
+------------------------------------------+-----------+---------------+--------+--------------------------------+-----------+
| NodeID-CwGkAH4JDVo59AicyzLhxEfJ4qjE2x6B8 | false     | Not available |     50 | 363 days 23 hours 37 minutes 3 |           |
|                                          |           |               |        | seconds                        |           |
+------------------------------------------+-----------+---------------+--------+--------------------------------+-----------+
```


dasconnor:
Can you post a screenshot?

cmd/subnetcmd/stats.go
```diff
@@ -0,0 +1,214 @@
...
+
+	var (
+		startTime, endTime                        time.Time
+		localNodeID                               ids.NodeID
+		remaining, connected, stakeAmount, weight string
+		localVersionStr, versionStr               string
+	)
+
+	// try querying the local node for its node version
+	reply, err := infoClient.GetNodeVersion(ctx)
```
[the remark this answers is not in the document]
dasconnor: #291

internal/mocks/info.go
```diff
@@ -0,0 +1,356 @@
+// Code generated by mockery v2.14.0. DO NOT EDIT.
```
dasconnor, line 1: Don't you need to add this to scripts/regenerate_mocks.sh?

dasconnor:
I'm not sure what `Connected` means, we may also want to remove stake amount if it's going to be empty.

cmd/subnetcmd/stats.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	table.Render()
+	return nil
+}
+
+func buildStats(pClient platformvm.Client, infoClient info.Client, table *tablewriter.Table, subnetID ids.ID) ([][]string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// TODO do we need to query pending validators?
```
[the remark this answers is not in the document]
holisticode: Not too hard, but adds some stuff. Needs to add a column (pending or not, might overflow the current table), then merge the validators into one list and iterate them all.

Will try to add this to see how it looks.

internal/mocks/info.go
holisticode: Didn't find the PChain mock there either. In fact, it's not straight forward as the other ones because both are named just Client and needs a bit of editing. So concluded that's the reason PClient is not there.

I think it can be added with different flags than the other mocks.

cmd/subnetcmd/stats.go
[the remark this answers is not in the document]
holisticode: Can't add them all into one list...different types...trying diff approach

internal/mocks/info.go
holisticode: Hmm actually the reason is that these reside in a different repository. `mockery` seems not able to handle that very well;
`outside main module or its selected dependencies`.

Looks like mocks for other repo objects need to be created manually. That seems the reason why the platformvm Client was also not in this file.

cmd/subnetcmd/stats.go
```diff
@@ -81,29 +86,129 @@ func stats(cmd *cobra.Command, args []string) error {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// 2nd param are delegators, we don't need them here?
+	pendingValidatorsIface, _, err := pClient.GetPendingValidators(ctx, subnetID, []ids.NodeID{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to query the API endpoint for the pending validators: %w", err)
+	}
+
+	pendingValidators := make([]api.PermissionlessValidator, len(pendingValidatorsIface))
```
dasconnor, line 124: I don't think we can merge this until blueberry rolls out

cmd/subnetcmd/stats.go
```diff
@@ -81,29 +86,129 @@ func stats(cmd *cobra.Command, args []string) error {
...
+		// and this comparison fails
+		if v.NodeID == localNodeID {
+			versionStr = localVersionStr
+		}
+		// query peers for IP address of this NodeID...
+		rows = append(rows, []string{
+			v.NodeID.String(),
+			stakeAmount,
+			weight,
+			// TODO: Print in local time zone vs UTC?
```
dasconnor, line 190: definitely local

dasconnor (commented):
The code is very close, but I'm not sure what we should do about the blueberry part.

cmd/subnetcmd/stats.go
holisticode: Yep agreed. If blueberry is close, we may want to hold on.

cmd/subnetcmd/stats.go
holisticode: Ok fixed.

cmd/subnetcmd/stats.go
```diff
@@ -0,0 +1,317 @@
...
+		if v.NodeID == localNodeID {
+			versionStr = localVersionStr
+		}
+		// query peers for IP address of this NodeID...
+		rows = append(rows, []string{
+			v.NodeID.String(),
+			connected,
+			stakeAmount,
+			weight,
+			// startTime.Local().String(),
```
dasconnor, line 268: remove?

tests/e2e/testcases/subnet/public/suite.go
```diff
@@ -79,5 +79,13 @@ var _ = ginkgo.Describe("[Public Subnet]", func() {
 		gomega.Expect(err).Should(gomega.BeNil())
 		err = utils.RunHardhatTests(utils.BaseTest)
 		gomega.Expect(err).Should(gomega.BeNil())
+
+		// this is a simulation, so app is probably saving the info in the local sidecar section...TODO IS THIS SO?
```
dasconnor, line 83: not sure what to do with this comment

dasconnor (commented):
I think this is really close, we can merge soon

internal/mocks/info.go
dasconnor: It might be nice if we could add this to the readme somehow/somewhere. Might be confusing for us later if the avalanchego side changes.

holisticode: The mocks are in an `internal` package - I think it'd be weird to add this to the public README. 

I'll add a new README to the `internal/mocks` package and also add a note to the regeneration script.

cmd/subnetcmd/stats.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+	table.Render()
+	return nil
+}
+
+func buildPendingValidatorStats(pClient platformvm.Client, infoClient info.Client, table *tablewriter.Table, subnetID ids.ID) ([][]string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// 2nd param are delegators, we don't need them here?
```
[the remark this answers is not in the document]
holisticode: Great spotting! After investigating, it appears that the total weight is the sum of a validator's stake plus all its delegator's stake.

tests/e2e/testcases/subnet/public/suite.go
holisticode: We are using the local network to simulate `fuji` locally. Therefore, the actually used sidecar info is in the local section...I assumed...

cmd/subnetcmd/stats.go
```diff
@@ -183,7 +289,7 @@ func findAPIEndpoint(network models.Network) (platformvm.Client, info.Client) {
 	_, err := c.GetHeight(ctx)
 	if err == nil {
 		i = info.NewClient(constants.LocalAPIEndpoint)
-		_, err := i.GetNodeID(ctx)
+		_, _, err := i.GetNodeID(ctx)
```
dasconnor, line 292: Is this just checking to make sure the client works? Maybe add a comment to explain

internal/mocks/README.md
```diff
@@ -0,0 +1,19 @@
+# Mocks
+
+The `mocks` package contains mock objects used for testing.
+They have been generated with `mockery`.
+
+There is a `scripts/regenerate_mocks.sh` script which can be executed to regenerate **most** mocks when required (interface change).
+
+However, there is an exception:
+Interfaces in another repository (e.g. `avalanchego`) seem to pose problems to `mockery`. We get
```
dasconnor, line 9: Beautiful!

tests/e2e/testcases/subnet/public/suite.go
dasconnor: Does it matter? Don't we just delete the config at the end of the test? Or are you wondering if we test the sidecar as well?

tests/e2e/utils/helpers.go
```diff
@@ -94,6 +94,44 @@ func SubnetConfigExists(subnetName string) (bool, error) {
 	return gen && sc, nil
 }
 
+func UpdateSubnetConfig(subnetName string, network models.Network, subnetID string) error {
```
dasconnor, line 97: Could you rename this to `AddSubnetIDToSidecar` or something more specific. The name is a little misleading and vague.

dasconnor (commented):
Almost there, just a couple small points

cmd/subnetcmd/stats.go
holisticode: yep that's what it's doing. Ok adding comment.

tests/e2e/utils/helpers.go
holisticode: done

holisticode:
https://github.com/ava-labs/avalanche-cli/pull/287#discussion_r998437457

The comment tries to say that the underlying network for the `fuji` simulation is most probably a `Local Network`. Therefore, the `app` object of the e2e tests stores the generated subnetID into the `local` section, not the `fuji` section of the sidecar. 

So in order for the `stats` e2e test for `fuji` to be successful, we need to add the subnetID "manually" into the `fuji` section of the e2e test sidecar file so that the `stats` command can find it.

dasconnor (approved):
LGTM

PR #314, opened 2022-10-04
Fixed bad airdrop unit bug

### LINKED ISSUE #309: Wrong units
We are using 1 avax = 1e9 when it should be 1e18.
https://github.com/ava-labs/avalanche-cli/blob/main/pkg/vm/evmSettings.go#L43

dasconnor:
Fixed wrong units bug in our airdrop logic. I also added a test to check this.

Closes #309

### DIFF
--- a/pkg/vm/airdrop.go
+++ b/pkg/vm/airdrop.go
@@ -14,9 +14,9 @@ import (
 )
 
 const (
-	defaultAirdrop        = "Airdrop 1 million tokens to the default address (do not use in production)"
-	customAirdrop  string = "Customize your airdrop"
-	extendAirdrop         = "Would you like to airdrop more tokens?"
+	defaultAirdrop = "Airdrop 1 million tokens to the default address (do not use in production)"
+	customAirdrop  = "Customize your airdrop"
+	extendAirdrop  = "Would you like to airdrop more tokens?"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/vm/airdrop.go
```diff
@@ -13,6 +13,12 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
+const (
+	defaultAirdrop        = "Airdrop 1 million tokens to the default address (do not use in production)"
+	customAirdrop  string = "Customize your airdrop"
```
felipemadero, line 18: type seems not needed

pkg/vm/airdrop.go
```diff
@@ -13,6 +13,12 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
+const (
```
felipemadero, line 16: not sure is this should be module level or function level consts

pkg/vm/airdrop.go
dasconnor: My editor was complaining about a type checking thing, but I think it fixed itself so I removed this.

pkg/vm/airdrop.go
dasconnor: I needed to do this to mock the prompt in the test.

felipemadero (approved):
LGTM

Issue #333, opened 2022-10-11
Export Upgrade Bytes


dasconnor:
After creating upgrade bytes in #331 and #332, we will need to be able to export them to a file for distribution to other validators.

Currently blocked on #331 and #332.

holisticode:
Unblocked as #331 and #332 have both at least first iterations

PR #367, opened 2022-10-27
Export upgrade bytes

holisticode:
Closes https://github.com/ava-labs/avalanche-cli/issues/333


cmd/subnetcmd/upgradecmd/export_upgrade.go
```diff
@@ -0,0 +1,79 @@
...
+	cmd := &cobra.Command{
+		Use:   "export [subnetName]",
+		Short: "Export the upgrade bytes file to a location of choice on disk",
+		Long:  `Export the upgrade bytes file to a location of choice on disk`,
+		RunE:  upgradeExportCmd,
+		Args:  cobra.ExactArgs(1),
+	}
+
+	cmd.Flags().StringVar(&upgradeBytesFilePath, upgradeBytesFilePathKey, "", "Export upgrade bytes file to location of choice on disk")
+	if err := cmd.MarkFlagRequired(upgradeBytesFilePathKey); err != nil {
```
dasconnor, line 28: Let's not use required flags. Prompt for the location instead if it's not provided.

cmd/subnetcmd/upgradecmd/export_upgrade.go
```diff
@@ -0,0 +1,79 @@
...
+	subnetName := args[0]
+	if !app.GenesisExists(subnetName) {
+		ux.Logger.PrintToUser("The provided subnet name %q does not exist", subnetName)
+		return nil
+	}
+
+	if _, err := os.Stat(upgradeBytesFilePath); err == nil {
+		ux.Logger.PrintToUser("The file specified with path %q already exists!", upgradeBytesFilePath)
+
+		yes, err := app.Prompt.CaptureYesNo("Should we overwrite it?")
```
dasconnor, line 45: maybe add a `--force` flag

cmd/subnetcmd/upgradecmd/export_upgrade.go
```diff
@@ -0,0 +1,79 @@
...
+		if err != nil {
+			return err
+		}
+		if !yes {
+			ux.Logger.PrintToUser("Aborted by user. Nothing has been exported")
+			return nil
+		}
+	}
+
+	subnetPath := filepath.Join(app.GetUpgradeFilesDir(), subnetName)
```
dasconnor, line 55: Should we push this whole path generation function into `app`?

dasconnor (commented):
I think you actually want to merge this into `332-import-upgrade`

cmd/subnetcmd/upgradecmd/export_upgrade.go
holisticode: I will reuse the `pkg/subnet/upgrades` package for this

pkg/subnet/upgrades/upgrade_file.go
```diff
@@ -47,3 +48,22 @@ func WriteUpgradeFile(jsonBytes []byte, subnetName, upgradeFilesDir string) erro
 	ux.Logger.PrintToUser("File written successfully")
 	return nil
 }
+
+func ReadUpgradeFile(subnetName, upgradeFilesDir string) ([]byte, error) {
```
dasconnor, line 52: Why don't you get the `upgradesFilesDir` directly from `app` instead of asking for it as an argument? I think it will be the same argument everywhere.

holisticode: `app` is not injected into this package so both are equivalent

dasconnor (approved):

PR #368, opened 2022-10-27
Print upgrade bytes

holisticode:


cmd/subnetcmd/upgradecmd/print_upgrade.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+	"github.com/ava-labs/avalanchego/utils/storage"
+	"github.com/spf13/cobra"
+)
+
+// avalanche subnet upgrade import
+func newUpgradePrintCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "print [subnetName]",
```
dasconnor, line 22: Do we like `print` or `describe`?

cmd/subnetcmd/upgradecmd/print_upgrade.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func upgradePrintCmd(cmd *cobra.Command, args []string) error {
+	subnetName := args[0]
+	if !app.GenesisExists(subnetName) {
+		ux.Logger.PrintToUser("The provided subnet name %q does not exist", subnetName)
+		return nil
+	}
+
+	subnetPath := filepath.Join(app.GetUpgradeFilesDir(), subnetName)
```
dasconnor, line 39: Might want to make this whole filepath constructing process part of `app` like I mentioned in the other PR.

cmd/subnetcmd/upgradecmd/print_upgrade.go
```diff
@@ -0,0 +1,61 @@
...
+		Short: "Print the upgrade.json file content",
+		Long:  `Print the upgrade.json file content`,
+		RunE:  upgradePrintCmd,
+		Args:  cobra.ExactArgs(1),
+	}
+
+	return cmd
+}
+
+func upgradePrintCmd(cmd *cobra.Command, args []string) error {
```
dasconnor, line 32: One question I have, should we be supporting one set of upgrade bytes per network? Should we be able to have one set of upgrade bytes for fuji, one for local, and one for mainnet? Probably best to address that in your first PR or a follow up, but if you did add that it would require a flag/prompt here to pick a network.

dasconnor (commented):
Change the base branch to `333-export-upgrade`

cmd/subnetcmd/upgradecmd/print_upgrade.go
holisticode: In this case I think `print` is correct, as we are just printing the actual file contents to the screen.

cmd/subnetcmd/upgradecmd/print_upgrade.go
holisticode: Yep, just reusing the `upgrades.ReadFile` func from #333

cmd/subnetcmd/upgradecmd/print_upgrade.go
holisticode: Hmmm good question. We probably should, although I think the user is doing herself a disservice if she's keeping different versions for the same subnet - can quickly become pretty messy (in the sense that she really needs to know what she's doing).

If we do that, then all the previous cmds need this flag.

dasconnor: Let's address this in another ticket. My thinking is that you want to test any updates you make locally and then on Fuji before you push them to mainnet. Kind of like a pipeline.

dasconnor: Created #372 for this.

dasconnor (approved):
This is ready if you fix the lint

PR #390, opened 2022-11-01
Prevent subnet deploy if tx filepath exists

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/391


dasconnor (commented):
Do we not want to allow the ability to overwrite?

pkg/prompts/prompts.go
```diff
@@ -128,6 +129,13 @@ func validateExistingFilepath(input string) error {
 	return errors.New("file doesn't exist")
 }
 
+func validateNewFilepath(input string) error {
+	if _, err := os.Stat(input); err != nil && os.IsNotExist(err) {
+		return nil
+	}
+	return errors.New("file does exist")
```
dasconnor, line 136: `file already exists`

dasconnor (approved):

PR #399, opened 2022-11-05
Create subnet config and migrate top-level dirs

### LINKED ISSUE #376: Allow user to pass subnet config.json on subnet creation and subnet upgrade
- user should be able to pass a subnet config.json file on subnet create, and that will be saved with the other info of the subnet
- user should be able to pass new subnet config.json file using subnet upgrade
- subnet config.json should be used for local and public deploys
- if vm is known, probably check that json config format against the vm struct for it, if not, just check for valid json
- user should be able to get the subnet config.json with an export command

holisticode:
Closes #376


cmd/migrations.go
```diff
@@ -0,0 +1,93 @@
...
+		// instead of checking if it already exists, just let's try to create the dir
+		// if it already exists this will not return an error
+		if err := os.MkdirAll(newDir, constants.DefaultPerms755); err != nil {
+			return err
+		}
+		fileBytes, err := os.ReadFile(m)
+		if err != nil {
+			return fmt.Errorf("failed to read original file in `migrateTopLevelFiles` migration: %w", err)
+		}
+		if err := os.WriteFile(filepath.Join(newDir, suffix), fileBytes, constants.DefaultPerms755); err != nil {
```
dasconnor, line 75: Use `os.rename()`?

cmd/migrations.go
```diff
@@ -0,0 +1,93 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package cmd
```
dasconnor, line 3: I'm not sure this is the right package since this isn't really a command. Maybe make a migrations package in `internal`?

cmd/subnetcmd/configure.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/spf13/cobra"
+)
+
+// avalanche subnet configure
+func newConfigureCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "configure [subnetName]",
+		Short:        "",
```
dasconnor, line 21: Needs description

cmd/subnetcmd/configure.go
```diff
@@ -0,0 +1,103 @@
...
+	chains, err := validateSubnetNameAndGetChains(args)
+	if err != nil {
+		return err
+	}
+	subnetName := chains[0]
+
+	const (
+		subnetConf = "Subnet config"
+		chainConf  = "Chain config"
+		abort      = "Abort"
```
dasconnor, line 40: We don't usually provide this as an option, we usually just expect the user to ctrl + c

cmd/subnetcmd/configure.go
```diff
@@ -0,0 +1,103 @@
...
+	)
+
+	ux.Logger.PrintToUser("The " + logging.Cyan.Wrap("subnet") + logging.Reset.Wrap(" config file applies to *all* VMs in a subnet"))
```
dasconnor, line 43: I'm not sure we want these messages right now since we only support one chain per subnet. It might just confuse people

cmd/subnetcmd/configure.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func updateConf(subnet, filename string) error {
+	path, err := app.Prompt.CaptureExistingFilepath("Where can we find the configuration file?")
```
dasconnor, line 68: `Enter the path to your config file

cmd/subnetcmd/configure.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return err
+	}
+	fileBytes, err := validateJSON(path)
+	if err != nil {
+		return err
+	}
+	subnetDir := filepath.Join(app.GetBaseDir(), subnet)
+	if err := os.MkdirAll(subnetDir, constants.DefaultPerms755); err != nil {
```
dasconnor, line 77: shouldn't this directory already exist because the subnet exists?

cmd/subnetcmd/configure.go
```diff
@@ -0,0 +1,103 @@
...
+	fileName := filepath.Join(subnetDir, filename)
+	if err := os.WriteFile(fileName, fileBytes, constants.DefaultPerms755); err != nil {
+		return err
+	}
+	ux.Logger.PrintToUser("File %s successfully written", fileName)
+
+	return nil
+}
+
+func validateJSON(path string) ([]byte, error) {
```
dasconnor, line 89: This should probably be a more generic utils function in a different package.

cmd/subnetcmd/configure.go
```diff
@@ -0,0 +1,103 @@
...
+		Long:         ``,
+		SilenceUsage: true,
+		RunE:         configure,
+		Args:         cobra.ExactArgs(1),
+	}
+	return cmd
+}
+
+func configure(cmd *cobra.Command, args []string) error {
+	chains, err := validateSubnetNameAndGetChains(args)
```
dasconnor, line 31: We should validate the allowed characters in our subnet name and make sure none of them aren't illegal path characters like `/` or `?`.

cmd/migrations.go
```diff
@@ -0,0 +1,93 @@
...
+		migRunning = false
+	}
+	return nil
+}
+
+// Up to version 1.0.0 the sidecar and genesis files were stored at
+// {baseDir} in the top-level.
+// Due to new requirements and evolution of the tool, we now store
+// every subnet-specific file in {baseDir}/subnets/{subnetName}
+func migrateTopLevelFiles() error {
```
dasconnor, line 43: Can you write a test for this migration?

cmd/subnetcmd/list.go
```diff
@@ -45,7 +46,7 @@ func listSubnets(cmd *cobra.Command, args []string) error {
 	table.SetAutoMergeCells(true)
 	table.SetRowLine(true)
 
-	files, err := os.ReadDir(app.GetBaseDir())
+	subnets, err := os.ReadDir(filepath.Join(app.GetBaseDir(), constants.SubnetDir))
```
dasconnor, line 49: should we make this an `app` function? `app.GetAllSubnets()`?

cmd/subnetcmd/list.go
```diff
@@ -75,47 +76,58 @@ func listSubnets(cmd *cobra.Command, args []string) error {
...
+				if _, ok := deployedNames[sc.Subnet]; ok {
+					deployedLocal = "Yes"
+				}
+				deployedFuji := "No"
+				if _, ok := sc.Networks[models.Fuji.String()]; ok {
+					if sc.Networks[models.Fuji.String()].SubnetID != ids.Empty {
+						deployedFuji = "Yes"
+					}
 				}
+				deployedMain := "N/A"
```
dasconnor, line 119: Should we fix this to match Fuji?

tests/e2e/utils/helpers.go
```diff
@@ -58,7 +59,7 @@ func GetAPMDir() string {
 }
 
 func genesisExists(subnetName string) (bool, error) {
```
dasconnor, line 61: Can we add an E2E test for the chain config?

dasconnor (commented):
Nice work!

So this PR can create both a subnet config and a chain config, but it can only deploy the chain config, is that right?

cmd/migrations.go
holisticode: Way better! Wasn't aware this existed. Thanks.

cmd/migrations.go
holisticode: Agreed.Placing the README there.

cmd/subnetcmd/configure.go
holisticode: Of course

cmd/subnetcmd/configure.go
holisticode: But at the same time we are asking the user to select exactly those things - either subnet or chain config. If she is confused by the explication messages, then she shouldn't be using this command at all in my opinion. I vote to keep.

cmd/subnetcmd/configure.go
holisticode: It should. However, this cmd is now independent. It should never happen but it's just an additional double stitch...

cmd/subnetcmd/configure.go
holisticode: We have `binutils` and a `txutils` package, but no general `utils` package, so I thought it'd be maybe overkill to put just this there.

But OK then, let's create a new one

cmd/subnetcmd/configure.go
holisticode: Not sure I understand. Aren't we doing this already in `checkInvalidSubnetNames` which is checked in `validateSubnetNameAndGetChains`?

cmd/migrations.go
holisticode: I guess I should :smile:

cmd/subnetcmd/list.go
holisticode: Ah I get that slipped through the cracks when we did the mainnet update

cmd/subnetcmd/list.go
holisticode: We actually have `GetSidecarNames` there, which does nearly the same - instead of returning all subdirs of `{cli-root}/subnets`, "just" those which actually have a sidecar inside - which should be the same set really.

cmd/subnetcmd/configure.go
```diff
@@ -17,9 +16,11 @@ import (
 // avalanche subnet configure
 func newConfigureCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:          "configure [subnetName]",
-		Short:        "",
-		Long:         ``,
+		Use:   "configure [subnetName]",
+		Short: "Allows to add additional config files for the avalanchego nodes",
```
dasconnor, line 20: `Allows to add` -> `Adds`

cmd/subnetcmd/configure.go
```diff
@@ -17,9 +16,11 @@ import (
...
+		Long: `Avalanchego nodes can be configured at different levels. 
+For example, subnets have their own subnet config (applies to all chains/VMs in the subnet).
+And each chain or VM can have its own specific chain config file.`,
```
dasconnor, line 23: Need to add that this command allows you to set both the subnet config and the chain config

cmd/subnetcmd/configure.go
dasconnor: Do they share the same settings? I thought they were more or less totally different?

internal/migrations/README.md
```diff
@@ -0,0 +1,35 @@
+#Migrations
```
dasconnor, line 1: Some markdown viewers require a space between the `#` and the text to render a header properly.

internal/migrations/README.md
```diff
@@ -0,0 +1,35 @@
...
+##General structure
+* The application calls all migrations implemnted when booting 
+* Each migration is iterated in order (the order is "enforced" via the index in the migrations map)
+* Each migration checks itself if it needs to be applied
+* The first migration which is getting applied prints a message to the user
+* At the end of the iteration, it is checked if any migration ran. If one did, a closing message is printed
+* If no migration ran at all, nothing is being printed
+
+##Implementation
+* migrations need to reside in `internal/migrations`
```
dasconnor, line 28: nit: capitalize `migrations`

internal/migrations/migrations_test.go
```diff
@@ -0,0 +1,112 @@
...
+	type migTest struct {
+		migs           map[int]migrationFunc
+		name           string
+		shouldErr      bool
+		expectedOutput string
+	}
+
+	expectedIfRan := runMessage + "\n" + endMessage + "\n"
+
+	tests := []migTest{
```
dasconnor, line 37: what about 2 migrations, first one errors, then 2 migrations, second one errors

pkg/statemachine/stateMachine.go
```diff
@@ -21,12 +21,14 @@ const notRunningState = ""
 //
 // usage example:
 //
-//   machine, err := NewStateMachine(states)
-//   [ err processing ]
-//   while machine.Running() {
-//     [do stuff based on machine.CurrentState(), and set direction]
-//     machine.NextState(direction)
-//   }
+// TODO: This is currently issuing linter errors if not formatted this way
```
dasconnor, line 24: I actually was able to fix this in my PR. If you run `gofmt -s -w pkg/` it will fix it nicely.

dasconnor: This passed: 
```
// usage example:
//
//	machine, err := NewStateMachine(states)
//	[ err processing ]
//	while machine.Running() {
//	  [do stuff based on machine.CurrentState(), and set direction]
//	  machine.NextState(direction)
//	}
```

dasconnor: golangci-lint has version 1.50.1 built from 8926a95 on 2022-10-22T10:48:48Z

tests/e2e/testcases/packageman/suite.go
```diff
@@ -144,7 +144,9 @@ var _ = ginkgo.Describe("[Package Management]", func() {
 
 		// check avago install
 		gomega.Expect(utils.CheckAvalancheGoExists(avagoVersion1)).Should(gomega.BeTrue())
-		gomega.Expect(utils.CheckAvalancheGoExists(avagoVersion2)).Should(gomega.BeFalse())
+		// FIXME: If we have to use twice the same version (because they are incompatible
```
dasconnor, line 147: Can you add this to the E2E cleanup ticket?

tests/e2e/testcases/subnet/local/suite.go
```diff
@@ -163,4 +163,47 @@ var _ = ginkgo.Describe("[Local Subnet]", func() {
 		commands.DeleteSubnetConfig(subnetName)
 		commands.DeleteSubnetConfig(secondSubnetName)
 	})
+
+	/*
```
dasconnor, line 167: Did you mean to comment this?

pkg/statemachine/stateMachine.go
holisticode: OK. I suggest to leave this as-is now to have it pass, then we merge your PR on top, and hopefully that would overwrite / fix this?

tests/e2e/testcases/packageman/suite.go
holisticode: https://github.com/ava-labs/avalanche-cli/issues/420#issuecomment-1314287382

internal/migrations/migrations_test.go
holisticode: done

tests/e2e/utils/helpers.go
holisticode: done

cmd/subnetcmd/configure.go
dasconnor: I still think we should comment this out. I think it just adds confusion.

internal/migrations/migrations.go
```diff
@@ -0,0 +1,67 @@
...
+type migrationRunner struct {
+	showMsg    bool
+	running    bool
+	migrations map[int]migrationFunc
+}
+
+var (
+	runMessage       = "The tool needs to apply some internal updates first..."
+	endMessage       = "Update process successfully completed"
+	failedEndMessage = "Sadly some updates succeeded - others failed. Check output for hints"
```
dasconnor, line 23: Super nit, but I would remove the sadly.

dasconnor (approved):
LGTM when all convos are resolved.

Issue #398, opened 2022-11-04
Generate a ton of addresses and store them internally 


holisticode:
Generate a ton of addresses and store them internally  so if the user gives a random P chain address from web wallet it will know which one it is

kanatliemre:
Currently the CLI seems to only recognize address at index `0` at `m/44'/9000'/0'/0/0`
Web wallet uses dynamic addresses so the address can be any index `n` at `m/44'/9000'/0'/0/n`
The CLI should derive a bunch of addresses for these `n` values and when requested to sign with a given address it should compare agains the list and use the correct index

Issue #346, opened 2022-10-18
enable using any ledger address


felipemadero:
currently we are using only the first ledger address. 
for this issue we may either enable using the first N ledger-derived addresses, or use specific addresses of the ledger

second option may require also a chance in avago wallet ledger api. the idea may be to have exact control on which ledger address/addresses to use for tx creation/signing. a possible change in avago wallet ledger api:
instead of: NewLedgerKeychain(l ledger.Ledger, numToDerive int) so the wallet uses a keychain with ledger addresses derived from 0 until numToDerive, and can build txs and sign with any of them
use: do NewLedgerKeychain(l ledger.Ledger, derivationNumbers []int)  so the wallet uses a keychain with the given ledger addresses (given using the derivation numbers), and use any of them for tx build/sign

dasconnor:
This is done

PR #408, opened 2022-11-09
Choose ledger addresses

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/398 https://github.com/ava-labs/avalanche-cli/issues/346
With this PR, the user can specify a list of addresses to use on deploy, addValidator and sign, with flag --ledger-addrs.
Also, if not specified, a prompt enables to select between addr at index 0 and other custom addr given by ther user.
In both cases, the given addresses are checked agains the first 1000 ledger derived addresses to find the correct indices.
Currently depends on avalanchego-internal for lint+test purposes


cmd/subnetcmd/addValidator.go
```diff
@@ -59,6 +58,8 @@ Testnet or Mainnet.`,
 	cmd.Flags().BoolVar(&deployMainnet, "mainnet", false, "join on `mainnet`")
 	cmd.Flags().StringSliceVar(&subnetAuthKeys, "subnet-auth-keys", nil, "control keys that will be used to authenticate add validator tx")
 	cmd.Flags().StringVar(&outputTxPath, "output-tx-path", "", "file path of the add validator tx")
+	cmd.Flags().BoolVarP(&useLedger, "ledger", "g", false, "use ledger instead of key (always true on mainnet, defaults to false on fuji)")
```
dasconnor, line 61: will need to update the reference in the docs

dasconnor (approved):
I think this looks good, we just need to create a little how to guide for how to use this. I'm not 100% what the best way to describe it is.

Issue #365, opened 2022-10-27
add support for mainnet, local networks, ledger to key list


felipemadero:
- probably the specific network should be selected, instead of all-networks
- ledger flag maybe should output information relative to currently connected ledger

PR #412, opened 2022-11-10
Show ledger addrs on key list

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/365

- Add --ledger flag that receives a list of indices and shows information for those, avoiding providing information for stored keys
- Add flags to control networks and chains: --local, --testnet, --fuji, --mainnet, --cchain
- Refactor of all code

### DIFF
--- a/cmd/keycmd/list.go
+++ b/cmd/keycmd/list.go
@@ -10,16 +10,16 @@ import (
 	"path/filepath"
 	"strings"
 
-	ledger "github.com/ava-labs/avalanche-ledger-go"
 	"github.com/ava-labs/avalanche-cli/pkg/constants"
 	"github.com/ava-labs/avalanche-cli/pkg/key"
 	"github.com/ava-labs/avalanche-cli/pkg/models"
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+	ledger "github.com/ava-labs/avalanche-ledger-go"
 	"github.com/ava-labs/avalanchego/ids"
 	avago_constants "github.com/ava-labs/avalanchego/utils/constants"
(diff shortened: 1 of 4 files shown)


cmd/keycmd/list.go
```diff
@@ -24,107 +26,351 @@ import (
...
+			return []addressInfo{}, err
+		}
+		addrInfos = append(addrInfos, ledgerAddrInfos...)
+	}
+	return addrInfos, nil
+}
 
-	if allNetworks {
-		supportedNetworks[models.Local.String()] = 0
+func getLedgerAddrInfo(
```
dasconnor, line 278: I'm not sure what the difference between this function and `getLedgerAddrInfos` is. The difference isn't clear from the name.

dasconnor (commented):
I think the code looks good, but could you include a video or some screenshots? It's a little hard for me to visualize what everything here looks like to the end user.

felipemadero:
> I think the code looks good, but could you include a video or some screenshots? It's a little hard for me to visualize what everything here looks like to the end user.

https://asciinema.org/a/537864

cmd/keycmd/list.go
felipemadero: renamed to getLedgerIndices info and getLedgerIndexInfo

dasconnor:
That video was great! Very helpful! I do have a couple recommendations. I'm not sure I like `mainnet` being the default network. I think we probably want to prompt for network by default.

Also, I don't like having to set `--mainnet=false`. If no flags, we should prompt. Then any flag can just be additive. So you can do `--mainnet` and `--testnet` and it will be fine.

dasconnor (approved):

PR #415, opened 2022-11-10
Fix Upgrade Compatibility

### LINKED ISSUE #377: Don't break Avalanche-CLI on avalanchego update
# Problem 

When a new version of Avalanchego releases, it may break compatibility with all previous versions of subnet-evm and spaces-vm. It may be hours or days before the appropriate version is released. In that in between window, avalanche-cli is broken by default and users must pass the `--avalanchego-version` flag. We need a scheme to prevent this from happening.

# Proposed Solution

Create a new API endpoint under `info`. It will list the avalanchego versions compatible with a given `rpcChainVMProtocolVersion`.

```
GET info.getVersionCompatibility

"params": {
  "rpcChainVmProtocolVersion": 17
}

Returns 
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "versions": ["v1.9.0", "v1.9.1"]
  }
}
```

## CLI Experience

1. At deploy time, CLI will check the `rpcChainVMProtocolVersion` of the desired VM.
2. The CLI will perform a lookup at api.avax.network to determine avago compatibility for the desired `rpcChainVMProtocolVersion`.
3. The CLI will launch with the latest avago from the compatibility list.

## Avago Release Experience

- When performing a release, the avago team will be responsible for adding the new release's compatibility into some config file prior to release

dasconnor:
This PR addresses the CLI compatibility breaks whenever Avalanchego updates. It now queries compatibility files hosted in the avalanchego and VM repos. It assumes that you are deploying a fixed VM version. First, we look up what RPC protocol version that VM uses, then we find the latest version of avalanchego compatible with that RPC version. That version of Avalanchego is the one we use for our deployment.

Closes #377

## Sumary of Changes
- Bump linter version
- Add RPCVersion to sidecar and track on subnet creation
- Select a valid version of avalanchego for the specified VM RPCVersion at deploy time
  - If you're running a custom VM, we don't track the RPCVersion and just give you latest
- Moved some of our http code to a new `Downloader` interface as part of the `application` package. I did this to facilitate testing by making it easier to mock network requests
- Created the `network` package to gather info about the currently running local network. Made it an interface to make it mockable.
- Fixed the spacing for the linter.
- Added code to fetch the current rpc version for a given VM version.
- E2E tests now use latest avago version by default
- Fixed package management tests to use a newer version
- Added a new compatibility E2E suite


pkg/application/downloader.go
```diff
@@ -0,0 +1,78 @@
...
+		return "", fmt.Errorf("failed to get latest binary version from %s: %w", releaseURL, err)
+	}
+
+	var jsonStr map[string]interface{}
+	if err := json.Unmarshal(jsonBytes, &jsonStr); err != nil {
+		return "", fmt.Errorf("failed to unmarshal binary json version string: %w", err)
+	}
+
+	version := jsonStr["tag_name"].(string)
+	if version == "" || version[0] != 'v' {
```
holisticode, line 73: validate with semver?

pkg/application/downloader.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get latest binary version from %s: %w", releaseURL, err)
+	}
+
+	var jsonStr map[string]interface{}
+	if err := json.Unmarshal(jsonBytes, &jsonStr); err != nil {
+		return "", fmt.Errorf("failed to unmarshal binary json version string: %w", err)
+	}
+
+	version := jsonStr["tag_name"].(string)
```
holisticode, line 72: This could probably also go into a constant

pkg/application/downloader.go
```diff
@@ -0,0 +1,78 @@
...
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected http status code: %d", resp.StatusCode)
+	}
+
+	return io.ReadAll(resp.Body)
+}
+
+// GetLatestReleaseVersion returns the latest available version from github
+func (downloader) GetLatestReleaseVersion(releaseURL string) (string, error) {
```
holisticode, line 41: A suggestion just for you to consider and decide: I think we are already using different downloaders, maybe we could have this be just a real downloader, without any "business" logic, downloading from the supplied URL, and then do version checks and stuff in a wrapper?

holisticode: Oh I see, this is basically just moved around. So we don't have maybe that many Downloaders after all.

pkg/network/network.go
```diff
@@ -0,0 +1,43 @@
...
+	GetCurrentNetworkVersion() (string, int, error)
+}
+
+type networkStatusChecker struct{}
+
+func NewStatusChecker() StatusChecker {
+	return networkStatusChecker{}
+}
+
+func (networkStatusChecker) GetCurrentNetworkVersion() (string, int, error) {
```
holisticode, line 26: If the `string` returned is a semver version, maybe we could use that as the data type?

cmd/subnetcmd/deploy.go
```diff
@@ -678,3 +692,45 @@ func PrintDeployResults(chain string, subnetID ids.ID, blockchainID ids.ID, isFu
 	table.Render()
 	return nil
 }
+
+// Has side effects and will set the version of avalanchego in the package-level avagoVersion variable.
```
felipemadero, line 696: can we avoid the side effects? not so easy to follow flow

cmd/subnetcmd/deploy.go
```diff
@@ -678,3 +692,45 @@ func PrintDeployResults(chain string, subnetID ids.ID, blockchainID ids.ID, isFu
...
+
+	// RPC Version was made available in the info API in avalanchego version v1.9.2. For prior versions,
+	// we will need to skip this check.
+	skipRPCCheck := false
+	if semver.Compare(runningAvagoVersion, "v1.9.2") == -1 {
+		skipRPCCheck = true
+	}
+
+	if networkRunning {
+		if avagoVersion == "latest" {
```
felipemadero, line 715: probably will me more clear by using a more descriptive name, not just avagoVersion

pkg/vm/compatibility.go
```diff
@@ -0,0 +1,95 @@
...
+		return 0, errors.New("unknown VM type")
+	}
+
+	compatibilityBytes, err := app.Downloader.Download(url)
+	if err != nil {
+		return 0, err
+	}
+
+	var parsedCompat models.VMCompatibility
+	err = json.Unmarshal(compatibilityBytes, &parsedCompat)
```
holisticode, line 39: if err ....

pkg/vm/compatibility.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func GetLatestAvalancheGoByProtocolVersion(app *application.Avalanche, rpcVersion int) (string, error) {
+	compatibilityBytes, err := app.Downloader.Download(constants.AvalancheGoCompatibilityURL)
+	if err != nil {
+		return "", err
+	}
+
+	var parsedCompat models.AvagoCompatiblity
+	err = json.Unmarshal(compatibilityBytes, &parsedCompat)
```
holisticode, line 59: if err....

pkg/constants/constants.go
```diff
@@ -100,4 +100,8 @@ const (
 
 	NotAvailableLabel = "Not available"
 	BackendCmd        = "avalanche-cli-backend"
+
+	AvalancheGoCompatibilityURL  = "https://raw.githubusercontent.com/ava-labs/avalanchego/master/version/compatibility.json"
```
felipemadero, line 104: also add RPC here?

pkg/vm/compatibility.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	eligibleVersions, ok := parsedCompat[strconv.Itoa(rpcVersion)]
```
holisticode, line 64: Are these not semver versions? If they are, semver has inbuilt sorting if I remember correctly. Otherwise the sorting can actually get tricky.

pkg/vm/compatibility_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+const (
+	testAvagoVersion         = "v0.4.2"
+	testUnlistedAvagoVersion = "v0.4.3"
+)
+
+var (
+	testSubnetEVMCompat = []byte("{\"rpcChainVMProtocolVersion\": {\"v0.4.2\": 18,\"v0.4.1\": 18,\"v0.4.0\": 17}}")
+	testAvagoCompat     = []byte("{\"19\": [\"v1.9.2\"],\"18\": [\"v1.9.1\"],\"17\": [\"v1.9.0\",\"v1.8.0\"]}")
+	testAvagoCompat2    = []byte("{\"19\": [\"v1.9.2\", \"v1.9.1\"],\"18\": [\"v1.9.0\"]}")
```
holisticode, line 24: It might be interesting to have tests with more "tricky" versions, e.g. v1.9.11 vs v.1.9.1 vs 1.9.01 etc.

pkg/vm/compatibility_test.go
```diff
@@ -0,0 +1,161 @@
...
+		latestVersion   string
+		expectedVersion string
+		expectedErr     error
+	}
+
+	tests := []versionTest{
+		{
+			rpc:             19,
+			testData:        testAvagoCompat,
+			latestVersion:   "v1.9.2",
```
holisticode, line 96: Not sure we want to limit the tests to just these hardcoded versions?

tests/e2e/commands/subnet.go
```diff
@@ -151,7 +151,13 @@ func DeleteSubnetConfig(subnetName string) {
 // Returns the deploy output
 /* #nosec G204 */
 func DeploySubnetLocally(subnetName string) string {
-	return DeploySubnetLocallyWithArgs(subnetName, utils.AvagoVersion, "")
+	return DeploySubnetLocallyWithArgs(subnetName, "", "")
+}
+
+// Returns the deploy output
```
felipemadero, line 157: obsolete comment

tests/e2e/commands/subnet.go
```diff
@@ -199,6 +205,27 @@ func DeploySubnetLocallyWithArgs(subnetName string, version string, confPath str
 	return string(output)
 }
 
+// Returns the deploy output
```
felipemadero, line 208: same

tests/e2e/testcases/subnet/local/suite.go
```diff
@@ -164,3 +167,81 @@ var _ = ginkgo.Describe("[Local Subnet]", func() {
 		commands.DeleteSubnetConfig(secondSubnetName)
 	})
 })
+
+var _ = ginkgo.Describe("[Subnet Compatibility]", func() {
+	ginkgo.AfterEach(func() {
+		commands.CleanNetwork()
+		err := utils.DeleteConfigs(subnetName)
```
holisticode, line 174: pattern if err.... (not repeating)

tests/e2e/testcases/packageman/suite.go
```diff
@@ -20,10 +20,14 @@ const (
 	// is compatible with avago v1.8.x.
 	// This also means we should consider this for the future: How to handle hardforks which make
 	// this test break.:w
-	subnetEVMVersion1 = "v0.3.0"
-	subnetEVMVersion2 = "v0.3.0"
-	avagoVersion1     = "v1.8.0"
-	avagoVersion2     = "v1.8.1"
+	subnetEVMVersion1 = "v0.4.2"
```
felipemadero, line 23: prefix with soloAvago?

tests/e2e/testcases/subnet/local/suite.go
```diff
@@ -164,3 +167,81 @@ var _ = ginkgo.Describe("[Local Subnet]", func() {
...
+		gomega.Expect(err).Should(gomega.BeNil())
+		err = utils.DeleteConfigs(secondSubnetName)
+		if err != nil {
+			fmt.Println("Delete config error:", err)
+		}
+		gomega.Expect(err).Should(gomega.BeNil())
+	})
+
+	ginkgo.It("can deploy a subnet-evm with old version", func() {
+		subnetEVMVersion := "v0.4.2"
```
holisticode, line 187: Couldn't these be quickly get outdated if they are hardcoded?

tests/e2e/utils/helpers.go
```diff
@@ -257,6 +257,12 @@ func DeleteBins() error {
 	return nil
 }
 
+func DeleteCustomBinary(vmName string) {
+	vmPath := path.Join(GetBaseDir(), constants.VMDir, vmName)
+	// ignore error, file may not exist
+	os.RemoveAll(vmPath)
```
holisticode, line 263: Doesn't this result in a linter errror?

felipemadero (commented):
LGTM. Nits

pkg/vm/compatibility.go
dasconnor: These are not actually semvers, they are just ints. Ex: https://github.com/ava-labs/subnet-evm/blob/master/compatibility.json

cmd/subnetcmd/deploy.go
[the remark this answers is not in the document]
dasconnor: I removed the side effect

cmd/subnetcmd/deploy.go
```diff
@@ -678,3 +692,45 @@ func PrintDeployResults(chain string, subnetID ids.ID, blockchainID ids.ID, isFu
 	table.Render()
 	return nil
 }
+
+// Has side effects and will set the version of avalanchego in the package-level avagoVersion variable.
+func checkForInvalidDeployAndSetAvagoVersion(network nw.StatusChecker, configuredRPCVersion int) error {
```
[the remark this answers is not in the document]
dasconnor: renamed

cmd/subnetcmd/deploy_test.go
```diff
@@ -3,12 +3,25 @@
...
+	"github.com/ava-labs/avalanche-cli/pkg/application"
+	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 )
 
+const (
+	testAvagoVersion1      = "v1.9.2"
+	testAvagoVersion2      = "v1.9.1"
+	testLatestAvagoVersion = "latest"
```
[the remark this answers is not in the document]
dasconnor: Not sure what you're suggesting. We don't just test latest, but latest is a key value to test, so we alias it with a constant here.

pkg/application/downloader.go
```diff
@@ -0,0 +1,78 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+)
+
+type Downloader interface {
```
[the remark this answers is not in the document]
dasconnor: Done. As I think you figured out below, I moved a bunch of our http/downloading functionality into this package so we could mock it cleanly for testing.

pkg/network/network.go
dasconnor: I don't think there is a semver data type, so I don't think there's anything to do here.

pkg/vm/compatibility_test.go
dasconnor: This might be interesting for fuzzing purposes, but I don't think it's really necessary. We're really just testing key value matches. The one part that is interesting is that we're testing some semantic version sorting, which could backfire if some strings aren't semvars, but avalanchego does have protections to make sure that's not the case. So I don't think we really need to do anything else here.

pkg/vm/compatibility_test.go
dasconnor: This isn't actually downloading compatibility files or running any vms, we're just testing that we're parsing the compatibility formats correctly. Fixed versions are the only way to test this.

tests/e2e/testcases/subnet/local/suite.go
dasconnor: for this particular test, that's actually the point

tests/e2e/utils/helpers.go
dasconnor: nope! only if you assign the error to a variable I believe

cmd/subnetcmd/deploy.go
dasconnor: removed side effect

cmd/subnetcmd/deploy.go
dasconnor: renamed

pkg/constants/constants.go
dasconnor: not sure what you mean?

pkg/vm/compatibility_test.go
holisticode: Are we sure that v1.9.11 is after v1.9.1 in this PR? v1.10.1 after v1.9.1 ? If the comparison is by string, not necessarily.

tests/e2e/utils/helpers.go
holisticode: Oh interesting. Maybe just with my local linter settings.

pkg/network/network.go
holisticode: You're right, there isn't actually such a type

holisticode (approved):

pkg/constants/constants.go
felipemadero: both SubnetEVMRPCCompatibilityURL and SpacesVMRPCCompatibilityURL are explicit regarding compatibility refers to RPC version compatibility, but AvalancheGoCompatibilityURL not

felipemadero (approved):

pkg/vm/compatibility_test.go
dasconnor: this was indeed a bug, I added a test and switched the sorting to use `semver` instead of string sorting

PR #446, opened 2022-11-18
Fix missing flags for the configure command

holisticode:


cmd/subnetcmd/configure.go
```diff
@@ -36,33 +45,58 @@ func configure(cmd *cobra.Command, args []string) error {
...
+		if yes {
+			configsToLoad[other], err = app.Prompt.CaptureExistingFilepath("Enter the path to your configuration file")
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// load each provided file
+	for f, c := range configsToLoad {
```
dasconnor, line 90: could you use `file` and `config` instead of f and c, I was a little confused at first

dasconnor (approved):
one small nit

cmd/subnetcmd/configure.go
holisticode: Used `fileName` and `configPath`

PR #459, opened 2022-11-21
Fix repos bug

### LINKED ISSUE #458: Fix repos dir creation bug
If you call `subnet publish`, it fails with a directory doesn't exist error.

dasconnor:
Creates the `repos` directory in the root command if it doesn't exist.

Closes #458

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -148,6 +148,13 @@ func setupEnv() (string, error) {
 		os.Exit(1)
 	}
 
+	// Create repos dir if it doesn't exist
+	repoDir := filepath.Join(baseDir, constants.ReposDir)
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -148,6 +148,13 @@ func setupEnv() (string, error) {
 		os.Exit(1)
 	}
 
+	// Create repos dir if it doesn't exist
+	repoDir := filepath.Join(baseDir, constants.ReposDir)
```
holisticode, line 152: Shouldn't this also be subnet dependent now?

holisticode: OK let's pass this now as a fix and address later

holisticode (approved):

PR #460, opened 2022-11-22
Import deployed subnets from a running network into avalanche-cli

holisticode:
This PR is a first iteration allowing imports of deployed network, supporting public networks only for now.

There are a lot of possible edge cases possible, but this PR should be good to go for the standard use case where a subnet has been deployed with subnet-cli.

Custom VMs are not yet supported but will be available in the next iteration.

### DIFF
--- a/cmd/subnetcmd/import_from_api.go
+++ b/cmd/subnetcmd/import_from_api.go
@@ -11,12 +11,12 @@ import (
 	"github.com/ava-labs/avalanche-cli/pkg/constants"
 	"github.com/ava-labs/avalanche-cli/pkg/models"
 	"github.com/ava-labs/avalanche-cli/pkg/ux"
+	"github.com/ava-labs/avalanche-cli/pkg/vm"
 	"github.com/ava-labs/avalanchego/api/info"
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/rpc"
 	"github.com/ava-labs/avalanchego/vms/platformvm"
 	"github.com/ava-labs/coreth/core"
-	"github.com/ava-labs/spacesvm/vm"
 	"github.com/spf13/cobra"
 )
 
@@ -167,7 +167,7 @@ func importRunningSubnet(cmd *cobra.Command, args []string) error {
 		if ch.SubnetID == subnetID {
 			blockchainID = ch.ID
 			vmID = ch.VMID
-			subnetName = vm.Name
+			subnetName = ch.Name
 			break
 		}
 	}
@@ -219,6 +219,34 @@ func importRunningSubnet(cmd *cobra.Command, args []string) error {
 		Version: constants.SidecarVersion,
 	}
 
+	var versions []string
+	switch vmType {
+	case models.SubnetEvm:
+		versions, err = app.Downloader.GetAllReleasesForRepo(constants.AvaLabsOrg, constants.SubnetEVMRepoName)
+		if err != nil {
+			return err
+		}
+		sc.VMVersion, err = app.Prompt.CaptureList("Pick the version for this VM", versions)
+	case models.SpacesVM:
+		versions, err = app.Downloader.GetAllReleasesForRepo(constants.AvaLabsOrg, constants.SpacesVMRepoName)
+		if err != nil {
+			return err
+		}
+		sc.VMVersion, err = app.Prompt.CaptureList("Pick the version for this VM", versions)
+	case models.CustomVM:
+		// versions,err = prompts.Capt
+	default:
+		return fmt.Errorf("unexpected VM type: %v", vmType)
+	}
+
+	// hasn't been set in reply
+	if sc.RPCVersion == 0 {
+		sc.RPCVersion, err = vm.GetRPCProtocolVersion(app, vmType, sc.VMVersion)
+		if err != nil {
+			return fmt.Errorf("failed getting RPCVersion for VM type %s with version %s", vmType, sc.VMVersion)
+		}
+	}
+
 	if vmType == models.SubnetEvm {
 		var genesis core.Genesis
 		if err := json.Unmarshal(genBytes, &genesis); err != nil {
@@ -242,5 +270,11 @@ func importRunningSubnet(cmd *cobra.Command, args []string) error {
 		sc.RPCVersion = int(reply.RPCProtocolVersion)
 	}
 
-	return app.CreateSidecar(sc)
+	if err := app.CreateSidecar(sc); err != nil {
+		return fmt.Errorf("failed creating the sidecar for import: %w", err)
+	}
+
+	ux.Logger.PrintToUser("Subnet %s imported successfully", sc.Name)
+
+	return nil
 }
--- a/cmd/subnetcmd/import_from_api.go
+++ b/cmd/subnetcmd/import_from_api.go
@@ -234,7 +234,7 @@ func importRunningSubnet(cmd *cobra.Command, args []string) error {
 		}
 		sc.VMVersion, err = app.Prompt.CaptureList("Pick the version for this VM", versions)
 	case models.CustomVM:
-		// versions,err = prompts.Capt
+		return fmt.Errorf("importing custom VMs is not yet implemented, but will be available soon")
 	default:
 		return fmt.Errorf("unexpected VM type: %v", vmType)
 	}
--- a/cmd/subnetcmd/import_from_api.go
+++ b/cmd/subnetcmd/import_from_api.go
@@ -1,0 +1,246 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package subnetcmd
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+	"github.com/ava-labs/avalanche-cli/pkg/models"
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+	"github.com/ava-labs/avalanchego/api/info"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/rpc"
+	"github.com/ava-labs/avalanchego/vms/platformvm"
+	"github.com/ava-labs/coreth/core"
+	"github.com/ava-labs/spacesvm/vm"
+	"github.com/spf13/cobra"
+)
+
+var (
+	genesisFilePath string
+	subnetIDstr     string
+	nodeURL         string
+)
+
+// avalanche subnet import
+func newImportFromNetworkCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "import-running [subnetPath]",
+		Short:        "Import an existing subnet config from running subnets",
+		RunE:         importRunningSubnet,
+		SilenceUsage: true,
+		Args:         cobra.MaximumNArgs(1),
+		Long: `The subnet import command will import a subnet configuration from a running network.
+
+The genesis file should be available from the disk for this to work. 
+By default, an imported subnet will not overwrite an existing subnet with the same name. 
+To allow overwrites, provide the --force flag.`,
+	}
+
+	cmd.Flags().StringVar(&nodeURL, "node-url", "", "[optional] URL of an already running subnet validator")
+
+	cmd.Flags().BoolVar(&deployTestnet, "fuji", false, "import from `fuji` (alias for `testnet`)")
+	cmd.Flags().BoolVar(&deployTestnet, "testnet", false, "import from `testnet` (alias for `fuji`)")
+	cmd.Flags().BoolVar(&deployMainnet, "mainnet", false, "import from `mainnet`")
+	cmd.Flags().BoolVar(&useSubnetEvm, "evm", false, "import a subnet-evm")
+	cmd.Flags().BoolVar(&useSpacesVM, "spacesvm", false, "use the SpacesVM as the base template")
+	cmd.Flags().BoolVar(&useCustom, "custom", false, "use a custom VM template")
+	cmd.Flags().BoolVarP(
+		&overwriteImport,
+		"force",
+		"f",
+		false,
+		"overwrite the existing configuration if one exists",
+	)
+	cmd.Flags().StringVar(
+		&genesisFilePath,
+		"genesis-file-path",
+		"",
+		"path to the genesis file",
+	)
+	cmd.Flags().StringVar(
+		&subnetIDstr,
+		"subnet-id",
+		"",
+		"the subnet ID",
+	)
+	return cmd
+}
+
+func importRunningSubnet(cmd *cobra.Command, args []string) error {
+	var err error
+
+	var network models.Network
+	switch {
+	case deployTestnet:
+		network = models.Fuji
+	case deployMainnet:
+		network = models.Mainnet
+	}
+
+	if network == models.Undefined {
+		networkStr, err := app.Prompt.CaptureList(
+			"Choose a network to import from",
+			[]string{models.Fuji.String(), models.Mainnet.String()},
+		)
+		if err != nil {
+			return err
+		}
+		network = models.NetworkFromString(networkStr)
+	}
+
+	if genesisFilePath == "" {
+		genesisFilePath, err = app.Prompt.CaptureExistingFilepath("Provide the path to the genesis file")
+		if err != nil {
+			return err
+		}
+	}
+
+	var reply *info.GetNodeVersionReply
+
+	if nodeURL == "" {
+		yes, err := app.Prompt.CaptureYesNo("Have nodes already been deployed to this subnet?")
+		if err != nil {
+			return err
+		}
+		if yes {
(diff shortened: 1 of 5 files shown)


cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+	var reply *info.GetNodeVersionReply
+
+	if nodeURL == "" {
+		yes, err := app.Prompt.CaptureYesNo("Have nodes already been deployed to this subnet?")
+		if err != nil {
+			return err
+		}
+		if yes {
+			nodeURL, err = app.Prompt.CaptureString(
+				"Please provide an API URL of such a node so we can query its VM version (e.g. 111.22.33.44:5555)")
```
dasconnor, line 112: can you put an `http` or `https`  on that example?

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+		return err
+	}
+
+	var (
+		blockchainID, vmID ids.ID
+		subnetName         string
+	)
+
+	for _, ch := range chains {
+		// NOTE: This supports only one chain per subnet
```
dasconnor, line 165: Why don't we error if `len(chains) > 1`. That way we catch the error case instead of having unexpected behavior.

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+	if blockchainID == ids.Empty || vmID == ids.Empty {
+		return fmt.Errorf("subnet ID %s not found on this network", subnetIDstr)
+	}
+
+	ux.Logger.PrintToUser("Retrieved information. BlockchainID: %s, Name: %s, VMID: %s",
+		blockchainID.String(),
+		subnetName,
+		vmID.String(),
+	)
+	// TODO: it's probably possible to deploy VMs with the same name on a public network
```
dasconnor, line 183: ticket?

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+	// In this case, an import could clash because the tool supports unique names only
+
+	genBytes, err := os.ReadFile(genesisFilePath)
```
dasconnor, line 186: `application.CopyGenesisFile`

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+			[]string{models.SubnetEvm, models.SpacesVM, models.CustomVM},
+		)
+		if err != nil {
+			return err
+		}
+		vmType = models.VMTypeFromString(subnetTypeStr)
+	}
+
+	sc := &models.Sidecar{
+		Name:            subnetName,
```
dasconnor, line 208: you should set the token name to `TEST`

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+		}
+	}
+
+	if vmType == models.SubnetEvm {
+		var genesis core.Genesis
+		if err := json.Unmarshal(genBytes, &genesis); err != nil {
+			return err
+		}
+		sc.ChainID = genesis.Config.ChainID.String()
+		/*
```
dasconnor, line 258: No, it's not encoded in the genesis, we only track it for metamask purposes. Set above and delete here.

cmd/subnetcmd/import_from_api.go
dasconnor: Optionally you could query for the token name.

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+		}
+		sc.VMVersion, err = app.Prompt.CaptureList("Pick the version for this VM", versions)
+	case models.SpacesVM:
+		versions, err = app.Downloader.GetAllReleasesForRepo(constants.AvaLabsOrg, constants.SpacesVMRepoName)
+		if err != nil {
+			return err
+		}
+		sc.VMVersion, err = app.Prompt.CaptureList("Pick the version for this VM", versions)
+	case models.CustomVM:
+		return fmt.Errorf("importing custom VMs is not yet implemented, but will be available soon")
```
dasconnor, line 236: ticket

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+	default:
+		return fmt.Errorf("unexpected VM type: %v", vmType)
+	}
+	if err != nil {
+		return err
+	}
+
+	// hasn't been set in reply
+	if sc.RPCVersion == 0 {
```
dasconnor, line 245: I don't think you need this if. You don't set it anywhere above. It's always true.

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+		sc.RPCVersion, err = vm.GetRPCProtocolVersion(app, vmType, sc.VMVersion)
+		if err != nil {
+			return fmt.Errorf("failed getting RPCVersion for VM type %s with version %s", vmType, sc.VMVersion)
+		}
+	}
+
+	if vmType == models.SubnetEvm {
```
dasconnor, line 252: I think you could do the same thing for spaces and its genesis format.

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+		}
+		sc.ChainID = genesis.Config.ChainID.String()
+		/*
+			  // TODO: can we get this one?
+				TokenName       string
+		*/
+	}
+
+	vmIDstr := vmID.String()
+	sc.ImportedVMID = vmIDstr // TODO: Is this correct?
```
dasconnor, line 265: please resolve this. It looks right to me, but double check

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+	if reply != nil {
+		for _, v := range reply.VMVersions {
+			if v == vmIDstr {
+				sc.VMVersion = v
```
dasconnor, line 269: nit, but you can add a `break` statement here

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+		/*
+			  // TODO: can we get this one?
+				TokenName       string
+		*/
+	}
+
+	vmIDstr := vmID.String()
+	sc.ImportedVMID = vmIDstr // TODO: Is this correct?
+	if reply != nil {
+		for _, v := range reply.VMVersions {
```
dasconnor, line 267: If you're going to do this check, you don't need to select the version around line 223

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+			},
+		},
+		Subnet:  subnetName,
+		Version: constants.SidecarVersion,
+	}
+
+	var versions []string
+	switch vmType {
+	case models.SubnetEvm:
+		versions, err = app.Downloader.GetAllReleasesForRepo(constants.AvaLabsOrg, constants.SubnetEVMRepoName)
```
dasconnor, line 224: This is a serious question. I'm not implying the answer is yes. Is this a better approach than just asking the user to enter the version with text? That's currently what we do in the creation flow.

dasconnor: If we decide yes, then we should add this into the creation flow.

pkg/application/downloader.go
```diff
@@ -43,13 +44,40 @@ func (downloader) Download(url string) ([]byte, error) {
...
+		if !semver.IsValid(version) {
+			return nil, fmt.Errorf("invalid version string: %s", version)
+		}
+		releases[i] = version
+	}
+
+	return releases, nil
+}
+
+func (downloader) doAPIRequest(url string) (io.ReadCloser, error) {
```
dasconnor, line 77: is this github specific? If so, consider renaming

pkg/prompts/prompts.go
```diff
@@ -110,6 +111,11 @@ func validateTime(input string) error {
 	return err
 }
 
+func validateID(input string) error {
```
dasconnor, line 114: I think we moved our validation functions to a different file.

cmd/subnetcmd/import_from_api.go
```diff
@@ -0,0 +1,282 @@
...
+var (
+	genesisFilePath string
+	subnetIDstr     string
+	nodeURL         string
+)
+
+// avalanche subnet import
+func newImportFromNetworkCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "import-running [subnetPath]",
```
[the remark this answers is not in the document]
holisticode: Hmm, maybe we could. However, due to the independence of the flags for each, this can quickly become very confusing, requiring multiple checks for mutual exclusion (i.e. the flags required for one can't be set for the other and vice-versa).

cmd/subnetcmd/import_from_api.go
holisticode: That's a good remark, however my concern is that many such checks will make the refactor harder once we want to support more chains per subnet?

cmd/subnetcmd/import_from_api.go
holisticode: I can create a ticket for sure, but I think it was an initial design decision here to do `subnetName == vmName`.  Not sure how easy we can move away from that?

cmd/subnetcmd/import_from_api.go
holisticode: Thanks for the pointer

cmd/subnetcmd/import_from_api.go
holisticode: https://github.com/ava-labs/avalanche-cli/issues/473

cmd/subnetcmd/import_from_api.go
holisticode: My opinion is that asking the user could result in more uncertainty. We also have the general pattern to make things for the user as easy as possible.

pkg/application/downloader.go
holisticode: That's exactly the reason to extract this code: it can be reused for any kind of request.
However, I realize that the error texts suggested a specific use case (most probably because I just moved existing code).
For true generality, I am going to rename the error texts.

I also will make the token a param to the function for true reusability.

pkg/prompts/prompts.go
holisticode: I thought so too, but that file doesn't exist! I guess some mishap during some merge....

cmd/subnetcmd/import_from_api.go
holisticode: https://github.com/ava-labs/avalanche-cli/issues/475

cmd/subnetcmd/import_from_api.go
holisticode: You're right! That's a leftover from a previous iteration. Thanks.

cmd/subnetcmd/import_from_api.go
holisticode: Not so sure about this? The idea of this code block is to get to the ChainID - which is subnet-evm specific as far as I understand?

cmd/subnetcmd/import_from_api.go
holisticode: Yep, resolved through DM chat

cmd/subnetcmd/import_from_api.go
holisticode: Also I think that the creation flow is somewhat different in that the user is probably indeed creating the subnet for the first time, and might be in a different mind (and time) context than when importing a subnet from a running network. The latter argument also gets strengthened by the fact that this probably stems from having created a subnet via subnet-cli, which indeed is a different context altogether than our cli.

cmd/subnetcmd/import_from_api.go
holisticode: Hmmm nicely spotted! This is a little messy though, as it's also a bit of a leftover from previous iterations.

The idea initially: 
* If the user chooses to query a running node, get info from that node
* In that case the version could be gathered from there
* If the node was not queried, then have the user pick the version manually

However this check now completely overrides whatever we prompted from the user, which of course is useless. 

So we could actually:
A) also only prompt for the version if a node wasn't queried.
I need to check though if there are edge cases where this breaks down.
B) Just get rid of this check and allow prompts only

I am currently in favor of A) but very slightly. I am ok doing B)

cmd/subnetcmd/import_from_api.go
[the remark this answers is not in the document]
holisticode: Your suggestion of making two separate sub-commands seems best.

cmd/subnetcmd/import_from_api.go
holisticode: We'll leave it to the pre-set TEST token for now. We can address this later if the need arises to make this configurable

pkg/prompts/prompts.go
holisticode: We'll extract this again in a follow up ticket: https://github.com/ava-labs/avalanche-cli/issues/477

cmd/subnetcmd/import_from_api.go
holisticode: A) is the decision

cmd/subnetcmd/import_from_api.go
holisticode: Keep the prompting here but also replace it in the creation flow

cmd/subnetcmd/import_from_api.go
holisticode: For spacesvm just do the unmarshalling to verify it's valid

cmd/subnetcmd/import.go
```diff
@@ -3,282 +3,28 @@
...
-				repoAlias = installedRepo
-				break
-			}
-		}
-		if repoAlias == "" {
-			repoAlias = customRepo
-			repoURL, err = url.ParseRequestURI(repoOrURL)
+		Use:   "import",
+		Short: "Import subnets into avalanche-cli",
+		Long: `The subnet command suite provides a collection of tools for developing
```
dasconnor, line 16: This will require an update to the https://docs.avax.network/subnets/reference-cli-commands page

cmd/subnetcmd/import_file.go
```diff
@@ -0,0 +1,284 @@
...
+func newImportFileCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "file [subnetPath]",
+		Short:        "Import an existing subnet config",
+		RunE:         importSubnet,
+		SilenceUsage: true,
+		Args:         cobra.MaximumNArgs(1),
+		Long: `The subnet import command will import a subnet configuration from a file or a git repository.
+
+To import from a file, you can optionally provide the filepath as a command line argument.
```
dasconnor, line 36: I may have changed this slightly in #467

cmd/subnetcmd/import_from_api.go
```diff
@@ -204,72 +205,79 @@ func importRunningSubnet(cmd *cobra.Command, args []string) error {
...
-		return err
 	}
 
-	// hasn't been set in reply
-	if sc.RPCVersion == 0 {
-		sc.RPCVersion, err = vm.GetRPCProtocolVersion(app, vmType, sc.VMVersion)
-		if err != nil {
-			return fmt.Errorf("failed getting RPCVersion for VM type %s with version %s", vmType, sc.VMVersion)
-		}
+	sc.RPCVersion, err = vm.GetRPCProtocolVersion(app, vmType, sc.VMVersion)
```
dasconnor, line 263: Does this need to be inside the else? It looks like it's set in the `if`

dasconnor (commented):
Looks good, just fix the merge conflict and answer my one question and should be good to go

cmd/subnetcmd/import_from_api.go
holisticode: Great catch! I think so too.

cmd/subnetcmd/import.go
holisticode: https://github.com/ava-labs/avalanche-docs/pull/1096

cmd/subnetcmd/import_file.go
holisticode: Indeed, incorporated.

cmd/subnetcmd/import_file.go
```diff
@@ -33,10 +33,10 @@ func newImportFileCmd() *cobra.Command {
 		Args:         cobra.MaximumNArgs(1),
 		Long: `The subnet import command will import a subnet configuration from a file or a git repository.
 
-To import from a file, you can optionally provide the filepath as a command line argument.
-Alternatively, running the command without any arguments will trigger an interactive wizard.
-To import from a repo, go through the wizard. By default, an imported subnet will not
-overwrite an existing subnet with the same name. To allow overwrites, provide the --force
+To import from a file, you can optionally provide the path as a commandR-line argument.
```
dasconnor, line 36: typo, `commandR`

dasconnor (approved):
one typo, but lgtm

Issue #441, opened 2022-11-18
Set default snapshot to use permissionless validators with BLS keys in the primary subnet


felipemadero:
By using the snapshot:
1. local network nodes will always have the save BLS public key and proof of possession (which can be queried with info.getNodeID), among different network starts
2. local network nodes will be permissionless validators using those same BLS keys (which can be queried with platform.getCurrentValidators)

PR #464, opened 2022-11-24
point to bls enabled ANR and use snapshot with bls validators

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/441


dasconnor (approved):
The pipeline passed, so I think this is ok, but I do just want to double check that the new snapshot is ok. I'm a little concerned that it's 94% smaller haha

PR #465, opened 2022-11-28
use ~/bin for install

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/466


dasconnor (approved):
We'll need to update the docs repo with this.

PR #467, opened 2022-11-28
Updated doc strings to match doc repo

dasconnor:


cmd/subnetcmd/addValidator.go
```diff
@@ -42,7 +42,8 @@ the subnetName and the validator's unique NodeID. The command then prompts
 for the validation start time, duration and stake weight. These values can
 all be collected with flags instead of prompts.
 
-This command currently only works on subnets deployed to the Fuji testnet.`,
+This command currently only works on subnets deployed to either the Fuji
```
holisticode, line 45: I think somewhere we defined how we write fuji, testnet, mainnet in the docs. Like `fuji` vs `Fuji` vs Fuji etc.

I think we should try to be consistent.

holisticode (approved):

felipemadero (approved):

PR #468, opened 2022-11-28
Don't add dependabot PRs to project

dasconnor:
Dependabot pipelines are currently failing because dependabot-triggered actions can't access repo secrets. This disables the action for dependabot PRs and should allow their pipelines to pass.


felipemadero (approved):
LGTM

Issue #427, opened 2022-11-16
make install script to not fail on macos if brew is not present


felipemadero:
notify the CLI user (and/or update doc) that bash completion on macos needs a working brew

PR #470, opened 2022-11-28
Dont fail if brew not installed

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/427

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -400,7 +400,7 @@ then
         sed_in_place "/.*# avalanche completion/d" $BASHRC
         echo "source $(brew --prefix)/etc/bash_completion # avalanche completion" >> $BASHRC
     else 
-        echo "brew not found on macos: bash avalanche command completion not installed"
+        echo "warning: brew not found on macos. bash avalanche command completion not installed"
     fi
 fi
 
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -391,8 +391,8 @@ sed_in_place "/.*# avalanche completion/d" $BASH_COMPLETION_MAIN
 echo "source $BASH_COMPLETION_SCRIPT_PATH # avalanche completion" >> $BASH_COMPLETION_MAIN
 if [ $(uname) = Darwin ]
 then
-    BREW_INSTALLED=true
-    which brew > /dev/null 2> /dev/null || BREW_INSTALLED=false
+    BREW_INSTALLED=false
+    which brew >/dev/null 2>&1 && BREW_INSTALLED=true
     if [ $BREW_INSTALLED = true ]
     then
         BASHRC=~/.bashrc
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -391,10 +391,17 @@ sed_in_place "/.*# avalanche completion/d" $BASH_COMPLETION_MAIN
 echo "source $BASH_COMPLETION_SCRIPT_PATH # avalanche completion" >> $BASH_COMPLETION_MAIN
 if [ $(uname) = Darwin ]
 then
-    BASHRC=~/.bashrc
-    touch $BASHRC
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -391,10 +391,17 @@ sed_in_place "/.*# avalanche completion/d" $BASH_COMPLETION_MAIN
 echo "source $BASH_COMPLETION_SCRIPT_PATH # avalanche completion" >> $BASH_COMPLETION_MAIN
 if [ $(uname) = Darwin ]
 then
-    BASHRC=~/.bashrc
-    touch $BASHRC
-    sed_in_place "/.*# avalanche completion/d" $BASHRC
-    echo "source $(brew --prefix)/etc/bash_completion # avalanche completion" >> $BASHRC
+    BREW_INSTALLED=true
```
holisticode, line 394: I suggest to start with a false value, and then set only if installed. This is a general pattern when evaluating booleans I think: first the case which doesn't assume the value (`BREW_INSTALLED=true` assumes it is installed already, even before we evaluate.)

scripts/install.sh
```diff
@@ -391,10 +391,17 @@ sed_in_place "/.*# avalanche completion/d" $BASH_COMPLETION_MAIN
...
+    which brew > /dev/null 2> /dev/null || BREW_INSTALLED=false
```
holisticode, line 395: Why do we need this double indirection? Can we do this? `which brew &>/dev/null` ?

scripts/install.sh
```diff
@@ -391,10 +391,17 @@ sed_in_place "/.*# avalanche completion/d" $BASH_COMPLETION_MAIN
...
+    if [ $BREW_INSTALLED = true ]
+    then
+        BASHRC=~/.bashrc
```
holisticode, line 398: If we are generally limiting scripts to `bash` only, I guess this is fine. Obviously this wouldn't work for `zsh`. I don't know if we have any such limitation though.

scripts/install.sh
```diff
@@ -391,10 +391,17 @@ sed_in_place "/.*# avalanche completion/d" $BASH_COMPLETION_MAIN
...
+        touch $BASHRC
+        sed_in_place "/.*# avalanche completion/d" $BASHRC
+        echo "source $(brew --prefix)/etc/bash_completion # avalanche completion" >> $BASHRC
+    else 
+        echo "warning: brew not found on macos. bash avalanche command completion not installed"
+    fi
 fi
 
 ZSH_COMPLETION_MAIN=~/.zshrc
```
holisticode, line 407: In fact this suggests we are also supporting zsh.

scripts/install.sh
felipemadero: to avoid the script to abort on `which` failure, a `||` connector is used and for that reason the flag is set this way. either way the flag is correctly set.

scripts/install.sh
felipemadero: we are in general using `sh` not `bash` , `sh` in general does not support that redirection

scripts/install.sh
felipemadero: the brew dependency is related to bash configuration

scripts/install.sh
felipemadero: the brew dependency is related to bash configuration

scripts/install.sh
holisticode: Then a simple rename to `INSTALL_BREW` would be better. This expresses that the installation should be done, and only aborted if it's already installed.

scripts/install.sh
holisticode: ? 
I don't understand this comment.
Either we support both bash and zsh or only bash, and it seems we do both?

scripts/install.sh
felipemadero: you are right

holisticode (approved):

Issue #469, opened 2022-11-28
Download updated bootstrap snapshots


felipemadero:
Currently, an old user will not update to a new snapshot file, even if updating CLI version. Use checksum to check for new snapshot files.

PR #472, opened 2022-11-29
Download new snapshots

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/469


dasconnor (commented):
Looks pretty good, but I think we should try to prevent this one potential bug

dasconnor (approved):
LGTM when pipeline passes

PR #479, opened 2022-11-30
Replace assert with require

holisticode:
This is literally a search/replace iteration through the code base, which might have issues - especially because there are hardhat / node_modules JS directories after an e2e test which do have `assert` occurrences that need to stay.

So there might be issues.


dasconnor (approved):
This looks good, but we might want to wait till we merge most of our PR backlog to prevent a bunch of merge conflicts from popping up

PR #496, opened 2022-12-07
Added Subnet-EVM name migration

### LINKED ISSUE #492: Fix subnet describe bug
`subnet describe` is printing the genesis by default instead of the summary page.

Add a test to catch this in the future.

dasconnor:
Our subnet describe bug was caused by us migrating the constant `SubnetEVM` to `Subnet-EVM` for documentation purposes. This caused some string matching errors. The fix was to introduce a new migration to move all `SubnetEVM` references to use the new `Subnet-EVM` name.

Closes #492

### DIFF
--- a/cmd/subnetcmd/delete.go
+++ b/cmd/subnetcmd/delete.go
@@ -4,9 +4,9 @@ package subnetcmd
 
 import (
 	"os"
+	"path/filepath"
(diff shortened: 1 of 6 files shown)


holisticode (approved):
lgtm

PR #518, opened 2022-12-16
Update linter

danlaine:
Updates the linters and linter settings to match avalanchego, and updates the code to pass the linter.

There are some places where it wasn't clear to me whether to handle a previously unhandled error, or to ignore it. For example, if deletion of a file fails, should the error be propagated or should the program ignore it? I will leave it to you all to decide.

Also, if you feel that this repo shouldn't have certain linters/settings used in avalanchego, then by all means feel free to remove them.

### DIFF
--- a/cmd/subnetcmd/delete.go
+++ b/cmd/subnetcmd/delete.go
@@ -21,7 +21,7 @@ func newDeleteCmd() *cobra.Command {
 	}
 }
 
-func deleteSubnet(cmd *cobra.Command, args []string) error {
+func deleteSubnet(_ *cobra.Command, args []string) error {
 	// TODO sanitize this input
 	subnetName := args[0]
 	subnetDir := filepath.Join(app.GetSubnetDir(), subnetName)
@@ -34,10 +34,12 @@ func deleteSubnet(cmd *cobra.Command, args []string) error {
 	}
 
 	if sidecar.VM == models.CustomVM {
-		if _, err := os.Stat(customVMPath); err == nil {
-			// exists
-			os.Remove(customVMPath)
-		} else {
+		if _, err := os.Stat(customVMPath); err != nil {
+			return err
+		}
+
+		// exists
+		if err := os.Remove(customVMPath); err != nil {
 			return err
 		}
 	}
@@ -48,12 +50,13 @@ func deleteSubnet(cmd *cobra.Command, args []string) error {
 	// but only if no other subnet is using it.
 	// More info: https://github.com/ava-labs/avalanche-cli/issues/246
 
-	if _, err := os.Stat(subnetDir); err == nil {
-		// exists
-		os.RemoveAll(subnetDir)
-	} else {
+	if _, err := os.Stat(subnetDir); err != nil {
 		return err
 	}
 
+	// exists
+	if err := os.RemoveAll(subnetDir); err != nil {
+		return err
+	}
 	return nil
 }
--- a/cmd/subnetcmd/delete.go
+++ b/cmd/subnetcmd/delete.go
@@ -3,6 +3,8 @@
 package subnetcmd
 
 import (
+	"errors"
+	"io/fs"
 	"os"
 	"path/filepath"
 
@@ -35,7 +37,10 @@ func deleteSubnet(_ *cobra.Command, args []string) error {
 
 	if sidecar.VM == models.CustomVM {
 		if _, err := os.Stat(customVMPath); err != nil {
-			return err
+			if !errors.Is(err, fs.ErrNotExist) {
+				return err
+			}
+			app.Log.Warn("tried to remove custom VM path but it actually does not exist. Ignoring")
 		}
 
 		// exists
@@ -51,6 +56,10 @@ func deleteSubnet(_ *cobra.Command, args []string) error {
 	// More info: https://github.com/ava-labs/avalanche-cli/issues/246
 
 	if _, err := os.Stat(subnetDir); err != nil {
+		if !errors.Is(err, fs.ErrNotExist) {
+			return err
+		}
+		app.Log.Warn("tried to remove the Subnet dir path but it actually does not exist. Ignoring")
 		return err
 	}
 
--- a/cmd/subnetcmd/delete.go
+++ b/cmd/subnetcmd/delete.go
@@ -41,6 +41,7 @@ func deleteSubnet(_ *cobra.Command, args []string) error {
 				return err
 			}
 			app.Log.Warn("tried to remove custom VM path but it actually does not exist. Ignoring")
+			return nil
 		}
 
 		// exists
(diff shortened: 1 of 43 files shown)


cmd/subnetcmd/delete.go
```diff
@@ -48,12 +50,13 @@ func deleteSubnet(cmd *cobra.Command, args []string) error {
 	// but only if no other subnet is using it.
 	// More info: https://github.com/ava-labs/avalanche-cli/issues/246
 
-	if _, err := os.Stat(subnetDir); err == nil {
-		// exists
-		os.RemoveAll(subnetDir)
-	} else {
+	if _, err := os.Stat(subnetDir); err != nil {
```
holisticode, line 53: We will want to review all of these `Stat`s for the above issue.

cmd/subnetcmd/publish.go
```diff
@@ -294,39 +294,36 @@ func getNewAlias() (string, error) {
...
+		}
+		repoURL, err = app.Prompt.CaptureList("Which is the remote URL for this repo?", remotes)
 		if err != nil {
+			// should never happen
 			return err
 		}
+		return nil
 	}
-	return nil
+	app.Log.Debug(
```
holisticode, line 323: Please add a quick comment how we got here (`err != nil` which happened pretty early)

cmd/subnetcmd/publish.go
```diff
@@ -294,39 +294,36 @@ func getNewAlias() (string, error) {
...
+		"opening repo failed - alias might have not been created yet, so ignore", zap.String("alias", alias), zap.Error(err))
+	repoURL, err = app.Prompt.CaptureString("Provide the repository URL")
```
holisticode, line 325: This is not from your PR, but could you pls add a `TODO` here? It's a bit dangerous we are assigning to a global down here.

pkg/config/config.go
```diff
@@ -15,7 +15,9 @@ func New() *Config {
 	return &Config{}
 }
 
-func (c *Config) LoadNodeConfig() (string, error) {
+// TODO should this be a function instead of a method?
```
holisticode, line 18: Probably

pkg/prompts/prompts.go
```diff
@@ -280,11 +280,11 @@ func (*realPrompter) CaptureID(promptStr string) (ids.ID, error) {
 		Validate: validateID,
 	}
 
-	IDStr, err := prompt.Run()
+	idStr, err := prompt.Run()
```
holisticode, line 283: Huh, why does the linter not complain here? We always use `ID` for such labels?

pkg/ux/output.go
```diff
@@ -32,7 +32,7 @@ func NewUserLog(log logging.Logger, userwriter io.Writer) {
 // PrintToUser prints msg directly on the screen, but also to log file
 func (ul *UserLog) PrintToUser(msg string, args ...interface{}) {
 	formattedMsg := fmt.Sprintf(msg, args...)
-	fmt.Fprintln(ul.writer, formattedMsg)
+	_, _ = fmt.Fprintln(ul.writer, formattedMsg)
```
holisticode, line 35: Huh, this looks a bit ugly, do we really need to do this?

cmd/subnetcmd/delete.go
```diff
@@ -34,10 +34,12 @@ func deleteSubnet(cmd *cobra.Command, args []string) error {
 	}
 
 	if sidecar.VM == models.CustomVM {
-		if _, err := os.Stat(customVMPath); err == nil {
-			// exists
-			os.Remove(customVMPath)
-		} else {
+		if _, err := os.Stat(customVMPath); err != nil {
+			return err
```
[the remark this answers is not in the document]
danlaine: Like I said in the PR description
```
There are some places where it wasn't clear to me whether to handle a previously unhandled error, or to ignore it.
```
Totally up to you guys to decide what the right thing to do is in cases like this one

pkg/prompts/prompts.go
danlaine: It's ok with "id" and "ID", but it doesn't like "Id" and it doesn't like local scoped things starting with capital letters:

```
pkg/prompts/prompts.go:283:2: unexported-naming: the symbol IDStr is local, its name should start with a lowercase letter (revive)
```

pkg/config/config.go
danlaine: I'll leave that to you guys to do -- I don't think the `Config` struct is actually used anywhere (it's empty) and can be removed altogether. Maybe I'm missing something, though

pkg/ux/output.go
danlaine: changed linting rules so this isn't needed

cmd/subnetcmd/publish.go
danlaine: added comment on this function

cmd/subnetcmd/publish.go
danlaine: refactored this function -- see if it's alright with you now

cmd/subnetcmd/delete.go
danlaine: Same as above -- totally up to you guys to decide what the right thing to do is in cases like this one.

Would you mind handling such cases? You'd know better than me whether we want to ignore or propagate errors in various places.

cmd/subnetcmd/delete.go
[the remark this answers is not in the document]
holisticode: OK, but the point is that this is now changing the semantics inside your PR. 

So my suggestion here would be to ignore the error if the file does not exist and not return an non-nil `err`, requiring a change in this PR. @dasconnor @felipemadero

pkg/config/config.go
holisticode: Yeah was just a comment from my end agreeing with you. No action on your end required.

cmd/subnetcmd/delete.go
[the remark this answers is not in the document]
danlaine: Sure, do you want me to remove all changes in this PR that deal with the success of file deletions?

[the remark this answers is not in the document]
felipemadero: I think we can approve this if keeping the same sematics. That is not returning the new errors, instead just make a log.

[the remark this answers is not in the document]
dasconnor: Yeah, I think we want to keep the file deletion logic, that was intentional, mostly we don't care if we can't delete something because it's not actually there.

dasconnor (commented):
I'm good to merge this as soon as we revert the file deletion changes

dasconnor (approved):

felipemadero (approved):
LGTM

Issue #520, opened 2022-12-20
subnet create panic in check for admins balance


felipemadero:
with default values in subnet create, a panic is obtained

PR #521, opened 2022-12-20
fix subnet create panic in check for admins balance

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/520

### DIFF
--- a/pkg/vm/createEvm.go
+++ b/pkg/vm/createEvm.go
@@ -61,7 +61,11 @@ func CreateEvmSubnetConfig(app *application.Avalanche, subnetName string, genesi
 	return genesisBytes, sc, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/vm/createEvm.go
```diff
@@ -108,8 +112,10 @@ func createEvmGenesis(app *application.Avalanche, subnetName string, subnetEVMVe
 		subnetEvmState.NextState(direction)
 	}
 
-	if err := ensureAdminsHaveBalance(conf.TxAllowListConfig.AllowListAdmins, allocation); err != nil {
-		return nil, nil, err
+	if conf.TxAllowListConfig != nil {
```
holisticode, line 115: Right, but maybe we now even want to check that `conf` is not nil for safety/

holisticode (approved):

PR #536, opened 2022-12-26
Remove incorrect genesis validity check

StephenButtolph:
It seems like the avalanche-cli previously assumed that the `genesis` file needed to be `json` encoded... This is just not the case, as VMs define the genesis format. Generally, it would be better _not_ to use `json`, as that leads to bloat on the P-chain.

### DIFF
--- a/cmd/subnetcmd/deploy.go
+++ b/cmd/subnetcmd/deploy.go
@@ -196,9 +196,6 @@ func deploySubnet(_ *cobra.Command, args []string) error {
 	case models.SpacesVM:
 		var genesis spacesvmchain.Genesis
 		err = json.Unmarshal(chainGenesis, &genesis)
-	default:
-		var genesis map[string]interface{}
-		err = json.Unmarshal(chainGenesis, &genesis)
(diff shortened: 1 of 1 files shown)


cmd/subnetcmd/deploy.go
```diff
@@ -196,9 +196,6 @@ func deploySubnet(_ *cobra.Command, args []string) error {
 	case models.SpacesVM:
 		var genesis spacesvmchain.Genesis
 		err = json.Unmarshal(chainGenesis, &genesis)
-	default:
```
holisticode, line 199: Could we think of some other sensible `default`? Or should we have no default case whatsoever.

holisticode (approved):

felipemadero (approved):

cmd/subnetcmd/deploy.go
StephenButtolph: Because a custom VM can arbitrarily define the genesis format, there is no generalized verification that can be performed here.

PR #539, opened 2023-01-05
update to new ledger app

felipemadero:
- checked this works on local e2e with real ledger for app v0.6.7 with nano S and nano S+
- as the new ledger app no longer asks for confirmation of public key provisioning, this particular CLI change impacts in tutorials/howtos related to ledger


dasconnor (commented):
What happens if we connect the old ledger app? Would it error anywhere? Could we provide a warning?

felipemadero:
> What happens if we connect the old ledger app? Would it error anywhere? Could we provide a warning or recommendation to upgrade the app?

It would error on ledger operation. I have no old app currently so can not check for the error so as to provide a smart recommentation. But can add a recommendation for ledger update if there is any ledger error.

dasconnor (approved):
This looks good, not sure if we want to merge right now or not

PR #560, opened 2023-01-19
use anr that fix flags based on avago version

felipemadero:


pkg/subnet/local.go
```diff
@@ -115,7 +113,7 @@ func (d *LocalDeployer) BackendStartedHere() bool {
 //   - waits completion of operation
 //   - show status
 func (d *LocalDeployer) doDeploy(chain string, chainGenesis []byte, genesisPath string) (ids.ID, ids.ID, error) {
-	avagoVersion, avalancheGoBinPath, pluginDir, err := d.SetupLocalEnv()
+	_, avalancheGoBinPath, pluginDir, err := d.SetupLocalEnv()
```
holisticode, line 116: These are actually the only places where `d.SetupLocalEnv` seems to be returning the version.

If it's not needed anymore, should we eliminate that return param?

holisticode: Why do we actually not need it anymore?

felipemadero: It was used so as the CLI changed the ANR call based on avago version. It is not needed anymore if ANR handles that.

felipemadero: Will remove it from all functions

dasconnor (approved):
LGTM

Issue #372, opened 2022-10-28
Allow upgrade bytes to support separate fuji, mainnet, and local deployments


dasconnor:
Users should be able to test their upgrades on local network and fuji before pushing them to mainnet.

PR #566, opened 2023-01-23
UpgradeBytes 1st Phase: Install upgrade bytes on local network

holisticode:
This PR partly addresses https://github.com/ava-labs/avalanche-cli/issues/372

It allows to install upgrade bytes into a locally running network.
For example:
`avalanche subnet upgrade install <subnetName>`

The file should have previously been created with 
`avalanche subnet upgrade generate <subnetName` or been imported with the `import` command.


cmd/subnetcmd/upgradecmd/install.go
```diff
@@ -0,0 +1,116 @@
...
+
+const (
+	timestampFormat  = "20060102150405"
+	tmpSnapshotInfix = "-tmp-"
+)
+
+// avalanche subnet upgrade generate
+func newUpgradeInstallCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "install [subnetName]",
```
dasconnor, line 29: This will need a docs PR. We'll actually need one for all the commands I think

cmd/subnetcmd/upgradecmd/install.go
```diff
@@ -0,0 +1,116 @@
...
+		Short: "Installs upgrade bytes onto subnet nodes",
+		Long:  `Installs upgrade bytes onto VMs.`, // TODO fix this wording
```
dasconnor, line 31: Your long is shorter than your short lol

dasconnor: Apply generated upgrade bytes to running subnet nodes to trigger a network upgrade.

cmd/subnetcmd/upgradecmd/install.go
dasconnor, line 29: I think I like the term `apply` better than `install`

dasconnor: `commit` might be better too

cmd/subnetcmd/upgradecmd/install.go
```diff
@@ -0,0 +1,116 @@
...
+	if !app.SubnetConfigExists(subnetName) {
+		return errors.New("subnet does not exist")
+	}
+
+	sc, err := app.LoadSidecar(subnetName)
+	if err != nil {
+		return fmt.Errorf("unable to load sidecar: %w", err)
+	}
+
+	networkToUpgrade, err := selectNetworkToUpgrade(sc)
```
dasconnor, line 50: you'll need a flag for this for tests

cmd/subnetcmd/upgradecmd/install.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return err
+	}
+
+	switch networkToUpgrade {
+	// in this case, we just are going to generate new update bytes
+	case futureDeployment:
+		return upgradeGenerateCmd(cmd, args)
```
dasconnor, line 58: Instead of generating new upgrade bytes, should we suggest the user changes their precompile configuration and redeploys from scratch? Say it's much safer than using upgrade bytes

dasconnor: Maybe this should just edit the precompile config in the genesis?

cmd/subnetcmd/upgradecmd/install.go
```diff
@@ -0,0 +1,116 @@
...
+	// update a locally running network
+	case localDeployment:
+		return localInstallNetworkUpgrades(subnetName, sc)
+	}
+
+	return nil
+}
+
+func localInstallNetworkUpgrades(subnetName string, sc models.Sidecar) error {
```
dasconnor, line 67: Definitely should write an e2e test for this.

cmd/subnetcmd/upgradecmd/install.go
```diff
@@ -0,0 +1,116 @@
...
+	// For a already deployed subnet, the supported scheme is to
+	// save a snapshot, and to load the snapshot with the upgrade
```
dasconnor, line 69: Can we not just restart the nodes? Why do we need to mess with snapshots?

cmd/subnetcmd/upgradecmd/install.go
```diff
@@ -0,0 +1,116 @@
...
+	_, err = cli.SaveSnapshot(ctx, snapName)
+	if err != nil {
+		return err
+	}
+	app.Log.Debug(
+		"network stopped and named temporary snapshot created. Now starting the network with given snapshot")
+
+	// restart the network setting the upgrade bytes file
+	netUpgradeBytes, err := upgrades.ReadUpgradeFile(subnetName, app.GetSubnetDir())
+	if err != nil {
```
dasconnor, line 95: You definitely want to check this before you shutdown the network. You also don't restore it if there's an error which could cause problems.

cmd/subnetcmd/upgradecmd/install.go
```diff
@@ -0,0 +1,116 @@
...
+		return err
+	}
+
+	// TODO input validation
+	netUpgradeConfs := map[string]string{
+		blockchainID.String(): string(netUpgradeBytes),
+	}
+
+	opts := ANRclient.WithUpgradeConfigs(netUpgradeConfs)
+	_, err = cli.LoadSnapshot(ctx, snapName, opts)
```
dasconnor, line 109: ahh, so because we can pass the configs to the network directly, the snapshot loading is kinda elegant

cmd/subnetcmd/upgradecmd/install.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		return err
+	}
+
+	ux.Logger.PrintToUser("Network restarted and upgrade bytes have been applied to running nodes")
```
dasconnor, line 114: Would be cool if we could write, `The next upgrade will go into effect at 12:00 am GMT 01/24/2023` or whatever

pkg/subnet/upgrades/upgrade_file.go
```diff
@@ -13,33 +13,26 @@ import (
...
-func WriteUpgradeFile(jsonBytes []byte, subnetName, upgradeFilesDir string) error {
+func WriteUpgradeFile(jsonBytes []byte, subnetName, subnetsRoot string) error {
 	var (
 		exists bool
 		err    error
 	)
 
-	subnetPath := filepath.Join(upgradeFilesDir, subnetName)
-	updateBytesFileName := filepath.Join(subnetPath, constants.UpdateBytesFileName)
+	subnetDir := filepath.Join(subnetsRoot, subnetName)
```
dasconnor, line 22: This should be done in the application package

cmd/subnetcmd/upgradecmd/install.go
holisticode: Do you mean just adding the flags to the command right? Because the flag variables are already defined in your `vm.go`

cmd/subnetcmd/upgradecmd/install.go
holisticode: Because we need to pass a new flag pointing to the update bytes file to ANR. ANR before this doesn't know anything about the existence of this file.

Currently the only supported way is to restart from a snapshot and provide the new flags via `LoadSnapshot`.

Alternatively, we could add a new endpoint to ANR which would do this behind the scenes somehow, but that would require a new ANR release. cc @felipemadero

cmd/subnetcmd/upgradecmd/install.go
holisticode: Yep, of course. I thought we'd do that when we have all our phases finished (including public networks).

cmd/subnetcmd/upgradecmd/install.go
holisticode: Indeed, hence the TODO. Thanks for the suggestion.

cmd/subnetcmd/upgradecmd/install.go
holisticode: OK. I like `apply` best so I'll use that.

cmd/subnetcmd/upgradecmd/install.go
holisticode: I say if we are not going to allow to generate new bytes, then we should not allow this option in the first place. The use case becomes confused (edit in the genesis? redeploy from scratch - but maybe they actually never deployed in the first place, so what does it mean? etc.)

cmd/subnetcmd/upgradecmd/install.go
holisticode: Of course. I thought I'd add one to allow multiple iterations. But it's also integral part of this functionality, so I can add an e2e test in this iteration.

cmd/subnetcmd/upgradecmd/install.go
holisticode: You're very right about checking first. So changed to do that. 

With this new logic then, there's only a `SaveSnapshot` immediately followed by `LoadSnapshot`. If the former failed, I don't think we should try restoring from here. We don't do that anywhere else anyways.

pkg/subnet/upgrades/upgrade_file.go
holisticode: I wanted explicitly to avoid dragging the `app` reference in here as it currently doesn't need it.

cmd/subnetcmd/upgradecmd/install.go
holisticode: To clarify my position: This is the `apply` command - it will just apply whatever json it finds on the network.

It is not the task of the `apply` command to identify the state of the `upgrade.json` file. 

`futureDeployment` to me clearly signals something which doesn't exist yet and is going to be done in the future. If we want to even allow this option _for the `apply` command_ , it's only for non-existing updates scenario IMO. Alternatively we disable `futureDeployment` for `apply` and that's it.

cmd/subnetcmd/upgradecmd/apply.go
```diff
@@ -0,0 +1,144 @@
...
+	// restart the network setting the upgrade bytes file
+	opts := ANRclient.WithUpgradeConfigs(netUpgradeConfs)
+	_, err = cli.LoadSnapshot(ctx, snapName, opts)
+	if err != nil {
+		return err
+	}
+
+	// TODO as noted elsewhere, we need to extract the health polling from the deployer
+	// we shouldn't have to create the deployer here just to poll for healthy
+	sd := subnet.NewLocalDeployer(app, "", "")
```
felipemadero, line 130: hopefully we will have ANR WaitForHealthy soon...

cmd/subnetcmd/upgradecmd/apply.go
```diff
@@ -0,0 +1,144 @@
...
+	blockchainID := sc.Networks[models.Local.String()].BlockchainID
+	if blockchainID == ids.Empty {
+		return errors.New(
+			"failed to find deployment information about this subnet in state - aborting")
+	}
+
+	// save a temporary snapshot
+	snapName := subnetName + tmpSnapshotInfix + time.Now().Format(timestampFormat)
+	app.Log.Debug("saving temporary snapshot for upgrade bytes", zap.String("snapshot-name", snapName))
+	_, err = cli.SaveSnapshot(ctx, snapName)
```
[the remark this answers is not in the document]
felipemadero: what happen is the subnet has not been deployed yet?

cmd/subnetcmd/upgradecmd/install.go
holisticode: Implemented suggestion for this

pkg/constants/constants.go
```diff
@@ -110,9 +109,9 @@ const (
 
 	AvaLabsMaintainers = "ava-labs"
 
-	UpdateBytesFileName = "upgrade.json"
-	NotAvailableLabel   = "Not available"
-	BackendCmd          = "avalanche-cli-backend"
+	UpgradeBytesFileName = "upgrade.json"
```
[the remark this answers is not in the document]
holisticode: Subnet-evm distinguishes between NetworkUpgrades and PrecompileUpgrades, I think we should stick to the the latter in this context:

https://github.com/ava-labs/subnet-evm/blob/master/params/config.go#L168

tests/e2e/assets/test_subnet_evm_genesis.json
```diff
@@ -21,18 +21,6 @@
       "minBlockGasCost": 0,
       "maxBlockGasCost": 1000000,
       "blockGasCostStep": 200000
-    },
```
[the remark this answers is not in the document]
holisticode: We could. You think it's more clear if we set it back so it's visible that we are setting those variables to zero?

If you prefer that, then I'd appreciate you create a ticket for that so that we don't forget.

pkg/subnet/upgrades/upgrade_file.go
```diff
@@ -13,33 +13,26 @@ import (
...
+	updateBytesFileName := filepath.Join(subnetDir, constants.UpgradeBytesFileName)
 
-	ux.Logger.PrintToUser(fmt.Sprintf("Writing %q file to %q...", constants.UpdateBytesFileName, subnetPath))
+	ux.Logger.PrintToUser(fmt.Sprintf("Writing %q file to %q...", constants.UpgradeBytesFileName, subnetDir))
 
-	exists, err = storage.FolderExists(upgradeFilesDir)
+	// NOTE: This allows creating the update bytes file before a subnet has actually been created.
```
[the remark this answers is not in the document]
holisticode: Not sure we can call it a use case. It's just that this file is being written to disk, and it's expecting a specific path. Just making sure that path actually exists before we write the file. As noted in the comment, in practice it should never happen, but seems reasonable to make sure the code doesn't fail?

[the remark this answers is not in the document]
holisticode: Not sure we can call it a use case. It's just that this file is being written to disk, and it's expecting a specific path. Just making sure that path actually exists before we write the file. As noted in the comment, in practice it should never happen, but seems reasonable to make sure the code doesn't fail?

tests/e2e/commands/upgrade.go
```diff
@@ -10,6 +10,26 @@ import (
...
+		CLIBinary,
+		SubnetCmd,
+		UpgradeCmd,
+		"import",
+		subnetName,
+		"--upgrade-filepath",
+		filepath,
+	)
+
+	output, err := cmd.Output()
```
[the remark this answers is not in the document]
holisticode: Copy pasted from the other commands in this file, but agreed. Will change all calls in this file to use that.

tests/e2e/testcases/upgrade/suite.go
```diff
@@ -58,6 +63,32 @@ var _ = ginkgo.Describe("[Upgrade]", ginkgo.Ordered, func() {
 		_ = utils.DeleteKey(keyName)
 	})
 
+	ginkgo.It("can create and apply to locally running subnet", func() {
```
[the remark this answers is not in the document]
holisticode: Line 39 above already does that, and it's the main `Describe` of all tests in this suite, so I don't think it has to?

cmd/subnetcmd/upgradecmd/validate.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func getAllUpgrades(file []byte) ([]params.PrecompileUpgrade, error) {
+	var precompiles params.UpgradeConfig
+
+	if err := json.Unmarshal(file, &precompiles); err != nil {
+		return nil, fmt.Errorf("failed parsing JSON - %s: %w", err.Error(), errInvalidPrecompiles)
+	}
+
+	if precompiles.PrecompileUpgrades == nil || len(precompiles.PrecompileUpgrades) == 0 {
```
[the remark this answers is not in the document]
holisticode: It is of course not the same and semantically can mean something else, but for what we use it here the `len == 0` case is enough. So will change to this.

cmd/subnetcmd/upgradecmd/validate.go
```diff
@@ -0,0 +1,154 @@
...
+		return err
+	}
+
+	allTimestamps, err := getAllTimestamps(upgrades)
+	if err != nil {
+		return err
+	}
+
+	for _, ts := range allTimestamps {
+		if time.Unix(ts, 0).Before(time.Now()) {
```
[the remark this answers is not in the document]
holisticode: Yes that's possible and this is a very good point. However, we currently are not supporting state of upgrades very well anyways. 

The above check should be applied only to **new** upgrades - and for new ones it is **essential**.

We need to able to track state of upgrades because upgrades which have already been applied will have to stay in the `upgrade.json`. I was planning to do this in the next iteration (but before release), but I didn't have a specific ticket yet. Will create a new ticket for this.

cmd/subnetcmd/upgradecmd/apply.go
```diff
@@ -0,0 +1,144 @@
...
+func newUpgradeApplyCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "apply [subnetName]",
+		Short: "Apply upgrade bytes onto subnet nodes",
+		Long:  `Apply generated upgrade bytes to running subnet nodes to trigger a network upgrade`,
+		RunE:  applyCmd,
+		Args:  cobra.ExactArgs(1),
+	}
+
+	cmd.Flags().BoolVar(&useConfig, "config", false, "create upgrade config for future subnet deployments (same as generate)")
```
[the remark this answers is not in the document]
holisticode: There is an ongoing discussion here for this https://github.com/ava-labs/avalanche-cli/pull/566#discussion_r1084414697.

The point is that we are reusing `selectNetworkToUpgrade` here and that allows that option. I am fine not allowing it altogether, as I noted in the discussion above.

cmd/subnetcmd/upgradecmd/apply.go
```diff
@@ -0,0 +1,144 @@
...
+	cmd.Flags().BoolVar(&useLocal, "local", false, "apply upgrade existing `local` deployment")
+	cmd.Flags().BoolVar(&useFuji, "fuji", false, "apply upgrade existing `fuji` deployment (alias for `testnet`)")
+	cmd.Flags().BoolVar(&useFuji, "testnet", false, "apply upgrade existing `testbet` deployment (alias for `fuji`)")
```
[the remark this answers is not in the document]
holisticode: In fact copy pasted from vm.go, so changing it there too.

cmd/subnetcmd/upgradecmd/apply.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	cli, err := binutils.NewGRPCClient()
+	if err != nil {
+		return err
+	}
+	ctx := binutils.GetAsyncContext()
+
+	// get the blockchainID from the sidecar
+	blockchainID := sc.Networks[models.Local.String()].BlockchainID
```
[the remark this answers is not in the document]
holisticode: Yes we were discussing this with @dasconnor too. I think we should be putting more thought on this and make a decision.

cmd/subnetcmd/upgradecmd/apply.go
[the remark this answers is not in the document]
holisticode: All very good questions!

For the latter question we need state - therefore again, this goes into the next iteration.

For the first point - if we just start it, but the subnet has not been deployed yet, the upgrade would fail (and it should fail). Hence we need again state...

So this iteration only works with already deployed and running subnet. I suggest to address these things in the next iteration. However, I also think we should not release before we have addressed this issues.

pkg/subnet/upgrades/upgrade_file.go
holisticode: OK moved it to app finally

pkg/subnet/upgrades/upgrade_file.go
```diff
@@ -51,9 +44,8 @@ func WriteUpgradeFile(jsonBytes []byte, subnetName, upgradeFilesDir string) erro
 	return nil
 }
 
-func ReadUpgradeFile(subnetName, upgradeFilesDir string) ([]byte, error) {
-	subnetPath := filepath.Join(upgradeFilesDir, subnetName)
-	localUpgradeBytesFileName := filepath.Join(subnetPath, constants.UpdateBytesFileName)
+func ReadUpgradeFile(subnetName, subnetsRoot string) ([]byte, error) {
+	localUpgradeBytesFileName := filepath.Join(subnetsRoot, subnetName, constants.UpgradeBytesFileName)
```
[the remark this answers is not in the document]
holisticode: Ok moved to app

cmd/subnetcmd/upgradecmd/apply.go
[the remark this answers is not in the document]
dasconnor: Let's just remove this. I don't think we should make this an alias for generate.

cmd/subnetcmd/upgradecmd/apply.go
```diff
@@ -58,15 +67,32 @@ func installCmd(cmd *cobra.Command, args []string) error {
...
+func applyLocalNetworkUpgrade(subnetName string, sc models.Sidecar) error {
 	// For a already deployed subnet, the supported scheme is to
 	// save a snapshot, and to load the snapshot with the upgrade
+
+	// first let's check update bytes actually exist
+	netUpgradeBytes, err := upgrades.ReadUpgradeFile(subnetName, app)
+	if err != nil {
+		if err == os.ErrNotExist {
+			ux.Logger.PrintToUser("No file with upgrade specs for the given subnet has been found")
+			ux.Logger.PrintToUser("You may need to first create it with the `avalanche subnet update generate` command or import it")
```
dasconnor, line 85: `subnet upgrade generate`?

dasconnor (commented):
Looks close!

cmd/subnetcmd/upgradecmd/install.go
dasconnor: Yeah, just remove this option

holisticode: Removed the option but had to refactor `selectNetworkToUpgrade` a bit so it would take a default option list and send an empty one here.

cmd/subnetcmd/upgradecmd/apply.go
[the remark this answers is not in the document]
holisticode: Removed with small refactoring of `selectNetworkToUpgrade`

cmd/subnetcmd/upgradecmd/apply.go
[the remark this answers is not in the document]
holisticode: Decision: stick to the Sidecar for now

tests/e2e/assets/test_subnet_evm_genesis.json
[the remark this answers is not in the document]
felipemadero: not preference for this

pkg/subnet/upgrades/upgrade_file.go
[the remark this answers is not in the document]
felipemadero: Fair

felipemadero (approved):
LGTM. Of course would like @dasconnor to approve also

cmd/subnetcmd/upgradecmd/apply.go
```diff
@@ -0,0 +1,155 @@
...
+		blockchainID.String(): string(netUpgradeBytes),
+	}
+	// restart the network setting the upgrade bytes file
+	opts := ANRclient.WithUpgradeConfigs(netUpgradeConfs)
+	_, err = cli.LoadSnapshot(ctx, snapName, opts)
+	if err != nil {
+		return err
+	}
+
+	// TODO as noted elsewhere, we need to extract the health polling from the deployer
```
dasconnor, line 132: Can this be resolved with the new async wait PR?

dasconnor (approved):
One question, then LGTM

cmd/subnetcmd/upgradecmd/install.go
dasconnor: Added #580 for this

cmd/subnetcmd/upgradecmd/apply.go
holisticode: Yes, in fact it has to, because it has already been merged

PR #581, opened 2023-02-03
Upgrade Bytes phase 2

holisticode:


pkg/subnet/upgrades/upgrade_file.go
```diff
@@ -14,19 +13,34 @@ import (
 	"github.com/ava-labs/avalanchego/utils/storage"
 )
 
+func WriteLockUpgradeFile(jsonBytes []byte, subnetName string, app *application.Avalanche) error {
```
dasconnor, line 16: We manage all other file operations through the `application` package

dasconnor: Also, it's very unclear to me what the contents of the lock file actually are. I suggest creating a lock file type (or reusing an existing type) and accepting that as an argument, then do any type conversions in this method, then finally call out to the app package to write the file.

pkg/subnet/upgrades/upgrade_file.go
```diff
@@ -37,25 +51,31 @@ func WriteUpgradeFile(jsonBytes []byte, subnetName string, app *application.Aval
...
 }
 
 func ReadUpgradeFile(subnetName string, app *application.Avalanche) ([]byte, error) {
 	localUpgradeBytesFilePath := app.GetUpgradeBytesFilePath(subnetName)
 
-	exists, err := storage.FileExists(localUpgradeBytesFilePath)
+	return readFile(localUpgradeBytesFilePath)
+}
+
+func ReadLockUpgradeFile(subnetName string, app *application.Avalanche) ([]byte, error) {
```
dasconnor, line 63: Same comment as above. Use the app package to read the file, do type conversions and return a mature data type, not []byte

dasconnor (commented):
Just need to improve some of the file IO, then I think it'll be good

pkg/subnet/upgrades/upgrade_file.go
holisticode: OK moving to `application` package__

dasconnor (approved):
Looks a lot cleaner!

PR #590, opened 2023-02-08
Moved plugin dir to new dir

dasconnor:
Instead of installing plugins in to version specific dirs like `~/.avalanche-cli/bin/avalanchego/v1.9.5/plugins`, they are now installed to `~/.avalanche-cli/plugins`


felipemadero (approved):
LGTM

PR #591, opened 2023-02-08
Fix failed deploys

dasconnor:
Remove plugin binary if deployment fails.

### DIFF
--- a/pkg/subnet/local.go
+++ b/pkg/subnet/local.go
@@ -155,12 +155,20 @@ func (d *LocalDeployer) doDeploy(chain string, chainGenesis []byte, genesisPath 
 
 	if !networkBooted {
 		if err := d.startNetwork(ctx, cli, avalancheGoBinPath, runDir); err != nil {
+			secondErr := d.removeInstalledPlugin(chainVMID)
+			if secondErr != nil {
+				ux.Logger.PrintToUser("Failed to remove plugin binary: %s", secondErr)
+			}
 			return ids.Empty, ids.Empty, err
 		}
 	}
 
 	clusterInfo, err = WaitForHealthy(ctx, cli)
 	if err != nil {
+		secondErr := d.removeInstalledPlugin(chainVMID)
+		if secondErr != nil {
+			ux.Logger.PrintToUser("Failed to remove plugin binary: %s", secondErr)
+		}
 		return ids.Empty, ids.Empty, fmt.Errorf("failed to query network health: %w", err)
 	}
 	subnetIDs := clusterInfo.Subnets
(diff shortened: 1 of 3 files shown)


pkg/subnet/local.go
```diff
@@ -155,12 +155,20 @@ func (d *LocalDeployer) doDeploy(chain string, chainGenesis []byte, genesisPath
 
 	if !networkBooted {
 		if err := d.startNetwork(ctx, cli, avalancheGoBinPath, runDir); err != nil {
+			secondErr := d.removeInstalledPlugin(chainVMID)
```
felipemadero, line 158: we can avoid this one by moving installPlugin just before calling CreateBlockchains

pkg/subnet/local.go
```diff
@@ -155,12 +155,20 @@ func (d *LocalDeployer) doDeploy(chain string, chainGenesis []byte, genesisPath
...
+			if secondErr != nil {
+				ux.Logger.PrintToUser("Failed to remove plugin binary: %s", secondErr)
+			}
 			return ids.Empty, ids.Empty, err
 		}
 	}
 
 	clusterInfo, err = WaitForHealthy(ctx, cli)
 	if err != nil {
+		secondErr := d.removeInstalledPlugin(chainVMID)
```
felipemadero, line 168: we can avoid this one by moving installPlugin just before calling CreateBlockchains

pkg/subnet/local.go
```diff
@@ -217,6 +229,10 @@ func (d *LocalDeployer) doDeploy(chain string, chainGenesis []byte, genesisPath
...
```
felipemadero, line 232: (note this is related to LoadSidecard, not to this line). can we move LoadSidecard errors to before BlockchainCreation? Probably would like all the errors after CreateBlockchain, to cause plugin removal

pkg/subnet/local.go
```diff
@@ -207,6 +215,10 @@ func (d *LocalDeployer) doDeploy(chain string, chainGenesis []byte, genesisPath
 		blockchainSpecs,
 	)
 	if err != nil {
+		secondErr := d.removeInstalledPlugin(chainVMID)
+		if secondErr != nil {
```
felipemadero, line 219: would prefer other name, like pluginRemoveErr

pkg/subnet/local.go
dasconnor: moved it

pkg/subnet/local.go
dasconnor: moved it

pkg/subnet/local.go
dasconnor: moved it

pkg/subnet/local.go
dasconnor: renamed

PR #609, opened 2023-02-14
Upgrade local VMs

dasconnor:
An initial implementation of subnet upgrades for local deployments. There's definitely room to improve, but I encountered a lot of tricky corner cases, so I tried to simplify things as much as I could. As a compromise, I'm assumed the user shuts down the network first before performing the upgrade and then restarts it after.

Another area to improve, my e2e tests use hard coded versions. I'm not sure if this use case is covered by the current binary mapper, but I don't think it is. I'd like to improve that in follow-up work.

PR #610, opened 2023-02-14
fix status to correctly notify on stopped network

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/611


dasconnor:
I think I took care of this in #609. We should make the final fix by cutting an ANR release and bumping to the next real version. I think we can probably close this.

PR #633, opened 2023-02-22
First iteration of public upgrades

holisticode:
This is a first PR regarding upgrades.
There has been considerable thinking and discussing about the best way to implement public upgrades in `avalanche-cli`. There are many edge cases and issues to think about:
```
* No upgrade file for <public net> can be found - do we copy the existing file in the prev stage?   (for fuji: take the local, for main, take the fuji?)?
* If not, we exit, but then force the user to create a fuji file? Can be quite annoying!
* Do we validate that the fuji file is the same as local before applying? Or we just take whatever is there? For main, that it's the same as fuji and/or local or take whatever is there?
* What if the local deployment has applied different stages of upgrades, but they were for development only and fuji/main is going to be different (start from scratch)?
* What if someone isn't even doing local, just fuji and main...(or even just main...we may want to discourage that though...)
* User probably would never use the exact same file for local as for Fuji, because you’d probably want to change the timestamps...
```

For public networks we therefore limit ourselves to just "apply" the upgrades
This also means we are *ignoring* the lock file here!


tests/e2e/testcases/upgrade/suite.go
```diff
@@ -185,9 +245,8 @@ var _ = ginkgo.Describe("[Upgrade]", ginkgo.Ordered, func() {
 	})
 
 	ginkgo.It("upgrade SubnetEVM local deployment", func() {
-		// create and deploy
 		commands.CreateSubnetEvmConfigWithVersion(subnetName, utils.SubnetEvmGenesisPath, subnetEVMVersion1)
-		deployOutput := commands.DeploySubnetLocally(subnetName)
+		deployOutput := commands.DeploySubnetLocallyWithVersion(subnetName, avagoRPC1Version)
```
felipemadero, line 249: This change is not needed test will work anyway

tests/e2e/testcases/upgrade/suite.go
```diff
@@ -85,7 +86,64 @@ var _ = ginkgo.Describe("[Upgrade expect network failure]", ginkgo.Ordered, func
...
+// call the apply command which "just" installs the file at an expected path,
+// and then check the file is there and has the correct content.
+var _ = ginkgo.Describe("[Upgrade public network]", ginkgo.Ordered, func() {
+	ginkgo.AfterEach(func() {
+		commands.CleanNetworkHard()
+		err := utils.DeleteConfigs(subnetName)
+		gomega.Expect(err).Should(gomega.BeNil())
+	})
+
+	ginkgo.It("can create and apply to public node", func() {
```
felipemadero, line 100: do we need to be more clear regarding the upgrade that is beign performed? vm versus vs network upgrade...

tests/e2e/commands/upgrade.go
```diff
@@ -163,3 +163,24 @@ func ApplyUpgradeLocal(subnetName string) (string, error) {
 	}
 	return string(output), err
 }
+
+func ApplyUpgradePublic(subnetName, avagoChainConfDir string) (string, error) {
```
felipemadero, line 167: should be called ApplyNetwokUpgradePublic so it is clear which king of upgrade are we talking about

cmd/subnetcmd/upgradecmd/apply.go
```diff
@@ -97,34 +116,11 @@ func applyCmd(_ *cobra.Command, args []string) error {
...
-	}
-
-	// read the lock file right away
-	lockUpgradeBytes, err := app.ReadLockUpgradeFile(subnetName)
-	if err != nil {
-		// if the file doesn't exist, that's ok
-		if !os.IsNotExist(err) {
-			return err
-		}
+func applyLocalNetworkUpgrade(subnetName, networkKey string, sc *models.Sidecar) error {
```
felipemadero, line 119: do we need networkKey as param? always the same

cmd/subnetcmd/upgradecmd/apply.go
```diff
@@ -204,29 +200,151 @@ func applyLocalNetworkUpgrade(subnetName string, sc models.Sidecar) error {
...
+		return writeLockFile(precmpUpgrades, subnetName)
+	}
+
+	return errors.New("unexpected network size of zero nodes")
+}
+
+// applyPublicNetworkUpgrade applies an upgrade file to a locally running validator
+// for public networks (fuji, main)
+// the validation of the upgrade file has many things to consider:
+// * No upgrade file for <public net> can be found - do we copy the existing file in the prev stage?
```
felipemadero, line 212: are this questions going to be permanent here?

cmd/subnetcmd/upgradecmd/apply.go
```diff
@@ -204,29 +200,151 @@ func applyLocalNetworkUpgrade(subnetName string, sc models.Sidecar) error {
...
+// * What if someone isn't even doing local, just fuji and main...(or even just main...we may want to discourage that though...)
+// * User probably would never use the exact same file for local as for Fuji, because you’d probably want to change the timestamps
+//
+// For public networks we therefore limit ourselves to just "apply" the upgrades
+// This also means we are *ignoring* the lock file here!
+func applyPublicNetworkUpgrade(subnetName, networkKey string, sc *models.Sidecar) error {
+	if print {
+		blockchainIDstr := "<your-blockchain-id>"
+		if sc.Networks != nil &&
+			sc.Networks[networkKey] != (models.NetworkData{}) &&
```
felipemadero, line 228: is this needed if blockchain id is not empty?

cmd/subnetcmd/upgradecmd/apply.go
```diff
@@ -204,29 +200,151 @@ func applyLocalNetworkUpgrade(subnetName string, sc models.Sidecar) error {
...
+				return err
+			}
 		}
+	}
 
-		return nil
+	ux.Logger.PrintToUser("Trying to install the upgrade files at the provided %s path", avalanchegoChainConfigDir)
+	chainDir := filepath.Join(avalanchegoChainConfigDir, sc.Networks[networkKey].BlockchainID.String())
+	destPath := filepath.Join(chainDir, constants.UpgradeBytesFileName)
+	if err = os.Mkdir(chainDir, constants.DefaultPerms755); err != nil && !os.IsExist(err) {
```
felipemadero, line 280: does Mkdir expand env variables? Is the default path containing env variables?

tests/e2e/testcases/upgrade/suite.go
holisticode: Yes you provided proof that it does. I suggest that we still keep that, because in these tests all too many things are happening below the surface and they are difficult to debug. This makes it explicit with what version we are deploying.

tests/e2e/testcases/upgrade/suite.go
holisticode: I don't think so, if you look further up the chain, this is part of a `ginkgo.Describe` which sets the context.

cmd/subnetcmd/upgradecmd/apply.go
holisticode: Until we have found a way to answer these questions conclusively, yes. This keeps history and is important documentation. Next week already we will remember just half of it.

cmd/subnetcmd/upgradecmd/apply.go
holisticode: No but it is if it _is_ empty

tests/e2e/commands/upgrade.go
holisticode: Hmm not so sure about this. We are upgrading a single running node when doing public. Calling this Network is kinda misleading, we are not upgrading a network but a node. Maybe it should be called ApplyUpgradeToPublicNode?

cmd/subnetcmd/upgradecmd/apply.go
holisticode: you are right, we don't need it. I thought though it to be nice that both funcs have the same signature

cmd/subnetcmd/upgradecmd/apply.go
holisticode: Good catch. I am changing this right away in the flag configuration.

tests/e2e/commands/upgrade.go
felipemadero: I mean this is not a vm upgrade, but a network upgrade, or the way it needs to be called

felipemadero (approved):

Issue #636, opened 2023-02-23
Provide a way to isolate specific e2e tests when using the script


felipemadero:
Currently only --local versus non --local is considered. Probably use the same mechanism to isolate all tests.

PR #643, opened 2023-02-24
add --filter flag to E2E

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/636
Adds --filter flag to E2E script, for filtering based on test description. eg: ./scripts/run.e2e.sh --filter "can deploy a SubnetEvm subnet to local"

### DIFF
--- a/scripts/run.e2e.sh
+++ b/scripts/run.e2e.sh
@@ -9,6 +9,13 @@ else
     label_filter="!local_machine"
 fi
 
+description_filter=""
+if [ "$1" = "--filter" ]
+then
+    pat=$(echo $2 | tr ' ' '.')
+    description_filter="--ginkgo.focus=$pat"
+fi
+
 export RUN_E2E="true"
 
 if [ ! -d "tests/e2e/hardhat/node_modules" ]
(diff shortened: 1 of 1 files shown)


holisticode (commented):
This is a great addition and very helpful!

However, as you are using `label-filter` here, I suggest we actually really use the labeling system in our tests instead of having to then artificially hack our way around with the `tr` command. 

In fact, if I go to `tests/e2e` and run `ginkgo labels` I get:
```
e2e: No labels found
```

In other words, I think we should try to take advantage of the full power of labels

felipemadero:
> This is a great addition and very helpful!
> 
> However, as you are using `label-filter` here, I suggest we actually really use the labeling system in our tests instead of having to then artificially hack our way around with the `tr` command.
> 
> In fact, if I go to `tests/e2e` and run `ginkgo labels` I get:
> 
> ```
> e2e: No labels found
> ```
> 
> In other words, I think we should try to take advantage of the full power of labels

That would require adding label to all tests. I was trying to take advantage of the current available descriptions. Can that be made in a subsequent PR? I think we can have BOTH options and start with this one. @dasconnor what do you think?

PR #645, opened 2023-02-24
Temporarily disable failing e2e test

holisticode:
[can create and apply to locally running subnet] is currently failing due to missing expected content in the API response from subnet-vm v0.4.10.


felipemadero (commented):
could you use ginkgo.Skip ?

Issue #586, opened 2023-02-08
Avoid connecting to old backend server after updating CLI


felipemadero:
Add a ANR RPC protocol version and let the backend server return the version, if the client library has another one, give instructions to the user.
I think the goal here would be that it either fails very loudly and clearly and provides exact instructions of how to migrate from running the old server to the new server (which ideally does not require manually killing the previous server) or for the new client to seamlessly tell the old server to shutdown and then replace it by running the latest version.

PR #651, opened 2023-02-25
Check anr rpc version

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/586


pkg/binutils/processes.go
```diff
@@ -64,6 +64,22 @@ func NewGRPCClient() (client.Client, error) {
...
+		rpcVersion, err := client.RPCVersion(ctx)
+		if err != nil {
+			return nil, err
+		}
+		// obtained using server API
+		serverVersion := rpcVersion.Version
+		// obtained from ANR source code
+		clientVersion := server.RPCVersion
+		if serverVersion != clientVersion {
+			return nil, fmt.Errorf("trying to connect to a backend controller that uses a different RPC version (%d) than the CLI client (%d). Use network clean to stop the controller and then restart the operation",
```
dasconnor, line 78: hmm, do we want a way to do this that does not require clearing your deployed subnets? Maybe like a `network clean --server-only` flag that preserves your snapshot?

felipemadero: mmm do you prefer  `network clean --server-only` (need to add snapshot preservation to clean) or a `network stop --shutdown-server` (will add server shutdown to stop)

dasconnor: `network stop --shutdown-server` is better

cmd/networkcmd/stop.go
```diff
@@ -59,5 +60,12 @@ func StopNetwork(*cobra.Command, []string) error {
 		return fmt.Errorf("failed to stop network with a snapshot: %w", err)
 	}
 	ux.Logger.PrintToUser("Network stopped successfully.")
+
+	if err := binutils.KillgRPCServerProcess(app); err != nil {
+		app.Log.Warn("failed killing server process", zap.Error(err))
+	} else {
+		ux.Logger.PrintToUser("Server process terminated.")
```
dasconnor, line 67: maybe reword to `Server shutdown gracefully`

dasconnor (approved):
one small suggestion, then looks good

Issue #640, opened 2023-02-23
Trigger an error if upgrading a local network vm version under multi subnet scenario produces rpc inconsistency


felipemadero:

dasconnor:
I think we still want to perform the upgrade, we just want it to error on network restart until you get to a good configuration

PR #652, opened 2023-02-27
Choose correct avago version in network start

### LINKED ISSUE #639: Network start should select the right version based on VM RPC version

dasconnor:
This change stores local deployment status in the sidecar file and clears local deploy state when resetting the network. Using this change, the `network start` command now finds every deployed subnet and looks for a matching AvaGo version or throws an error if no match is possible.

Closes #639 #640

### DIFF
--- a/pkg/subnet/deployStatus.go
+++ b/pkg/subnet/deployStatus.go
@@ -1,0 +1,40 @@
+package subnet
+
+import (
+	"os"
+
+	"github.com/ava-labs/avalanche-cli/pkg/application"
+	"github.com/ava-labs/avalanche-cli/pkg/models"
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+)
+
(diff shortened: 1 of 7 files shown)


pkg/subnet/deployStatus.go
```diff
@@ -0,0 +1,43 @@
...
+			// don't fail on missing sidecar file, just warn
+			ux.Logger.PrintToUser("warning: inconsistent subnet directory. No sidecar file found for subnet %s", subnetDir.Name())
+			continue
+		}
+		if err != nil {
+			return nil, err
+		}
+
+		// check if sidecar contains local deployment info in Networks map
+
```
felipemadero, line 35: remove empty line

felipemadero (commented):
just a nit

pkg/subnet/deployStatus.go
dasconnor: fixed

felipemadero (approved):

PR #654, opened 2023-02-28
Fix Local Upgrades

### LINKED ISSUE #647: Subnet Upgrades aren't possible when the network is stopped

dasconnor:
Previously, you could not upgrade a VM on the local network without using command line flags. Now the upgrade wizard works correctly.

I also found a small bug in my last PR. Determining RPC version from the sidecar wouldn't work with upgrades, so I introduced a small change into the sidecar and bumped the version.

Closes #647


cmd/subnetcmd/upgradecmd/vm.go
```diff
@@ -109,6 +109,20 @@ func upgradeVM(_ *cobra.Command, args []string) error {
 		return err
 	}
 
+	// if upgrading local, check that the network is off otherwise fail here
```
felipemadero, line 112: probable also take network as off if server is off (depends on specific err return of binutils.NewGRPCClient())

felipemadero (approved):

PR #665, opened 2023-03-03
Scan logs for errors on deploy

holisticode:
If the deploy fails, this PR adds code which scans all known log files for errors.

First it applies a regex to search for all occurrences of "error" in any log file, then filters them to avoid cluttering the UI. Currently filters for 

`failing health check` (health check response, can contain the word error but due to periodical checks there are too many)
`DEBUG`                   (some debug logs contain the word error)
`WARN`                     (some warn logs contain the word error)


tests/e2e/e2e_test.go
```diff
@@ -9,15 +9,18 @@ import (
...
-	_ "github.com/ava-labs/avalanche-cli/tests/e2e/testcases/apm"
-	_ "github.com/ava-labs/avalanche-cli/tests/e2e/testcases/key"
-	_ "github.com/ava-labs/avalanche-cli/tests/e2e/testcases/network"
-	_ "github.com/ava-labs/avalanche-cli/tests/e2e/testcases/packageman"
-	_ "github.com/ava-labs/avalanche-cli/tests/e2e/testcases/root"
-	_ "github.com/ava-labs/avalanche-cli/tests/e2e/testcases/subnet"
-	_ "github.com/ava-labs/avalanche-cli/tests/e2e/testcases/subnet/local"
-	_ "github.com/ava-labs/avalanche-cli/tests/e2e/testcases/subnet/public"
-	_ "github.com/ava-labs/avalanche-cli/tests/e2e/testcases/upgrade"
+	/*
```
dasconnor, line 12: remove

tests/e2e/testcases/errhandling/suite.go
```diff
@@ -0,0 +1,49 @@
...
+	subnetName = "doFailSubnet"
+)
+
+/*
+		The tests in this suite are meant to trigger some errors so that
+	  the UI will try to find errors in the log files.
+		However, one or more tests only trigger errors on timeouts,
+		which is why it should not be run in normal CI.
+
+		Therefore the tests are `Skip`ed, but can be enabled manually for testing
```
dasconnor, line 24: how are they skipped? Just not imported in the main e22_test.go file?

pkg/utils/errors.go
```diff
@@ -0,0 +1,94 @@
...
+func FindErrorLogs(rootDirs ...string) {
+	alreadyNotified := false
+	foundErrors := false
+
+	for _, rootDir := range rootDirs {
+		if _, err := os.Stat(rootDir); err != nil {
+			return
+		}
+
+		_ = filepath.WalkDir(rootDir, func(path string, d fs.DirEntry, err error) error {
```
dasconnor, line 44: ahhh, you took a different approach than I was expecting. I thought we should just print root/node1/logs/XYZ.log. I think this code will replicate the same errors from every node, right?

Maybe default behavior should be we only look at the custom chain log from one node, then if you pass the `--debug` flag, we print everything here?

holisticode: This is reasonable. But, I have seen situations where errors appeared just in one log file, although probably less frequent. We could miss errors this way.

tests/e2e/testcases/errhandling/suite.go
holisticode: No by calling `Skip` in the test.

pkg/utils/errors.go
holisticode: Tried different approach: store found logs, and compare new ones with already found, and skip if already found. For already found, the criteria is to strip away the timestamp and compare the whole remaining string.

dasconnor (commented):
I think the code looks good. Before we merge, could put together a little demo video or test screenshot so we can see what the visual output looks like?

dasconnor (approved):

PR #673, opened 2023-03-07
Bump codeql version

dasconnor:
My editor reformatted this, but it really just bumps the codeql version.


felipemadero (commented):
Can we avoid the reformatting? Maybe is it good idea to have this file most similar to those of other repos in order to do quick comparisons.

dasconnor:
I think we should do this new formatting, this is what we've used for all our other actions and it will be harder to manage long term if we don't make the change.

felipemadero (approved):

PR #679, opened 2023-03-08
Check that at least 1 admin has balance for upgrade

### LINKED ISSUE #513: Check if TX allow list admins have balance for upgrades
https://github.com/ava-labs/avalanche-cli/issues/426 addresses the same issue but when a new subnet is being created.

This ticket is about making sure of the same, but applied to upgrades.

sukantoraymond:
Closes #513 

FOR LOCAL NETWORK ONLY

Follow up PR for Mainnet and Fuji required

Checks that at least 1 admin specified in subnet upgrade command has balance of more than 0. This check only applies to new admin being added to the precompile. 

For example, if during genesis the transaction allow precompile is configured with account A and account A has balance of 100 tokens, if the CLI user tries to upgrade the transaction allow precompile with a new admin address of account B and account B has balance 0, the `upgrade generate` command will fail because it only considers account B's balance and not account A

### DIFF
--- a/tests/e2e/assets/test_upgrade_2.json
+++ b/tests/e2e/assets/test_upgrade_2.json
@@ -1,0 +1,14 @@
+{
+  "precompileUpgrades": [
+    {
+      "txAllowListConfig": {
+        "adminAddresses": [
+          "0xF68C94B6a7000f1E695F6E49254e8Ac935009A70",
+          "0x992515DfE6246FAa8f871DDd9E42Fa4fde84d9b6"
+        ],
+        "enabledAddresses": null,
+        "blockTimestamp": 2526843696
+      }
+    }
+  ]
+}


cmd/subnetcmd/upgradecmd/generate.go
```diff
@@ -315,6 +324,80 @@ func promptTxAllowListParams(precompiles *[]params.PrecompileUpgrade, date time.
...
+		return err
+	}
+
+	for _, admin := range admins {
+		// we can break at the first admin who has a non-zero balance
+		accountBalance, err := getAccountBalance(context.Background(), cClient, admin.String())
+		if err != nil {
+			return err
+		}
+		if accountBalance > float64(0) {
```
holisticode, line 347: This is well thought but my suggestion is that the balance should be at least the minimum transaction fee. Above zero but below the fee will still result in a failure.

Frankly even just one tx fee is probably too small (the user most probably will want to ensure that more than just one tx goes through), but above the minimum transaction fee, we would have to arbitrarily set a minimum. We can quickly touch on this in standup and / or slack if you like.

cmd/subnetcmd/upgradecmd/generate.go
```diff
@@ -315,6 +324,80 @@ func promptTxAllowListParams(precompiles *[]params.PrecompileUpgrade, date time.
...
+			return nil
+		}
+	}
+
+	return errors.New("none of the addresses in the transaction allow list precompile have any tokens allocated to them. Currently, no address can transact on the network. Airdrop some funds to one of the allow list addresses to continue")
```
holisticode, line 352: This is a great description but I suggest you try to shorten it.

cmd/subnetcmd/upgradecmd/generate.go
```diff
@@ -315,6 +324,80 @@ func promptTxAllowListParams(precompiles *[]params.PrecompileUpgrade, date time.
...
+	case models.SubnetEvm:
+		// Currently only checking if admins have balance for subnets deployed in Local Network
+		if networkData, ok := sc.Networks["Local Network"]; ok {
+			blockchainID := networkData.BlockchainID.String()
+			if err = ensureAdminsHaveBalanceLocalNetwork(admins, blockchainID); err != nil {
+				return err
+			}
+		}
+	default:
+		app.Log.Warn("Unknown genesis format", zap.Any("vm-type", sc.VM))
```
holisticode, line 380: I feel this is error text is slightly incorrect. It's an unexpected VM type, in fact.

dasconnor (commented):
The code looks good! One extension I'd ask for is to look at creating an E2E test for this. Not the most crucial test in the world, but would give you a good overview of our test suite.

cmd/subnetcmd/upgradecmd/generate.go
```diff
@@ -297,12 +303,15 @@ func promptContractAllowListParams(precompiles *[]params.PrecompileUpgrade, date
...
 }
 
-func promptTxAllowListParams(precompiles *[]params.PrecompileUpgrade, date time.Time) error {
+func promptTxAllowListParams(precompiles *[]params.PrecompileUpgrade, date time.Time, subnetName string) error {
 	adminAddrs, enabledAddrs, err := promptAdminAndEnabledAddresses()
 	if err != nil {
 		return err
 	}
 
+	if err = ensureAdminsHaveBalance(adminAddrs, subnetName); err != nil {
```
holisticode, line 312: This might be a big ask, but although your solution is correct, we usually try to catch things early on. 

In this scenario, if a user enters an admin address which has no balance, the user experience will be that she gets here, and hits the error, which exits the tool.

A better experience would be that on adding the new address, the balance would be immediately checked via validation.  This would immediately prevent the address from being added, give the user instant feedback, and allow her to provide a correct address right away.

The validation API is limited so not sure it will easily work, but it's worth a try!

holisticode (commented):
Well done, the PR solves the required problem.

However, I suggest we change (or try to, at least) the workflow a bit and try early validation. See appropriate comment.

Also, please add at least one test. Either a unit test for `ensureAdminsHaveBalance`, and/or a e2e test. Ideally both, so that you also get a feel for the e2e test framework. But the e2e suite is getting big and execution time long, thus we might finally scrap that.

cmd/subnetcmd/upgradecmd/generate.go
dasconnor: I think this is a good idea, but it might be a little tricky to implement. We'd need to parse gas configs, doable but I'd be willing to backlog this

cmd/subnetcmd/upgradecmd/generate.go
dasconnor: I think this is also a good suggestion, but probably would've been more helpful in the design/issue conception stage. Feel free to address it if you want Raymond, but you can also create a ticket for the backlog since you already have a complete solution.

holisticode: Good point - but it still could be moved earlier, e.g. when the user is selecting `Done`, _before_ the prompt UI is returning.

holisticode (approved):

PR #696, opened 2023-03-14
Add telemetry tracking for CLI

### LINKED ISSUE #709: Switch from Mixpanel to Posthog
One key outcome, we want to know how much our usage will increase the cost of Posthog.

sukantoraymond:
Implements tracking user command usage with Mixpanel go sdk. Users ID are derived from hash(user name + uid), which is irreversible. 

Closes #709


dasconnor:
We might want to specify a branch here so we can merge finished PRs without going into main.

dasconnor (approved):

Issue #694, opened 2023-03-13
Upgrade to latest subnet-evm


dasconnor:
This has some breaking changes for the precompile configs.

Issue #688, opened 2023-03-10
Add enabled addresses to precompiles which support it


holisticode:
When creating a subnet-evm, currently the wizard only supports adding admins, but not enabled addresses.

PR #698, opened 2023-03-15
Update to latest subnet-evm code

holisticode:
Latest subnet-evm code introduced breaking changes to the precompiles data structures. This PR adjusts to them.
Also added enabled address support to the createEVM workflow.

Closes https://github.com/ava-labs/avalanche-cli/issues/694 and https://github.com/ava-labs/avalanche-cli/issues/688


cmd/subnetcmd/describe.go
```diff
@@ -227,6 +229,27 @@ func printPrecompileTable(genesis core.Genesis) {
 	}
 }
 
+func buildAddressTable(
```
dasconnor, line 232: Can you rewrite this method so it returns a table and doesn't have side effects?

dasconnor (approved):
One change I'd like, otherwise looks good

cmd/subnetcmd/describe.go
holisticode: renamed func to `appendToAddressTable`

PR #770, opened 2023-04-24
replace whitelisted-subnets with track-subnets in manual proceduce msg

felipemadero:
for user


dasconnor (commented):
We should add a sentence saying to migrate `whitelisted-subnets` to `track-subnets`

dasconnor (approved):

PR #776, opened 2023-04-28
add key fund command

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/777

P-Chain to P-Chain transfer among addresses managed by CLI.

It is a export to x, import at x, export to p, import at p operacion, with fee*4. From a given pchain addr to other pchain addr. Both
addr must be managed by CLI. It is a two step op so first step can be executed with a ledger in one machine, and second with another ledger in other machine.

eg: sending 0.1 avax from ledger addr index 0 in my machine to address P-fuji1p88rehlw8e45c9qrnqgyhpyjmjecqkkehyj92z, owned by another ledger on other machine

sender (owner of some addr at ledger index 0)
`avalanche key fund 0.1 P-fuji1p88rehlw8e45c9qrnqgyhpyjmjecqkkehyj92z --source --ledger 0 --fuji`

receiver (owner of addr P-fuji1p88rehlw8e45c9qrnqgyhpyjmjecqkkehyj92z with ledger index 0 of his own)
`avalanche key fund 0.1 P-fuji1p88rehlw8e45c9qrnqgyhpyjmjecqkkehyj92z --target --ledger 0 --fuji`


cmd/keycmd/fund.go
```diff
@@ -0,0 +1,289 @@
...
+		ledgerIndexFlag,
+		"i",
+		wrongLedgerIndexVal,
+		"ledger index to use for either source or target op",
+	)
+	return cmd
+}
+
+func fundKey(_ *cobra.Command, args []string) error {
+	if (!source && !target) || (source && target) {
```
dasconnor, line 113: instead of requiring a flag, prompt

cmd/keycmd/fund.go
```diff
@@ -0,0 +1,289 @@
...
+var (
+	source      bool
+	target      bool
+	keyName     string
+	ledgerIndex uint
+)
+
+func newFundCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "fund addr amount",
```
dasconnor, line 46: Normally, we would prompt for the address and amount, I don't think there's a reason to break with this convention here

dasconnor: `transfer` might be a better description than `fund`

cmd/keycmd/fund.go
```diff
@@ -0,0 +1,289 @@
...
+		if err != nil {
+			return err
+		}
+		if targetAddrStr != addrStr {
+			return fmt.Errorf("target addr inconsistency: %s vs %s", targetAddrStr, addrStr)
+		}
+		ux.Logger.PrintToUser("- receive %.9f AVAX at target address %s", float64(amount)/float64(units.Avax), addrStr)
+	}
+	ux.Logger.PrintToUser("")
+	confStr := "Are you sure you want to do it?"
```
dasconnor, line 212: `Confirm transfer`

dasconnor (commented):
We definitely need a test for this

cmd/keycmd/fund.go
```diff
@@ -0,0 +1,289 @@
...
+		if err != nil {
+			return err
+		}
+		output := &avax.TransferableOutput{
+			Asset: avax.Asset{ID: wallet.P().AVAXAssetID()},
+			Out: &secp256k1fx.TransferOutput{
+				Amt:          amount + fee*3,
+				OutputOwners: to,
+			},
+		}
```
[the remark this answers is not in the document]
felipemadero: it is expected here for the sender to pay all fees. it is for addressing situations where the target has no funds (even not  for fees) and the sender owner wants to do the funding transfer

PR #817, opened 2023-06-07
Add Permissionless Validator Fuji

sukantoraymond:


pkg/subnet/public.go
```diff
@@ -530,6 +559,44 @@ func (d *PublicDeployer) createTransformSubnetTX(
...
+	recipientAddr ids.ShortID,
+	stakeAmount uint64,
+	subnetID ids.ID,
+	nodeID ids.NodeID,
+	assetID ids.ID,
+	startTime uint64,
+	endTime uint64,
+	wallet primary.Wallet,
+) (ids.ID, error) {
+	options := d.getMultisigTxOptions([]ids.ShortID{})
```
felipemadero, line 572: mmm probably we need to be more clear on this, as there is no multisig involved here. dont? maybe change the function name

dasconnor (approved):
lgtm

cmd/subnetcmd/elastic.go
```diff
@@ -265,10 +265,7 @@ func transformElasticSubnet(_ *cobra.Command, args []string) error {
 			}
 		}
 	case models.Mainnet:
-		useLedger = true
```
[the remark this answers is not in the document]
felipemadero: This will be needed when enabled, maybe would be better to keep it to avoid forgetting about that later?

pkg/subnet/public.go
felipemadero: I should make two function now, keep the getMultisigTxOptions but make it to call another one just to set the change owner, and also use that function here

felipemadero (commented):
I am ok with the PR but would prefer those two conversations to be addressed if possible

PR #825, opened 2023-06-16
One command deploy and elastic command

sukantoraymond:
Checks for both deploy and elastic commands if previous steps are satisfied before continuing its own command.


dasconnor (approved):
Cool!

PR #835, opened 2023-06-29
add precompile address check

### LINKED ISSUE #832: Genesis precompile config sanity check
It would be great if CLI did some sanity checks on the genesis configuration before preforming the deploy.

In particular, if the addresses on the allowLists for precompiles don't include duplicate addresses in the `Admin` and `Enabled` lists. Because, if some addresses are in both lists, the chain won't start:
Screenshot 2023-06-23 at 20 17 49
And then the only way out if to abandon the Subnet and deploy a new one.

Adding other sanity checks would also be useful. There are many ways inexperienced (even those experienced, sometimes) developers may end up shooting themselves in the foot. Would be nice of us to help them avoid that if we can.

sukantoraymond:
closes #832

### DIFF
--- a/cmd/subnetcmd/deploy.go
+++ b/cmd/subnetcmd/deploy.go
@@ -131,6 +131,22 @@ func getChainsInSubnet(subnetName string) ([]string, error) {
 	return chains, nil
 }
 
+func checkDefaultAddressNotInAlloc(network models.Network, chain string) error {
+	if network != models.Local && os.Getenv(constants.SimulatePublicNetwork) == "" {
+		genesis, err := app.LoadEvmGenesis(chain)
+		if err != nil {
+			return err
+		}
+		allocAddressMap := genesis.Alloc
+		for address := range allocAddressMap {
+			if address.String() == vm.PrefundedEwoqAddress.String() {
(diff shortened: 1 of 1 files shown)


pkg/vm/precompiles.go
```diff
@@ -56,12 +57,7 @@ func configureRewardManager(app *application.Avalanche) (rewardmanager.Config, b
...
 		"https://docs.avax.network/subnets/customize-a-subnet#changing-fee-reward-mechanisms\n\n"
 
-	admins, cancelled, err := getAddressList(adminPrompt, info, app)
-	if err != nil || cancelled {
-		return config, false, err
-	}
-
-	enabled, cancelled, err := getAddressList(enabledPrompt, info, app)
+	admins, enabled, cancelled, err := getAdminAndEnabledAddresses(adminPrompt, enabledPrompt, info, app)
 	if err != nil {
```
dasconnor, line 61: should you check cancelled here?

dasconnor: eh, I guess we don't do it anywhere else really, probably fine

dasconnor (approved):
LGTM

PR #839, opened 2023-07-02
sum airdrops using the same address

### LINKED ISSUE #570: Sum airdrops to same address
If a user enters multiple airdrops for a single address in the subnet creation process, those sums aren't added. We should add them.

najeal:
This resolves and closes #570


dasconnor (dismissed):
Nice job, thanks for the assist! We can merge this when the pipeline passes and you sign the commit.

najeal:
@dasconnor lint and signed commits fixed 👍

PR #851, opened 2023-07-13
Use ledger simulator in CI mainnet e2e test

felipemadero:


tests/e2e/testcases/subnet/public/suite.go
```diff
@@ -88,9 +89,22 @@ var _ = ginkgo.Describe("[Public Subnet]", func() {
 		deploySubnetToFuji()
 	})
 
-	ginkgo.It("deploy subnet to mainnet", ginkgo.Label("local_machine"), func() {
+	ginkgo.It("deploy subnet to mainnet", func() {
+		if os.Getenv("LEDGER_SIM") != "" {
```
dasconnor, line 93: Does this enable the simulator for every test in the suite? Or is this just the only test that uses ledger?

dasconnor (commented):
Awesome job!

I'm a little confused about how this works. It might be helpful to write a little doc about how to use the simulator.

tests/e2e/testcases/subnet/public/suite.go
felipemadero: This enables the simulator only for this test, which is also currently the only one that uses ledger.
Next PR will have a multisig e2e with ledger. For that implementation probably a sequence of simulators will be started (and stopped), one per simulated ledger (one per each multisig step).

dasconnor (approved):
Excellent work on the readme

PR #865, opened 2023-07-20
Automatic cloud deploy milestone 2

### LINKED ISSUE #895: Automatic cloud deploy milestone 2
Spinning up node with terraform and ansible using existing key name and security group

sukantoraymond:
Closes #895  

Handles cases of using existing security group and key pair. This PR also enables user to define AWS credentials in ~/.aws/credentials.

If defined security group exists: 
- CLI will check if the security group in AWS contains current IP. It will add it to security group rule if current IP doesn't exist in security group. 

If defined key pair exists: 
- CLI will check if local machine contains the ssh key in ssh directory. If it doesn't, it will create new key pair using user defined key pair name.

### DIFF
--- a/cmd/nodecmd/create.go
+++ b/cmd/nodecmd/create.go
@@ -32,8 +32,10 @@ import (
 func newCreateCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create [clusterName]",
-		Short: "Create a new validator on cloud server",
-		Long: `The node create command sets up a validator on a cloud server of your choice. 
+		Short: "(ALPHA Warning) Create a new validator on cloud server",
+		Long: `(ALPHA Warning) This command is currently in experimental mode. 
+
+The node create command sets up a validator on a cloud server of your choice. 
 The validator will be validating the Avalanche Primary Network and Subnet 
 of your choice. By default, the command runs an interactive wizard. It 
 walks you through all the steps you need to set up a validator.
@@ -127,12 +129,12 @@ func printNoCredentialsOutput() {
 
 // getAWSCloudCredentials gets AWS account credentials defined in .aws dir in user home dir
 func getAWSCloudCredentials(region string) (*session.Session, error) {
+	if err := requestAWSAccountAuth(); err != nil {
+		return &session.Session{}, err
+	}
 	creds := credentials.NewSharedCredentials("", constants.AWSDefaultCredential)
 	if _, err := creds.Get(); err != nil {
 		printNoCredentialsOutput()
-		return &session.Session{}, err
-	}
-	if err := requestAWSAccountAuth(); err != nil {
 		return &session.Session{}, err
 	}
 	// Load session from shared config
--- a/pkg/application/app.go
+++ b/pkg/application/app.go
@@ -123,7 +123,7 @@ func (app *Avalanche) GetAnsibleDir() string {
 func (app *Avalanche) CreateAnsibleDir() error {
 	ansibleDir := app.GetAnsibleDir()
 	if _, err := os.Stat(ansibleDir); os.IsNotExist(err) {
-		err = os.Mkdir(ansibleDir, 0o755)
+		err = os.Mkdir(ansibleDir, constants.DefaultPerms755)
 		if err != nil {
 			return err
 		}
@@ -134,7 +134,7 @@ func (app *Avalanche) CreateAnsibleDir() error {
 func (app *Avalanche) CreateTerraformDir() error {
 	nodeTerraformDir := app.GetTerraformDir()
 	if _, err := os.Stat(nodeTerraformDir); os.IsNotExist(err) {
-		err = os.Mkdir(nodeTerraformDir, 0o755)
+		err = os.Mkdir(nodeTerraformDir, constants.DefaultPerms755)
 		if err != nil {
 			return err
 		}
@@ -145,7 +145,7 @@ func (app *Avalanche) CreateTerraformDir() error {
 func (app *Avalanche) CreateAnsibleInventoryDir() error {
 	inventoriesDir := filepath.Join(app.GetNodesDir(), constants.AnsibleInventoryDir)
 	if _, err := os.Stat(inventoriesDir); os.IsNotExist(err) {
-		err = os.Mkdir(inventoriesDir, 0o755)
+		err = os.Mkdir(inventoriesDir, constants.DefaultPerms755)
 		if err != nil {
 			return err
 		}
@@ -156,7 +156,7 @@ func (app *Avalanche) CreateAnsibleInventoryDir() error {
 func (app *Avalanche) CreateAnsiblePlaybookDir() error {
 	playbookDir := filepath.Join(app.GetAnsibleDir(), "playbook")
 	if _, err := os.Stat(playbookDir); os.IsNotExist(err) {
-		err = os.Mkdir(playbookDir, 0o755)
+		err = os.Mkdir(playbookDir, constants.DefaultPerms755)
 		if err != nil {
 			return err
 		}


cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclwrite"
+	"github.com/spf13/cobra"
+	"github.com/zclconf/go-cty/cty"
+)
+
+func newCreateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create [subnetName]",
```
dasconnor, line 33: usage is wrong

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	ux.Logger.PrintToUser("===========BEGINNING OF FILE===========")
+	ux.Logger.PrintToUser("[default]\naws_access_key_id=<AWS_ACCESS_KEY>\naws_secret_access_key=<AWS_SECRET_ACCESS_KEY>")
+	ux.Logger.PrintToUser("===========END OF FILE===========")
+	ux.Logger.PrintToUser("More info can be found at https://docs.aws.amazon.com/sdkref/latest/guide/file-format.html#file-format-creds")
+}
+
+func createNode(_ *cobra.Command, args []string) error {
+	clusterName := args[0]
+
+	err := removeExistingTerraformFiles()
```
dasconnor, line 172: do we want to actually delete these or move them to some kind of `old` directory?

dasconnor: Are they garbage or would the user ever want them?

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	if err != nil {
+		return err
+	}
+	usr, _ := user.Current()
```
dasconnor, line 176: don't ignore the error

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	alias += fmt.Sprintf("ansible_ssh_private_key_file=%s", certFilePath)
+	alias += " ansible_ssh_common_args='-o StrictHostKeyChecking=no'"
+	_, err = inventoryFile.WriteString(alias + "\n")
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func runTerraform() (string, string, error) {
```
dasconnor, line 605: You're sticking a ton in this file. I think we should move a lot of this to a new `terraform` package

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+
+func createNode(_ *cobra.Command, args []string) error {
+	clusterName := args[0]
+
+	err := removeExistingTerraformFiles()
+	if err != nil {
+		return err
+	}
+	usr, _ := user.Current()
+	keyPairName := usr.Username + "-avalanche-cli"
```
dasconnor, line 177: all these constants should be in our constants package

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	certName := keyPairName + "-kp.pem"
+	securityGroupName := keyPairName + "-sg"
+	region := "us-east-2"
+	ami := "ami-0430580de6244e02e"
+
+	hclFile := hclwrite.NewEmptyFile()
+	tfFile, err := os.Create("node_config.tf")
```
dasconnor, line 184: constant

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	if err != nil {
+		return err
+	}
+	rootBody := hclFile.Body()
+	creds := credentials.NewSharedCredentials("", "default")
```
dasconnor, line 189: We should print a warning that  tells people we're going to use their AWS credentials and ask permission before continuing

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	// Load session from shared config
+	sess, err := session.NewSession(&aws.Config{
+		Region:      aws.String(region),
+		Credentials: creds,
+	})
+	if err != nil {
+		return err
+	}
+
+	// Create new EC2 client
```
dasconnor, line 208: This should be its own method

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+		}
+		if checkCertInSSHDir(certFilePath) {
+			useExistingKeyPair = true
+		} else {
+			ux.Logger.PrintToUser(fmt.Sprintf("Default Key Pair named %s already exists", keyPairName))
+			keyPairName, err = getNewKeyPairName(ec2Svc)
+			if err != nil {
+				return err
+			}
+			certName = keyPairName + "-kp.pem"
```
dasconnor, line 230: constant

cmd/nodecmd/create.go
dasconnor: It would also be nice if we printed out some progress steps to the user that explain what we're doing. For example: `Creating Ec2 instance`

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	}
+
+	if !useExistingKeyPair {
+		err = handleCerts(certName)
+		if err != nil {
+			return err
+		}
+	}
+
+	inventoryPath := "inventories/" + clusterName
```
dasconnor, line 271: constant, app package

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+func checkCertInSSHDir(certFilePath string) bool {
+	_, err := os.Stat(certFilePath)
+	return err == nil
+}
+
+func checkCurrentIPInSg(sg *ec2.SecurityGroup, currentIP string) (bool, bool) {
+	var ipInTCP bool
+	var ipInHTTP bool
+	for _, ip := range sg.IpPermissions {
+		if *ip.FromPort == 22 || *ip.FromPort == 9650 {
```
dasconnor, line 332: constants

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+			}
+		}
+	}
+	return ipInTCP, ipInHTTP
+}
+
+func setCloudCredentials(rootBody *hclwrite.Body, accessKey, secretKey, region string) error {
+	provider := rootBody.AppendNewBlock("provider", []string{"aws"})
+	providerBody := provider.Body()
+	providerBody.SetAttributeValue("access_key", cty.StringVal(accessKey))
```
dasconnor, line 351: Does this cause secrets to be written to files unknowningly? Is this the best way to handle auth?

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+		return "", err
+	}
+	ipAddress := string(ipOutput)
+	if net.ParseIP(ipAddress) == nil {
+		return "", errors.New("invalid IP address")
+	}
+	return ipAddress, nil
+}
+
+func setSecurityGroup(rootBody *hclwrite.Body, ipAddress, securityGroupName string) {
```
dasconnor, line 407: We will probably want to run all this by infra and security before we include this in a release. We should also audit a sample deployment.

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	inboundGroupBody.SetAttributeValue("description", cty.StringVal("Outbound traffic"))
+	inboundGroupBody.SetAttributeValue("from_port", cty.NumberIntVal(0))
+	inboundGroupBody.SetAttributeValue("to_port", cty.NumberIntVal(0))
+	inboundGroupBody.SetAttributeValue("protocol", cty.StringVal("-1"))
+	ipList = []cty.Value{}
+	ipList = append(ipList, cty.StringVal("0.0.0.0/0"))
+	inboundGroupBody.SetAttributeValue("cidr_blocks", cty.ListVal(ipList))
+}
+
+func setSecurityGroupRule(rootBody *hclwrite.Body, ipAddress, sgID string, ipInTCP, ipInHTTP bool) {
```
dasconnor, line 464: We're going to need extensive documentation on exactly what we're setting up for the customer. We might want to start it now.

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+		securityGroupRuleBody.SetAttributeValue("to_port", cty.NumberIntVal(22))
+		securityGroupRuleBody.SetAttributeValue("protocol", cty.StringVal("tcp"))
+		var ipList []cty.Value
+		ipList = append(ipList, cty.StringVal(inputIPAddress))
+		securityGroupRuleBody.SetAttributeValue("cidr_blocks", cty.ListVal(ipList))
+		securityGroupRuleBody.SetAttributeValue("security_group_id", cty.StringVal(sgID))
+	}
+	if !ipInHTTP {
+		sgRuleName := "ipHttp" + strings.ReplaceAll(ipAddress, ".", "")
+		// sgRuleName := "ipHttp"
```
dasconnor, line 481: remove

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+		securityGroupRuleBody.SetAttributeValue("to_port", cty.NumberIntVal(9650))
+		securityGroupRuleBody.SetAttributeValue("protocol", cty.StringVal("tcp"))
+		var ipList []cty.Value
+		ipList = append(ipList, cty.StringVal(inputIPAddress))
+		securityGroupRuleBody.SetAttributeValue("cidr_blocks", cty.ListVal(ipList))
+		securityGroupRuleBody.SetAttributeValue("security_group_id", cty.StringVal(sgID))
+	}
+}
+
+func setElasticIP(rootBody *hclwrite.Body) {
```
dasconnor, line 495: Aren't nodes supposed to be configured with static IPs?

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+		hcl.TraverseAttr{
+			Name: "fuji_node[0]",
+		},
+		hcl.TraverseAttr{
+			Name: "id",
+		},
+	})
+}
+
+func createAnsibleHostInventory(inventoryPath, elasticIP, certFilePath string) error {
```
dasconnor, line 582: Maybe this goes in an `ansible` package

main.yml
dasconnor, line 1: Could we name this file something else or put it in a subfolder that describes what it's for?

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+
+func createNodeConfig(nodeID, region, ami, keyPairName, certPath, sg, eip, clusterName string) error {
+	elasticIPToUse := eip[1 : len(eip)-2]
+	nodeIDToUse := nodeID[1 : len(nodeID)-2]
+
+	nodeConfig := models.NodeConfig{
+		NodeID:        nodeIDToUse,
+		Region:        region,
+		AMI:           ami,
+		KeyPair:       keyPairName,
```
felipemadero, line 88: Do KeyPair and Cert refer to the same thing? If this is the case can we unify naming?

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	if clusterConfig.KeyPair == nil {
+		clusterConfig.KeyPair = make(map[string]string)
+	}
+	if _, ok := clusterConfig.KeyPair[keyPairName]; !ok {
+		clusterConfig.KeyPair[keyPairName] = certPath
```
felipemadero, line 115: Indeed it seems KeyPair and cert refer to the same thing

cmd/nodecmd/create.go
felipemadero: this should live in .avalanche-cli

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+		return err
+	}
+	rootBody := hclFile.Body()
+	creds := credentials.NewSharedCredentials("", "default")
+	credValue, err := creds.Get()
+	if err != nil {
+		printNoCredentialsOutput()
+		return err
+	}
+	err = setCloudCredentials(rootBody, credValue.AccessKeyID, credValue.SecretAccessKey, region)
```
felipemadero: probably all stuff related to creating the terraform config file should have its own function, and probably this aux functions may also indicate that they are updating the terraform config file

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	var stdBuffer bytes.Buffer
+	cmd := exec.Command("ansible-playbook", "main.yml", "-i", inventoryPath, "--ssh-extra-args='-o IdentitiesOnly=yes'")
+	mw := io.MultiWriter(os.Stdout, &stdBuffer)
+	cmd.Stdout = mw
+	cmd.Stderr = mw
+	return cmd.Run()
+}
+
+func PrintResults(instanceID, elasticIP, certFilePath, region string) {
+	instanceIDToUse := instanceID[1 : len(instanceID)-2]
```
felipemadero, line 389: same here

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	cmd := exec.Command("ssh-add", certFilePath)
+	mw := io.MultiWriter(os.Stdout, &stdBuffer)
+	cmd.Stdout = mw
+	cmd.Stderr = mw
+	return cmd.Run()
+}
+
+func runAnsiblePlaybook(inventoryPath string) error {
+	var stdBuffer bytes.Buffer
+	cmd := exec.Command("ansible-playbook", "main.yml", "-i", inventoryPath, "--ssh-extra-args='-o IdentitiesOnly=yes'")
```
felipemadero, line 381: main.yml should be generated by the tool, not preexisting

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	mw := io.MultiWriter(os.Stdout, &stdBuffer)
```
felipemadero, line 382: don't understand the usage of multiwriter here, what happens with stdBuffer?

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+func handleCerts(certName string) error {
+	err := os.Chmod(certName, 0o400)
+	if err != nil {
+		return err
+	}
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return err
+	}
+	certFilePath := homeDir + "/.ssh/" + certName
```
felipemadero, line 366: don't we have a function for this?

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	eipOutput, err := exec.Command("terraform", "output", "instance_eip").Output()
+	if err != nil {
+		return "", "", err
+	}
+	elasticIP = string(eipOutput)
+	return instanceID, elasticIP, nil
+}
+
+func handleCerts(certName string) error {
+	err := os.Chmod(certName, 0o400)
```
felipemadero, line 358: if certName a path?

felipemadero: can we change the param name and also make it clear in caller code because it seems is only cert name

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+		return err
+	}
+	certFilePath := homeDir + "/.ssh/" + certName
+	err = os.Rename(certName, certFilePath)
+	if err != nil {
+		return err
+	}
+	var stdBuffer bytes.Buffer
+	cmd := exec.Command("ssh-add", certFilePath)
+	mw := io.MultiWriter(os.Stdout, &stdBuffer)
```
felipemadero, line 373: same comment for multiwrite here

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	return nil
+}
+
+func runTerraform() (string, string, error) {
+	var instanceID string
+	var elasticIP string
+	if err := exec.Command("terraform", "init").Run(); err != nil {
+		return "", "", err
+	}
+	var stdBuffer bytes.Buffer
```
felipemadero: maybe we can move this multiwriter stuff to its own exec function

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+		},
+		hcl.TraverseAttr{
+			Name: "id",
+		},
+	})
+}
+
+func createAnsibleHostInventory(inventoryPath, elasticIP, certFilePath string) error {
+	if err := os.MkdirAll(inventoryPath, os.ModePerm); err != nil {
+		log.Fatal(err)
```
felipemadero, line 336: why Fatal?

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+	}
+	inventoryHostsFile := inventoryPath + "/hosts"
+	inventoryFile, err := os.Create(inventoryHostsFile)
+	if err != nil {
+		return err
+	}
+	alias := "aws-node "
+	elasticIPToUse := elasticIP[1 : len(elasticIP)-2]
```
felipemadero, line 344: same

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+func createAnsibleHostInventory(inventoryPath, elasticIP, certFilePath string) error {
+	if err := os.MkdirAll(inventoryPath, os.ModePerm); err != nil {
+		log.Fatal(err)
+	}
+	inventoryHostsFile := inventoryPath + "/hosts"
+	inventoryFile, err := os.Create(inventoryHostsFile)
+	if err != nil {
+		return err
+	}
+	alias := "aws-node "
```
felipemadero, line 343: all this formatting can go into single fmt.Sprintf

cmd/nodecmd/create.go
felipemadero: inside .avalanche-cli

cmd/nodecmd/create.go
dasconnor: We shouldn't be using this pattern to access AWS secrets, there are several methods to do this. We should probably first check environment variables and second do this method https://registry.terraform.io/providers/hashicorp/aws/latest/docs#shared-configuration-and-credentials-files

dasconnor: We actually might want to try to match the entire AWS CLI credentials precedence stack https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html#cli-configure-quickstart-precedence

dasconnor: As far as how we accept credentials that is. What we don't want to do is expose them in a way that the caller did not intend or expect. I think we really want to avoid writing them to files or anything like that. It'd be easy for a bug to accidentally not delete those files and then the user has credentials lying around.

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,377 @@
...
+	ux.Logger.PrintToUser("AvalancheGo and Avalanche-CLI installed and node is bootstrapping!")
+	err = terraform.RemoveExistingTerraformFiles()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func requestAWSAccountAuth() error {
+	confirm := "Do you authorize Avalanche-CLI to access your AWS account to set-up your Avalanche Validator node?"
```
dasconnor, line 263: Should probably add a bit more scare language to say that this will result in charges to their account and Ava Labs is not liable for any bills incurred.

dasconnor: We probably want to run by legal eventually, so not a super big deal atm

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,377 @@
...
+	if err != nil {
+		return err
+	}
+	if !yes {
+		return errors.New("user did not give authorization to Avalanche-CLI to access AWS account")
+	}
+	return nil
+}
+
+func checkKeyPairExists(ec2Svc *ec2.EC2, kpName string) (bool, error) {
```
dasconnor, line 274: This probably should also be in a separate package

pkg/ansible/ansible.go
```diff
@@ -0,0 +1,46 @@
...
+	if err := os.MkdirAll(inventoryPath, os.ModePerm); err != nil {
+		return err
+	}
+	inventoryHostsFile := inventoryPath + "/hosts"
+	inventoryFile, err := os.Create(inventoryHostsFile)
+	if err != nil {
+		return err
+	}
+	alias := "aws-node "
+	elasticIPToUse := elasticIP[1 : len(elasticIP)-2]
```
dasconnor, line 26: what are you stripping? comment?

cmd/nodecmd/create.go
dasconnor: If keeping them is a security issue, then we shouldn't be generating them like that in the first place

dasconnor (commented):
Much better structure. Main concern now is that we clean up AWS key storage. I'm also slightly concerned about how we're managing ssh keys. We probably want to throw up similar warnings around that

.gitignore
```diff
@@ -20,7 +20,11 @@
 
 # ignore GoLand metafiles directory
 .idea/
-
+.terraform/
+.terraform.lock.hcl
+node_config.tf
```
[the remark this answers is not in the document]
felipemadero: Temporary files should be created in temporary locations, not being created in current user working dir. At least, that is not so clean usually for a tool. If not .avalanche-cli, maybe /tmp/. Sensitive content is another issue. Would like to know @dasconnor ideas on this

ansible.cfg
```diff
@@ -0,0 +1,2 @@
+[defaults]
+host_key_checking = False
```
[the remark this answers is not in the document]
felipemadero: ok you can put the file in a desired location, like ./avalanche-cli/conf or temporary location, change the working directory for ansible in cmd.Dir field previously execution.

cmd/nodecmd/create.go
felipemadero: is it subnetName or clusterName?

pkg/application/app.go
```diff
@@ -194,6 +206,18 @@ func (app *Avalanche) GenesisExists(subnetName string) bool {
 	return err == nil
 }
 
+func (app *Avalanche) NodeDirExists() bool {
+	nodePath := app.GetNodeDir()
+	_, err := os.Stat(nodePath)
```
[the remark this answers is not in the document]
felipemadero: The function is unused

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,678 @@
...
+		}
+		if !keyPairExists {
+			return newKeyPairName, nil
+		}
+		ux.Logger.PrintToUser(fmt.Sprintf("Key Pair named %s already exists", newKeyPairName))
+	}
+}
+
+func removeExistingTerraformFiles() error {
+	nodeConfigFile := "node_config.tf"
```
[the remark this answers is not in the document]
felipemadero: probably move it to a temporary dir

cmd/nodecmd/create.go
felipemadero: this should live in temporary location, setting cmd.Dir for the terraform cmd

cmd/nodecmd/create.go
felipemadero: we should download playbooks and install them on .avalanche-cli. don't depending on the repo

felipemadero: or have constants for each playbook and made cli to create it live, whichever bests. what do you think @dasconnor ?

cmd/nodecmd/create.go
felipemadero: thanks! I was doing a lot of crazy stuff for the same thing

cmd/nodecmd/create.go
felipemadero: I continue not understanding why a certName is passed onto os.Chmod as if it were a path

cmd/nodecmd/create.go
felipemadero: oh this multiwriter stuff is being used 3 times for the same. and can also we used in e2e instead of a tricky goroutine. so a function will be appreciate now

cmd/nodecmd/create.go
felipemadero: could you move it to temporary and set cmd.Dir

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,377 @@
...
+		return err
+	}
+	usr, err := user.Current()
+	if err != nil {
+		return err
+	}
+	keyPairName := usr.Username + constants.AvalancheCLISuffix
+	certName := keyPairName + constants.CertSuffix
+	securityGroupName := keyPairName + constants.AWSSecurityGroupSuffix
+	region := "us-east-2"
```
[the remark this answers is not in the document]
dasconnor: I was referring to the region, resolving since we will address later

cmd/nodecmd/create.go
```diff
@@ -252,15 +298,12 @@ func createNode(_ *cobra.Command, args []string) error {
...
-	err = terraform.RemoveExistingTerraformFiles()
-	if err != nil {
-		return err
-	}
 	return nil
 }
 
 func requestAWSAccountAuth() error {
-	confirm := "Do you authorize Avalanche-CLI to access your AWS account to set-up your Avalanche Validator node?"
+	confirm := "Do you authorize Avalanche-CLI to access your AWS account to set-up your Avalanche Validator node? " +
```
dasconnor, line 305: We can do this in a separate ticket if you want, but I'd like to list everything that command does before they accept here so they know what they're signing up for.

pkg/ansible/ansible.go
```diff
@@ -37,10 +40,18 @@ func CreateAnsibleHostInventory(inventoryPath, elasticIP, certFilePath string) e
...
-	mw := io.MultiWriter(os.Stdout, &stdBuffer)
-	cmd.Stdout = mw
-	cmd.Stderr = mw
+	utils.SetUpMultiWrite(cmd)
 	return cmd.Run()
 }
+
+func CheckIsInstalled() error {
+	if err := exec.Command(constants.AnsiblePlaybook).Run(); errors.Is(err, exec.ErrNotFound) { //nolint:gosec
+		ux.Logger.PrintToUser("Ansible tool is not available. It is a needed dependency for CLI to set up a remote node.")
```
dasconnor, line 50: needed -> necessary

pkg/terraform/terraform.go
```diff
@@ -35,13 +39,11 @@ func SaveTerraformFile(tfFile *os.File, hclFile *hclwrite.File) error {
...
 
-func SetCloudCredentials(rootBody *hclwrite.Body, accessKey, secretKey, region string) error {
+func SetCloudCredentials(rootBody *hclwrite.Body, region string) error {
 	provider := rootBody.AppendNewBlock("provider", []string{"aws"})
 	providerBody := provider.Body()
-	providerBody.SetAttributeValue("access_key", cty.StringVal(accessKey))
-	providerBody.SetAttributeValue("secret_key", cty.StringVal(secretKey))
 	providerBody.SetAttributeValue("region", cty.StringVal(region))
-
+	providerBody.SetAttributeValue("profile", cty.StringVal("default"))
```
dasconnor, line 46: You might actually want to make the profile configurable. You can do this in another issue if you want.

dasconnor (commented):
Nice work. The code is getting much more clear and readable. One comment I have though is that we're adding **a lot** of new code here and we haven't added any new tests (unit or e2e). I understand that testing on AWS presents its own issues, but there are also other solutions we could try like localstack if we want to get creative. 

I'm fine adding testing in a follow-on PR (this one is honestly too big to reason about), but I think we should prioritize that before adding more new features.

ansible.cfg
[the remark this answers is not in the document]
felipemadero: Imagine a user installs cli with the install script. Where is ansible.cfg and playbooks going to be found?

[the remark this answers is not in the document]
felipemadero: I propose to:
- have a golang constant for the contents of ansible.cfg
- keep ansible playbooks as files but use go:embed to put the files into golang constants
- so cli having golang constant for both, can generate those files at execution time, under .avalanche-cli/nodes/ansible/
- set ansible cmd cmd.Dir to .avalanche-cli/nodes/ansible/ previously execution

cmd/nodecmd/create.go
felipemadero: I sugest follow https://github.com/ava-labs/avalanche-cli/pull/865#discussion_r1284906244

felipemadero (commented):
Really great job. Lot of things happening here. But needs a little more to be more readable.
After addressing commnet LGTM

felipemadero (commented):
Really great job. Lot of things happening here. But needs a little more to be more readable.
After addressing commnet LGTM

pkg/ansible/ansible.go
```diff
@@ -0,0 +1,54 @@
...
+	"os"
+	"os/exec"
+
+	"github.com/ava-labs/avalanche-cli/pkg/utils"
+
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+)
+
+func CreateAnsibleHostInventory(inventoryPath, elasticIP, certFilePath string) error {
```
Nuttymoon, line 18: Terraform has a templating feature. This might be easier than managing the file with Go. You can see an example of how we do things in Ash:
- [hosts.tftpl](https://github.com/AshAvalanche/ansible-avalanche-getting-started/blob/main/terraform/multipass/hosts.tftpl) (template)
- [local_file resource definition](https://github.com/AshAvalanche/ansible-avalanche-getting-started/blob/eae729cdcaa20382e8ab58849e11f294aefacfba/terraform/multipass/main.tf#L57) in the Terraform file

playbook/setupNode.yml
```diff
@@ -0,0 +1,9 @@
+---
+- hosts: all
+  tasks:
+    - name: get avalanche go script
+      shell: wget -nd -m https://raw.githubusercontent.com/ava-labs/avalanche-docs/master/scripts/avalanchego-installer.sh
+    - name: modify permissions
+      shell: chmod 755 avalanchego-installer.sh
+    - name: call avalanche go install script
+      shell: ./avalanchego-installer.sh --ip static --rpc private --state-sync on --fuji
```
Nuttymoon, line 9:
```suggestion
      command: ./avalanchego-installer.sh --ip static --rpc private --state-sync on --fuji
```
Same reason as above

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,393 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/spf13/cobra"
+)
+
+func newCreateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create [clusterName]",
```
patrick-ogrady, line 34: Can you add an "ALPHA" warning to both the SHORT and LONG fields here saying this is still undergoing testing?

cmd/nodecmd/create.go
```diff
@@ -0,0 +1,393 @@
...
+	err := terraform.RemoveDirectory(app.GetTerraformDir())
+	if err != nil {
+		return err
+	}
+	usr, err := user.Current()
+	if err != nil {
+		return err
+	}
+	region := "us-east-2"
+	ami := "ami-0430580de6244e02e"
```
patrick-ogrady, line 278: why is this hardcoded? can't we just use the latest ubuntu image?

felipemadero (approved):
Great Job. LGTM. After addressing other people's comments.

cmd/nodecmd/create.go
```diff
@@ -122,19 +125,14 @@ func printNoCredentialsOutput() {
...
 		printNoCredentialsOutput()
 		return &session.Session{}, err
 	}
-	err = requestAWSAccountAuth()
-	if err != nil {
-		return &session.Session{}, err
-	}
-	err = terraform.SetCloudCredentials(rootBody, region)
-	if err != nil {
+	if err := requestAWSAccountAuth(); err != nil {
```
dasconnor, line 135: Should the order of this and the `creds.Get()` check above be flipped?

pkg/application/app.go
```diff
@@ -113,11 +113,26 @@ func (app *Avalanche) GetConfigPath() string {
...
-func (app *Avalanche) CreateNodeTerraformDir() error {
-	nodeTerraformDir := app.GetNodeTerraformDir()
+func (app *Avalanche) GetAnsibleDir() string {
+	return filepath.Join(app.GetNodesDir(), constants.AnsibleDir)
+}
+
+func (app *Avalanche) CreateAnsibleDir() error {
+	ansibleDir := app.GetAnsibleDir()
+	if _, err := os.Stat(ansibleDir); os.IsNotExist(err) {
+		err = os.Mkdir(ansibleDir, 0o755)
```
dasconnor, line 126: we have a constant for these perms

pkg/application/app.go
```diff
@@ -127,12 +142,38 @@ func (app *Avalanche) CreateNodeTerraformDir() error {
 	return nil
 }
 
-func (app *Avalanche) GetNodeTerraformDir() string {
-	return filepath.Join(app.GetNodeDir(), constants.TerraformDir)
+func (app *Avalanche) CreateAnsibleInventoryDir() error {
+	inventoriesDir := filepath.Join(app.GetNodesDir(), constants.AnsibleInventoryDir)
+	if _, err := os.Stat(inventoriesDir); os.IsNotExist(err) {
+		err = os.Mkdir(inventoriesDir, 0o755)
```
dasconnor, line 148: few more occurences here

dasconnor (commented):
Couple nits, then should be good to go

dasconnor (approved):
LGTM

PR #938, opened 2023-08-16
Stop cloud server

### LINKED ISSUE #873: Enable users to stop cloud server

sukantoraymond:
closes #873

### DIFF
--- a/cmd/nodecmd/stop.go
+++ b/cmd/nodecmd/stop.go
@@ -89,38 +89,39 @@ func stopNode(_ *cobra.Command, args []string) error {
 			return err
 		}
 	}
-	if clusterNodes, ok := clusterConfig.Clusters[clusterName]; ok {
-		if len(clusterNodes) == 0 {
-			return fmt.Errorf("no nodes found in cluster %s", clusterName)
-		}
-		nodeConfig, err := app.LoadClusterNodeConfig(clusterNodes[0])
-		if err != nil {
-			return err
-		}
-		if err = getDeleteConfigConfirmation(nodeConfig.NodeID); err != nil {
-			return err
-		}
-		sess, err := getAWSCloudCredentials(nodeConfig.Region)
-		if err != nil {
-			return err
-		}
-		ec2Svc := ec2.New(sess)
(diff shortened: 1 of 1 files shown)


cmd/nodecmd/stop.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+
+	"github.com/spf13/cobra"
+)
+
+func newStopCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "stop [clusterName]",
+		Short: "(ALPHA Warning) Stop all nodes in a cluster",
+		Long: `(ALPHA Warning) This command is currently in experimental mode.
```
dasconnor, line 23: You should add a note that stopping nodes may be insufficient to stop future charges and you may still be incur further costs

dasconnor (commented):
One small comment then lgtm

cmd/nodecmd/stop.go
```diff
@@ -0,0 +1,124 @@
...
+		delete(clusterConfig.Clusters, clusterName)
+	}
+	return app.WriteClusterConfigFile(&clusterConfig)
+}
+
+func removeDeletedNodeDirectory(clusterName string) error {
+	return os.RemoveAll(app.GetNodeInstanceDirPath(clusterName))
+}
+
+func removeClusterInventoryDir(clusterName string) error {
```
felipemadero, line 55: this should be remove node from cluster inventory dir (similar in spirit to remove node from cluster config). can we add an issue for this? probably the inventory file should be regenerated based on the contents of the cluster config

cmd/nodecmd/stop.go
```diff
@@ -0,0 +1,124 @@
...
+	clusterName := args[0]
+	var err error
+	clusterConfig := models.ClusterConfig{}
+	if app.ClusterConfigExists() {
+		clusterConfig, err = app.LoadClusterConfig()
+		if err != nil {
+			return err
+		}
+	}
+	clusterNodes := clusterConfig.Clusters[clusterName]
```
felipemadero, line 92: doesn't it panic if there is no cluster config?

felipemadero (approved):
LGTM. Suggest wait also for connor approval for all this PRs

PR #941, opened 2023-08-21
Split join cmd into validateSubnet and validatePrimaryNetwork

sukantoraymond:


dasconnor:
Why not make them sub commands? `validate primary` and `validate subnet`?

cmd/nodecmd/validatePrimaryNetwork.go
```diff
@@ -30,38 +28,18 @@ import (
...
 		Long: `(ALPHA Warning) This command is currently in experimental mode.
 
-The node join command enables all nodes in a cluster to be validators of a Subnet.
-The command also enables nodes to be Primary Network validators by omitting the --subnet flag.
-If The command is run before the nodes are Primary Network validators, the command will first
-make the nodes Primary Network validators before making them Subnet validators. 
-If The command is run before the nodes are bootstrapped on the Primary Network, the command will fail. 
-You can check the bootstrap status by calling avalanche node status <clusteName>`,
+The node validatePrimaryNetwork command enables all nodes in a cluster to be validators of Primary
+Network.`,
```
dasconnor, line 38: Just curious, what happens if you call this command multiple times? What if you call it, add another node, then call it again?

cmd/nodecmd/validateSubnet.go
```diff
@@ -0,0 +1,181 @@
...
+)
+
+func newValidateSubnetCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "validateSubnet [clusterName]",
+		Short: "(ALPHA Warning) Join a subnet as a validator",
+		Long: `(ALPHA Warning) This command is currently in experimental mode.
+
+The node validateSubnet command enables all nodes in a cluster to be validators of a Subnet.
+If The command is run before the nodes are Primary Network validators, the command will first
```
dasconnor, line 45: capitalization `The`

dasconnor: would it make more sense to fail if you're not already a primary network validator? Genuine question, not a binding suggestion

PR #943, opened 2023-08-22
Check avalanchego compatibility with subnet evm rpc when tracking subnet

### LINKED ISSUE #906: Handle invalid avalanche go in validator based on subnet vm and vice versa

sukantoraymond:
Closes #906

### DIFF
--- a/cmd/nodecmd/sync.go
+++ b/cmd/nodecmd/sync.go
@@ -81,9 +81,9 @@ func parseAvalancheGoOutput(fileName string) (string, error) {
 	if ok {
 		vmVersions, ok := nodeIDInterface["vmVersions"].(map[string]interface{})
 		if ok {
-			version, ok := vmVersions["platform"].(string)
+			avalancheGoVersion, ok := vmVersions["platform"].(string)
 			if ok {
-				return version, nil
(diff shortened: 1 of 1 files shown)


cmd/nodecmd/sync.go
```diff
@@ -52,9 +59,80 @@ func syncSubnet(_ *cobra.Command, args []string) error {
...
+
+	var result map[string]interface{}
+	if err = json.Unmarshal(byteValue, &result); err != nil {
+		return "", err
+	}
+	nodeIDInterface, ok := result["result"].(map[string]interface{})
+	if ok {
+		vmVersions, ok := nodeIDInterface["vmVersions"].(map[string]interface{})
+		if ok {
+			avalancheGoVersion, ok := vmVersions["platform"].(string)
```
felipemadero, line 84: should we use result/version ?

felipemadero (commented):
only one question. after that, LGTM

cmd/nodecmd/sync.go
felipemadero: I mean, instead of using the version of vm platform in result/vmVersions/platform, to use the root version in results/version

felipemadero (approved):

PR #954, opened 2023-08-24
Delete instance without public ip if account limit reached

sukantoraymond:
There is limit of 5 Elastic IP in a region per AWS account

### DIFF
--- a/cmd/nodecmd/create.go
+++ b/cmd/nodecmd/create.go
@@ -321,14 +321,17 @@ func createNode(_ *cobra.Command, args []string) error {
 	// Create new EC2 client
 	instanceID, elasticIP, certFilePath, keyPairName, err := createEC2Instance(rootBody, ec2Svc, hclFile, region, ami, certName, prefix, securityGroupName)
 	if err != nil {
-		ux.Logger.PrintToUser("Failed to create AWS cloud server, please try creating again in a different region")
-		ux.Logger.PrintToUser("Stopping AWS cloud server... ")
-		// we stop created instance if EIP is not created due to limit reached of 5 EIP per AWS account
-		// so that user doesn't pay for unused EC2 instance
+		if err.Error() == constants.EIPLimitErr {
+			ux.Logger.PrintToUser("Failed to create AWS cloud server, please try creating again in a different region")
+		} else {
+			ux.Logger.PrintToUser("Failed to create AWS cloud server")
+		}
+		// we stop created instance so that user doesn't pay for unused EC2 instance
 		instanceID, instanceIDErr := terraform.GetInstanceID(app.GetTerraformDir())
 		if instanceIDErr != nil {
 			return instanceIDErr
 		}
+		ux.Logger.PrintToUser(fmt.Sprintf("Stopping AWS cloud server %s...", instanceID))
 		if stopErr := awsAPI.StopInstance(ec2Svc, instanceID, "", false); stopErr != nil {
 			ux.Logger.PrintToUser(fmt.Sprintf("Failed to stop cloud server instance %s", instanceID))
 			ux.Logger.PrintToUser(fmt.Sprintf("Stop cloud server instance %s on AWS console to prevent charges", instanceID))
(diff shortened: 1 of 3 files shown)


pkg/aws/aws.go
```diff
@@ -121,6 +121,16 @@ func CheckInstanceIsRunning(ec2Svc *ec2.EC2, nodeID string) (bool, error) {
 	return false, nil
 }
 
+func StopInstanceWoEIP(ec2Svc *ec2.EC2, instanceID string) error {
```
felipemadero, line 124: can we add a bool to StopIntance to avoid releasing ip? so as to have one function only

cmd/nodecmd/create.go
```diff
@@ -321,6 +321,19 @@ func createNode(_ *cobra.Command, args []string) error {
...
+		ux.Logger.PrintToUser("Failed to create AWS cloud server, please try creating again in a different region")
+		ux.Logger.PrintToUser("Stopping AWS cloud server without Public IP... ")
+		// we stop created instance if EIP is not created due to limit reached of 5 EIP per AWS account
+		// so that user doesn't pay for unused EC2 instance
+		instanceID, instanceIDErr := terraform.GetInstanceID(app.GetTerraformDir())
+		if instanceIDErr != nil {
+			return instanceIDErr
+		}
+		if stopErr := awsAPI.StopInstanceWoEIP(ec2Svc, instanceID); stopErr != nil {
+			ux.Logger.PrintToUser(fmt.Sprintf("Failed to stop cloud server instance %s without Public IP", instanceID))
```
felipemadero, line 333: I would not say "without Public IP" probably confuses the user

cmd/nodecmd/create.go
```diff
@@ -321,6 +321,19 @@ func createNode(_ *cobra.Command, args []string) error {
 	// Create new EC2 client
 	instanceID, elasticIP, certFilePath, keyPairName, err := createEC2Instance(rootBody, ec2Svc, hclFile, region, ami, certName, prefix, securityGroupName)
 	if err != nil {
+		ux.Logger.PrintToUser("Failed to create AWS cloud server, please try creating again in a different region")
+		ux.Logger.PrintToUser("Stopping AWS cloud server without Public IP... ")
```
felipemadero, line 325: I would not say "without Public IP"

cmd/nodecmd/create.go
```diff
@@ -321,6 +321,19 @@ func createNode(_ *cobra.Command, args []string) error {
 	// Create new EC2 client
 	instanceID, elasticIP, certFilePath, keyPairName, err := createEC2Instance(rootBody, ec2Svc, hclFile, region, ami, certName, prefix, securityGroupName)
 	if err != nil {
+		ux.Logger.PrintToUser("Failed to create AWS cloud server, please try creating again in a different region")
```
felipemadero, line 324: maybe you suggest to use a different region, but maybe the solution is other one? how can we sure
the solution is to use a different region?

felipemadero: ie: can we check that the error is EIP limit? can createEC2Instance notify that?

pkg/terraform/terraform.go
```diff
@@ -245,36 +245,58 @@ func RemoveDirectory(terraformDir string) error {
...
-	var instanceID string
-	var publicIP string
 	cmd := exec.Command(constants.Terraform, "init") //nolint:gosec
 	cmd.Dir = terraformDir
 	if err := cmd.Run(); err != nil {
 		return "", "", err
 	}
 	cmd = exec.Command(constants.Terraform, "apply", "-auto-approve") //nolint:gosec
 	cmd.Dir = terraformDir
-	utils.SetupRealtimeCLIOutput(cmd)
```
felipemadero, line 257: can we make SetupRealtimeCLIOutput to setup two multiwriters, one for os.Stdout + stdoutBuffer and another
for os.Stderr + stderrBuffer, and return the two buffers?
Then we can assing one to stderrBuffer (not stdErr) and continue from there

felipemadero (commented):
Last one

felipemadero (approved):

PR #993, opened 2023-09-05
Multi Nodes Validate Primary Network & Subnets

sukantoraymond:
Changes are mainly : 
- change Ansible Playbook so that it targets specific host instead of all hosts in the Ansible Inventory File
- change from single to multi nodes for validate Primary network, validate Subnet and sync Subnet

### DIFF
--- a/cmd/nodecmd/validatePrimary.go
+++ b/cmd/nodecmd/validatePrimary.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/ava-labs/avalanchego/genesis"
@@ -29,15 +30,15 @@ import (
 )
 
 var (
-	deployTestnet   bool
-	deployMainnet   bool
-	keyName         string
-	subnetName      string
-	useLedger       bool
-	ledgerAddresses []string
-	weight          uint64
-	duration        time.Duration
-
+	deployTestnet                bool
+	deployMainnet                bool
+	keyName                      string
+	subnetName                   string
+	useLedger                    bool
+	ledgerAddresses              []string
+	weight                       uint64
+	duration                     time.Duration
+	useCustomDuration            bool
 	ErrMutuallyExlusiveKeyLedger = errors.New("--key and --ledger,--ledger-addrs are mutually exclusive")
 	ErrStoredKeyOnMainnet        = errors.New("--key is not available for mainnet operations")
 	ErrNoBlockchainID            = errors.New("failed to find the blockchain ID for this subnet, has it been deployed/created on this network?")
@@ -128,7 +129,7 @@ func getMinStakingAmount(network models.Network) (uint64, error) {
 	return minValStake, nil
 }
 
-func joinAsPrimaryNetworkValidator(nodeID ids.NodeID, network models.Network) error {
+func joinAsPrimaryNetworkValidator(nodeID ids.NodeID, network models.Network, nodeIndex int) error {
 	ux.Logger.PrintToUser(fmt.Sprintf("Adding node %s as a Primary Network Validator...", nodeID.String()))
 	var (
 		start time.Time
@@ -177,7 +178,7 @@ func joinAsPrimaryNetworkValidator(nodeID ids.NodeID, network models.Network) er
 	if weight < minValStake {
 		return fmt.Errorf("illegal weight, must be greater than or equal to %d: %d", minValStake, weight)
 	}
-	start, duration, err = getTimeParametersPrimaryNetwork(network)
+	start, duration, err = getTimeParametersPrimaryNetwork(network, nodeIndex)
 	if err != nil {
 		return err
 	}
@@ -220,53 +221,71 @@ func promptWeightPrimaryNetwork(network models.Network) (uint64, error) {
 	}
 }
 
-func getTimeParametersPrimaryNetwork(network models.Network) (time.Time, time.Duration, error) {
+func getTimeParametersPrimaryNetwork(network models.Network, nodeIndex int) (time.Time, time.Duration, error) {
 	const (
 		defaultDurationOption = "Minimum staking duration on primary network"
 		custom                = "Custom"
 	)
+	var err error
 	start := time.Now().Add(constants.PrimaryNetworkValidatingStartLeadTime)
-	if duration == 0 {
-		msg := "How long should your validator validate for?"
-		durationOptions := []string{defaultDurationOption, custom}
-		durationOption, err := app.Prompt.CaptureList(msg, durationOptions)
+	if useCustomDuration && duration != 0 {
+		return start, duration, nil
+	}
+	if duration != 0 {
+		duration, err = getDefaultValidationTime(start, network, nodeIndex)
 		if err != nil {
 			return time.Time{}, 0, err
 		}
-
-		switch durationOption {
-		case defaultDurationOption:
-			duration, err = getDefaultMaxValidationTime(start, network)
-			if err != nil {
-				return time.Time{}, 0, err
-			}
-		default:
-			duration, err = subnetcmd.PromptDuration(start, network)
-			if err != nil {
-				return time.Time{}, 0, err
-			}
+		return start, duration, nil
+	}
+	msg := "How long should your validator validate for?"
+	durationOptions := []string{defaultDurationOption, custom}
+	durationOption, err := app.Prompt.CaptureList(msg, durationOptions)
+	if err != nil {
+		return time.Time{}, 0, err
+	}
+	switch durationOption {
+	case defaultDurationOption:
+		duration, err = getDefaultValidationTime(start, network, nodeIndex)
+		if err != nil {
+			return time.Time{}, 0, err
+		}
+	default:
+		useCustomDuration = true
+		duration, err = subnetcmd.PromptDuration(start, network)
+		if err != nil {
+			return time.Time{}, 0, err
 		}
 	}
 	return start, duration, nil
 }
 
-func getDefaultMaxValidationTime(start time.Time, network models.Network) (time.Duration, error) {
+func getDefaultValidationTime(start time.Time, network models.Network, nodeIndex int) (time.Duration, error) {
 	durationStr := constants.DefaultFujiStakeDuration
 	if network == models.Mainnet {
 		durationStr = constants.DefaultMainnetStakeDuration
 	}
+	durationInt, err := strconv.Atoi(durationStr[:len(durationStr)-1])
+	if err != nil {
+		return 0, err
+	}
+	// stagger expiration time by 1 day for each added node
+	durationAddition := 24 * nodeIndex
+	durationStr = strconv.Itoa(durationInt+durationAddition) + "h"
 	d, err := time.ParseDuration(durationStr)
 	if err != nil {
 		return 0, err
 	}
 	end := start.Add(d)
-	confirm := fmt.Sprintf("Your validator will finish staking by %s", end.Format(constants.TimeParseLayout))
-	yes, err := app.Prompt.CaptureYesNo(confirm)
-	if err != nil {
-		return 0, err
-	}
-	if !yes {
-		return 0, errors.New("you have to confirm staking duration")
+	if nodeIndex == 0 {
+		confirm := fmt.Sprintf("Your validator will finish staking by %s", end.Format(constants.TimeParseLayout))
+		yes, err := app.Prompt.CaptureYesNo(confirm)
+		if err != nil {
+			return 0, err
+		}
+		if !yes {
+			return 0, errors.New("you have to confirm staking duration")
+		}
 	}
 	return d, nil
 }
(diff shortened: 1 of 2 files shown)


cmd/nodecmd/status.go
```diff
@@ -34,6 +38,10 @@ func statusSubnet(_ *cobra.Command, args []string) error {
 	if err := setupAnsible(); err != nil {
 		return err
 	}
+	hostAliases, err := ansible.GetAnsibleHostsFromInventory(app.GetAnsibleInventoryFilePath(clusterName))
```
felipemadero, line 41: ansibleHostIDs

cmd/nodecmd/status.go
```diff
@@ -46,9 +54,58 @@ func statusSubnet(_ *cobra.Command, args []string) error {
...
+	if err != nil {
 		return err
 	}
-	_, err := checkNodeIsBootstrapped(clusterName, true)
-	return err
+	printOutput(hostAliases, notBootstrappedNodes, clusterName, true)
+	return nil
+}
+
+func printOutput(hostAliases, notBootstrappedHosts []string, clusterName string, primaryNetwork bool) {
```
felipemadero, line 78: let's receive subnetName instead of bool, if subnetName is empty assumes it is primaryNetwork, and we avoid accessing global state also here

cmd/nodecmd/status.go
```diff
@@ -46,9 +54,58 @@ func statusSubnet(_ *cobra.Command, args []string) error {
...
+		} else {
+			ux.Logger.PrintToUser(fmt.Sprintf("All nodes in cluster %s are synced to Subnet %s", clusterName, subnetName))
+		}
+		return
+	}
+	ux.Logger.PrintToUser(fmt.Sprintf("Node(s) Status For Cluster %s", clusterName))
+	ux.Logger.PrintToUser("======================================")
+	for _, host := range hostAliases {
+		hostIsBootstrapped := true
+		for _, notBootstrappedHost := range notBootstrappedHosts {
```
felipemadero, line 91: let's make the function to receive a SET of notBootstrappedHosts and use set contains/belongs

cmd/nodecmd/status.go
```diff
@@ -46,9 +54,58 @@ func statusSubnet(_ *cobra.Command, args []string) error {
...
+			} else {
+				ux.Logger.PrintToUser(fmt.Sprintf("Node %s is synced to Subnet %s", host, subnetName))
+			}
+		} else {
+			if primaryNetwork {
+				ux.Logger.PrintToUser(fmt.Sprintf("Node %s is not bootstrapped to Primary Network", host))
+			} else {
+				ux.Logger.PrintToUser(fmt.Sprintf("Node %s is not synced to Subnet %s", host, subnetName))
+			}
+		}
```
felipemadero, line 109: would not be nicer to have a table? with one nodes per row and node + primary + subnet as columns,
and OK/ERR or something like that on cells

felipemadero: maybe an issue for this, or maybe make it here

felipemadero: maybe colors to put emphasis on the problems

cmd/nodecmd/status.go
```diff
@@ -46,9 +54,58 @@ func statusSubnet(_ *cobra.Command, args []string) error {
...
+		hostIsBootstrapped := true
+		for _, notBootstrappedHost := range notBootstrappedHosts {
+			if notBootstrappedHost == host {
+				hostIsBootstrapped = false
+			}
+			break
+		}
+		if hostIsBootstrapped {
+			if primaryNetwork {
+				ux.Logger.PrintToUser(fmt.Sprintf("Node %s is bootstrapped to Primary Network", host))
```
felipemadero, line 99: this may be compacted based on the difference between sentences: "is" vs "is not"

cmd/nodecmd/stop.go
```diff
@@ -84,38 +81,70 @@ func stopNode(_ *cobra.Command, args []string) error {
...
+	for _, node := range clusterNodes {
+		nodeConfig, err := app.LoadClusterNodeConfig(node)
+		if err != nil {
+			ux.Logger.PrintToUser(fmt.Sprintf("Failed to stop node %s due to %s", node, err.Error()))
+			failedNodes = append(failedNodes, node)
+			nodeErrors = append(nodeErrors, err)
+			continue
+		}
+		if nodeConfig.Region != lastRegion {
+			sess, err := getAWSCloudCredentials(nodeConfig.Region, true)
```
felipemadero, line 104: do we want to ask the user credentials confirmation for each region change?

cmd/nodecmd/stop.go
```diff
@@ -84,38 +81,70 @@ func stopNode(_ *cobra.Command, args []string) error {
...
+		}
+		if !isRunning {
+			noRunningNodeErr := fmt.Errorf("no running node with instance id %s is found in cluster %s", nodeConfig.NodeID, clusterName)
+			ux.Logger.PrintToUser(fmt.Sprintf("Failed to stop node %s due to %s", node, noRunningNodeErr))
+			failedNodes = append(failedNodes, node)
+			nodeErrors = append(nodeErrors, noRunningNodeErr)
+			continue
+		}
+		ux.Logger.PrintToUser(fmt.Sprintf("Stopping node instance %s in cluster %s...", nodeConfig.NodeID, clusterName))
+		if err = awsAPI.StopInstance(ec2Svc, nodeConfig.NodeID, nodeConfig.ElasticIP, true); err != nil {
```
felipemadero, line 126: you can use err := here

cmd/nodecmd/validateSubnet.go
```diff
@@ -75,30 +71,42 @@ func parseSubnetSyncOutput(filePath string, printOutput bool) (string, error) {
...
 	}
-	ux.Logger.PrintToUser("Node successfully added as Subnet validator!")
+	ux.Logger.PrintToUser(fmt.Sprintf("Node %s successfully added as Subnet validator! (%d / %d)", nodeID, currentNodeIndex+1, nodeCount))
+	ux.Logger.PrintToUser("======================================")
 	return nil
 }
 
-func getNodeSubnetSyncStatus(blockchainID, clusterName string, printOutput, errOnValidating bool) (bool, error) {
-	ux.Logger.PrintToUser("Checking if node is synced to subnet ...")
+func getNodeSubnetSyncStatus(blockchainID, clusterName, hostAlias string, statusOutput, errOnValidating bool) (bool, error) {
```
felipemadero, line 84: ansibleNodeID

cmd/nodecmd/validateSubnet.go
```diff
@@ -75,30 +71,42 @@ func parseSubnetSyncOutput(filePath string, printOutput bool) (string, error) {
...
+	if statusOutput {
```
felipemadero, line 85: why this diffference. probably just print one of the options

cmd/nodecmd/validateSubnet.go
felipemadero, line 84: comment what this function checks and it what cases
maybe it would be better for this function to check subnet status against a expected status given as funtion param, where status cmd pass in validating, and this command pass in syncing, and make the external function fail with error. it seems to me that the logic from two different usages is beign put inside here and may be better to have outside, in the callers

cmd/nodecmd/validateSubnet.go
```diff
@@ -127,28 +135,20 @@ func waitForNodeToBePrimaryNetworkValidator(nodeID ids.NodeID) error {
 
 func validateSubnet(_ *cobra.Command, args []string) error {
 	clusterName := args[0]
-	subnetName := args[1]
+	subnetName = args[1]
```
felipemadero, line 138: I rather prefer the option of not using the global variable, but instead passing this as argument, as in another PR.
in any case let's the other PR implementation prevail. I prefer no global as there is no need to use global here (there is no such flag)

cmd/nodecmd/validateSubnet.go
```diff
@@ -161,22 +161,65 @@ func validateSubnet(_ *cobra.Command, args []string) error {
 	if blockchainID == ids.Empty {
 		return ErrNoBlockchainID
 	}
-	// we have to check if node is synced to subnet before adding the node as a validator
-	isSubnetSynced, err := getNodeSubnetSyncStatus(blockchainID.String(), clusterName, false, true)
+	hostAliases, err := ansible.GetAnsibleHostsFromInventory(app.GetAnsibleInventoryDirPath(clusterName))
```
felipemadero, line 164: ansibleNodeIDs

cmd/nodecmd/validateSubnet.go
```diff
@@ -161,22 +154,63 @@ func validateSubnet(_ *cobra.Command, args []string) error {
...
 	}
-	if addedNodeAsPrimaryNetworkValidator {
-		if err := waitForNodeToBePrimaryNetworkValidator(nodeID); err != nil {
-			return err
+	if len(failedNodes) > 0 {
+		ux.Logger.PrintToUser("Failed nodes: ")
+		for i, node := range failedNodes {
+			ux.Logger.PrintToUser(fmt.Sprintf("node %s failed due to %s", node, nodeErrors[i]))
 		}
+		return fmt.Errorf("node(s) %s failed to validate subnet %s", failedNodes, subnetName)
```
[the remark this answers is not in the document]
felipemadero: Nice

felipemadero (commented):
nits

cmd/nodecmd/create.go
```diff
@@ -443,6 +449,20 @@ func requestAWSAccountAuth() error {
 	return nil
 }
 
+func requestStopAWSNodeAuth() error {
+	ux.Logger.PrintToUser("Do you authorize Avalanche-CLI to access your AWS account to stop your Avalanche Validator node?")
```
patrick-ogrady, line 453: is this stopping or deleting?

We should say that this could cause Subnet instability and is not reversible?

cmd/nodecmd/status.go
```diff
@@ -34,6 +38,10 @@ func statusSubnet(_ *cobra.Command, args []string) error {
 	if err := setupAnsible(); err != nil {
 		return err
 	}
+	ansibleHostIDs, err := ansible.GetAnsibleHostsFromInventory(app.GetAnsibleInventoryDirPath(clusterName))
+	if err != nil {
+		return err
+	}
 	if subnetName != "" {
```
patrick-ogrady, line 45: nit: len(subnetName) > 0

cmd/nodecmd/sync.go
```diff
@@ -51,17 +50,33 @@ func syncSubnet(_ *cobra.Command, args []string) error {
...
+		return fmt.Errorf("node(s) %s are not bootstrapped yet, please try again later", notBootstrappedNodes)
 	}
-	if err := checkAvalancheGoVersionCompatible(clusterName, subnetName); err != nil {
+	incompatibleNodes, err := checkAvalancheGoVersionCompatible(clusterName, subnetName)
+	if err != nil {
 		return err
 	}
-	return trackSubnet(clusterName, subnetName, models.Fuji)
+	if len(incompatibleNodes) > 0 {
+		ux.Logger.PrintToUser("Either modify your Avalanche Go version or modify your Subnet-EVM version")
```
patrick-ogrady, line 65: Should this be Custom VM? Or are we guaranteed this can only happen with SubnetEVM?

cmd/nodecmd/validatePrimary.go
```diff
@@ -286,33 +271,40 @@ func getDefaultMaxValidationTime(start time.Time, network models.Network) (time.
...
+		}
 	}
-	return false, nil
+	return notBootstrappedNodes, nil
 }
 
-func getClusterNodeID(clusterName string) (string, error) {
-	ux.Logger.PrintToUser("Getting node id ...")
+func getClusterNodeID(clusterName, hostAlias string) (string, error) {
+	ux.Logger.PrintToUser(fmt.Sprintf("Getting Avalanche node id for node %s...", hostAlias))
```
patrick-ogrady, line 303: nit: for `host` not `node`

patrick-ogrady: node is a loaded term in this context

pkg/ansible/ansible.go
```diff
@@ -56,6 +58,27 @@ func Setup(ansibleDir string) error {
...
+// GetAnsibleHostsFromInventory gets alias of all hosts in an inventory file
+func GetAnsibleHostsFromInventory(inventoryDirPath string) ([]string, error) {
+	inventoryHostsFile := filepath.Join(inventoryDirPath, constants.AnsibleHostInventoryFileName)
+	ansibleHostIDs := []string{}
+	file, err := os.Open(inventoryHostsFile)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+	scanner := bufio.NewScanner(file)
```
patrick-ogrady, line 70: Is there not some official package for reading this?

felipemadero (approved):
LGTM. After addressing remaining comments from patrick. Already approving

PR #1000, opened 2023-09-07
stagger primary validation time

sukantoraymond:
add 24 hour validation time for each added node (add 24 h the higher the index of the node is)

### DIFF
--- a/cmd/nodecmd/validatePrimary.go
+++ b/cmd/nodecmd/validatePrimary.go
@@ -385,6 +385,7 @@ func validatePrimaryNetwork(_ *cobra.Command, args []string) error {
 	}
 	failedNodes := []string{}
 	nodeErrors := []error{}
+	ux.Logger.PrintToUser("Note that we have staggered the end time of validation period to increase by 24 hours for each node added if multiple nodes are added as Primary Network validators simultaneously")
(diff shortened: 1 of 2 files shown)


cmd/nodecmd/validatePrimary.go
```diff
@@ -220,53 +221,71 @@ func promptWeightPrimaryNetwork(network models.Network) (uint64, error) {
...
+	durationInt, err := strconv.Atoi(durationStr[:len(durationStr)-1])
 	if err != nil {
 		return 0, err
 	}
-	end := start.Add(d)
-	confirm := fmt.Sprintf("Your validator will finish staking by %s", end.Format(constants.TimeParseLayout))
-	yes, err := app.Prompt.CaptureYesNo(confirm)
+	// stagger expiration time by 1 day for each added node
+	durationAddition := 24 * nodeIndex
+	durationStr = strconv.Itoa(durationInt+durationAddition) + "h"
```
patrick-ogrady, line 274: You may want to better clarify that the validation period gets longer per node you add in bulk?

patrick-ogrady: I think that is a better option than setting the start to be delayed, but we may want to log that

PR #1014, opened 2023-09-08
Add node update command to update subnet configurations + vm

felipemadero:


cmd/nodecmd/update.go
```diff
@@ -0,0 +1,84 @@
...
+
+The node update command updates all nodes in a cluster with latest Subnet configuration and virtual machine.
+You can check the updated subnet bootstrap status by calling avalanche node status <clusterName> --subnet <subnetName>`,
+		SilenceUsage: true,
+		Args:         cobra.ExactArgs(2),
+		RunE:         updateSubnet,
+	}
+
+	return cmd
+}
```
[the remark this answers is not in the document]
felipemadero: this command can also be used to update subnet evm version, as subnet conf can be changed to point to a specific subnet evm version and then this command can (or should be user)
how do you propose to differentiate? should I call it `node update subnet [clusterName] [subnetName]` ?

cmd/nodecmd/updateSubnet.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+	if err := ansible.RunAnsiblePlaybookExportSubnet(app.GetAnsibleDir(), app.GetAnsibleInventoryPath(clusterName), subnetPath, "/tmp"); err != nil {
+		return err
+	}
+	// runs avalanche join subnet command
+	if err := ansible.RunAnsiblePlaybookUpdateSubnet(app.GetAnsibleDir(), subnetName, subnetPath, app.GetAnsibleInventoryPath(clusterName)); err != nil {
+		return err
+	}
+	ux.Logger.PrintToUser("Node successfully updated Subnet!")
```
patrick-ogrady, line 81: Is there an option to sleep some time between updates? You will likely not want to run through everything at once.

cmd/subnetcmd/join.go
```diff
@@ -277,9 +285,95 @@ func joinCmd(_ *cobra.Command, args []string) error {
...
+	subnetName string,
+	sc models.Sidecar,
+	network models.Network,
+) error {
+	if dataDir == "" {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return err
+		}
+		dataDir = filepath.Join(home, ".avalanchego")
```
patrick-ogrady, line 313: is this run on the host? if not, the `os.UserHomeDir` could be different there.

cmd/subnetcmd/configure.go
```diff
@@ -102,10 +103,31 @@ func configure(_ *cobra.Command, args []string) error {
...
-	fileBytes, err := utils.ValidateJSON(path)
+	sc, err := app.LoadSidecar(subnet)
 	if err != nil {
 		return err
 	}
+	var (
+		fileBytes    []byte
+		validateJSON bool
+	)
+	if filename == constants.SubnetConfigFileName {
```
patrick-ogrady, line 114: why not validateJSON if it just ends in `.json`?

cmd/subnetcmd/join.go
```diff
@@ -277,9 +285,95 @@ func joinCmd(_ *cobra.Command, args []string) error {
 
 	ux.Logger.PrintToUser("VM binary written to %s", vmPath)
 
+	if forceWrite {
```
patrick-ogrady, line 288: When would we not want to write this?

cmd/nodecmd/updateSubnet.go
felipemadero: which is the use case you are considering? this just updating everything defined by the user: it can be vm binary and/or subnet conf and/or chain conf and/or avago conf.
for this proposed implementation, if the user wants a dependence in the updated, he should make a change locally and then apply it in the node, make a second change locally and then apply it to the node and so on

cmd/subnetcmd/join.go
felipemadero: the join command originally presented the user the choice of automatic changing a node conf, or manual chaging it (printing the changes to be made to the user which himself made the change). the original command only change the tracked subnets in an avago conf.
I just kept part of the original spirit of the command and only make an avago  node conf change if the used explicit  asks to do so by using the --force-write flag
to fully continue with that spirit, if not force write, the command should tell the user where to put all the confs, but I opted to do that in a subsequent step

cmd/nodecmd/updateSubnet.go
felipemadero: https://github.com/ava-labs/avalanche-cli/issues/1034

PR #1070, opened 2023-10-04
update ubuntu when build env

sukantoraymond:
Need to update ubuntu before installing GCC otherwise there will be error


pkg/ansible/playbook/setupBuildEnv.yml
```diff
@@ -1,6 +1,9 @@
 ---
 - hosts: "{{ target }}"
   tasks:
+    - name: update Ubuntu
```
arturrez, line 4: name should be `apt update`

arturrez (approved):

arturrez (approved):

PR #1075, opened 2023-10-06
Stop gcp instance

sukantoraymond:


cmd/nodecmd/create_gcp.go
```diff
@@ -226,6 +227,31 @@ func createGCPInstance(usr *user.User, gcpClient *compute.Service, zone, imageID
 	instanceIDs, elasticIPs, certFilePath, keyPairName, err := createGCEInstances(rootBody, gcpClient, hclFile, zone, imageID, defaultAvalancheCLIPrefix, gcpProjectName, gcpCredentialFilepath)
 	if err != nil {
 		ux.Logger.PrintToUser("Failed to create GCP cloud server")
+		if err.Error() == constants.ErrCreatingGCPNode {
```
felipemadero, line 230: It seems that we need to print a clear msg that we started to stop all nodes due to create failure

felipemadero: can we have the same code for this and for stop command? it seems to be doing pretty similar stuff (stop all nodes of the cluster). this to reduce code/logic repetition

cmd/nodecmd/create_aws.go
```diff
@@ -229,30 +230,32 @@ func createAWSInstance(ec2Svc *ec2.EC2, region, ami string, usr *user.User) (Clo
...
-		}
-		failedNodes := []string{}
-		nodeErrors := []error{}
-		for _, instanceID := range instanceIDs {
-			ux.Logger.PrintToUser(fmt.Sprintf("Stopping AWS cloud server %s...", instanceID))
-			if stopErr := awsAPI.StopInstance(ec2Svc, instanceID, "", false); stopErr != nil {
-				failedNodes = append(failedNodes, instanceID)
-				nodeErrors = append(nodeErrors, stopErr)
+		if strings.Contains(err.Error(), constants.ErrCreatingAWSNode) {
+			// we stop created instances so that user doesn't pay for unused EC2 instances
```
felipemadero, line 234: same here. print a msg of what is happening and probably encapsulate this code together with the stop command code

felipemadero (commented):
LGTM. Pretty near

felipemadero (approved):
LGTM

Issue #970, opened 2023-08-25
Upgrade/downgrade avalanchego on node


felipemadero:
Just used it and there is inconsistency between latest subnet-evm and latest avalanchego. We should give the user the option to downgrade/upgrade avalanchego as convenient on node sync

martineckardt:
We’ve got a few reports from academy users about this. This occurs when a new avalancheGo version with a VM interface version change is released, but an updated subnet-EVM has not been released yet. 

I think it would be most effective for Subnet-CLI not to use the latest version of AvalancheGo, but the latest version of AvalancheGo compatible with the selected Subnet-EVM version

PR #1107, opened 2023-10-24
Upgrade avago and subnetevm 

sukantoraymond:
Upgrades avalanche go and subnet EVM if node is tracking a subnet. 

Node will check the latest available subnet EVM version and upgrade the latest avalanche go that is still compatible with that version of subnet EVM. 

If node is not tracking any subnet, it will be upgraded to the latest available avalanche go version

### DIFF
--- a/cmd/nodecmd/upgrade.go
+++ b/cmd/nodecmd/upgrade.go
@@ -19,7 +19,7 @@ import (
 	"golang.org/x/exp/slices"
 )
 
-type NodeUpgradeInfo struct {
+type nodeUpgradeInfo struct {
 	AvalancheGoVersion    string   // avalanche go version to update to on cloud server
 	SubnetEVMVersion      string   // subnet EVM version to update to on cloud server
 	SubnetEVMIDsToUpgrade []string // list of ID of Subnet EVM to be upgraded to subnet EVM version to update to
@@ -89,7 +89,7 @@ func upgrade(_ *cobra.Command, args []string) error {
 // nodes needs to have Avalanche Go & SubnetEVM upgraded. It first checks the subnet EVM version -
 // it will install the newest subnet EVM version and install the latest avalanche Go that is still compatible with the Subnet EVM version
 // if the node is not tracking any subnet, it will just install latestAvagoVersion
-func getNodesUpgradeInfo(clusterName string) (map[string]NodeUpgradeInfo, error) {
+func getNodesUpgradeInfo(clusterName string) (map[string]nodeUpgradeInfo, error) {
 	latestAvagoVersion, err := app.Downloader.GetLatestReleaseVersion(binutils.GetGithubLatestReleaseURL(
 		constants.AvaLabsOrg,
 		constants.AvalancheGoRepoName,
@@ -114,7 +114,7 @@ func getNodesUpgradeInfo(clusterName string) (map[string]NodeUpgradeInfo, error)
 	}
 	failedNodes := []string{}
 	nodeErrors := []error{}
-	nodesToUpgrade := make(map[string]NodeUpgradeInfo)
+	nodesToUpgrade := make(map[string]nodeUpgradeInfo)
 	for _, host := range ansibleNodeIDs {
 		if err := app.CreateAnsibleStatusFile(app.GetAvalancheGoJSONFile()); err != nil {
 			failedNodes = append(failedNodes, host)
@@ -134,7 +134,7 @@ func getNodesUpgradeInfo(clusterName string) (map[string]NodeUpgradeInfo, error)
 		}
 		currentAvalancheGoVersion := vmVersions[constants.PlatformKeyName]
 		avalancheGoVersionToUpdateTo := latestAvagoVersion
-		nodeUpgradeInfo := NodeUpgradeInfo{}
+		nodeUpgradeInfo := nodeUpgradeInfo{}
 		nodeUpgradeInfo.SubnetEVMIDsToUpgrade = []string{}
 		for vmName, vmVersion := range vmVersions {
 			// when calling info.getNodeVersion, this is what we get
--- a/cmd/nodecmd/update.go
+++ b/cmd/nodecmd/update.go
@@ -11,7 +11,7 @@ import (
 func newUpdateCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "update",
-		Short: "(ALPHA Warning) Update avalanchego or VM config  for all node in a cluster",
+		Short: "(ALPHA Warning) Update avalanchego or VM config for all node in a cluster",
 		Long: `(ALPHA Warning) This command is currently in experimental mode.
 
 The node update command suite provides a collection of commands for nodes to update


cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+func upgrade(_ *cobra.Command, args []string) error {
+	clusterName := args[0]
+	if err := setupAnsible(clusterName); err != nil {
+		return err
+	}
+	toUpgradeNodes, err := getNodesToBeUpgraded(clusterName)
+	if err != nil {
+		return err
+	}
+	for node, toUpgrade := range toUpgradeNodes {
```
felipemadero, line 49: please use a some struct instead of a string (with avago version, subnet evm version, vm id)

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+		// toUpgradeNodes either has value of avalancheGo or subnetEvm or avalancheGo,subnetEvm
+		toUpgradeItems := strings.Split(toUpgrade, ",")
+		for _, toUpGradeItem := range toUpgradeItems {
+			// toUpGradeItem either has value format of avalancheGo=<versionNum> or subnetEvm=<versionNum>
+			upgradeInfo := strings.Split(toUpGradeItem, "=")
+			if strings.Contains(toUpGradeItem, constants.AvalancheGoRepoName) {
```
felipemadero, line 55: if should install a specific avago if so needed by the vm (here it is either one or the other)

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+				if err = upgradeAvalancheGo(clusterName, node, upgradeInfo[1]); err != nil {
+					return err
+				}
+			} else if strings.Contains(toUpGradeItem, constants.SubnetEVMRepoName) {
+				// subnetEVM version has value format of n8Anw9kErmgk7KHviddYtecCmziLZTphDwfL1V2DfnFjWZXbE:<versionNum>
+				subnetEVMVersionInfo := strings.Split(upgradeInfo[1], ":")
+				subnetEMVersionToUpgradeTo := subnetEVMVersionInfo[1]
+				subnetEMVersionToUpgradeToWoPrefix := strings.TrimPrefix(subnetEMVersionToUpgradeTo, "v")
+				subnetEVMReleaseURL := fmt.Sprintf("https://github.com/ava-labs/subnet-evm/releases/download/%s/subnet-evm_%s_linux_amd64.tar.gz", subnetEMVersionToUpgradeTo, subnetEMVersionToUpgradeToWoPrefix)
+				subnetEVMArchive := fmt.Sprintf("subnet-evm_%s_linux_amd64.tar.gz", subnetEMVersionToUpgradeToWoPrefix)
```
felipemadero, line 65: first generate archive, then use it to generate url

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+			if strings.Contains(toUpGradeItem, constants.AvalancheGoRepoName) {
+				if err = upgradeAvalancheGo(clusterName, node, upgradeInfo[1]); err != nil {
+					return err
+				}
+			} else if strings.Contains(toUpGradeItem, constants.SubnetEVMRepoName) {
+				// subnetEVM version has value format of n8Anw9kErmgk7KHviddYtecCmziLZTphDwfL1V2DfnFjWZXbE:<versionNum>
+				subnetEVMVersionInfo := strings.Split(upgradeInfo[1], ":")
+				subnetEMVersionToUpgradeTo := subnetEVMVersionInfo[1]
+				subnetEMVersionToUpgradeToWoPrefix := strings.TrimPrefix(subnetEMVersionToUpgradeTo, "v")
+				subnetEVMReleaseURL := fmt.Sprintf("https://github.com/ava-labs/subnet-evm/releases/download/%s/subnet-evm_%s_linux_amd64.tar.gz", subnetEMVersionToUpgradeTo, subnetEMVersionToUpgradeToWoPrefix)
```
felipemadero, line 64: use constant for the format string

cmd/nodecmd/upgrade.go
felipemadero, line 65: use constant for the format string

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+
+	return cmd
+}
+
+func upgrade(_ *cobra.Command, args []string) error {
+	clusterName := args[0]
+	if err := setupAnsible(clusterName); err != nil {
+		return err
+	}
+	toUpgradeNodes, err := getNodesToBeUpgraded(clusterName)
```
felipemadero, line 45: name is misleaving, this is getting a upgrade spec or something like that, not only nodes

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+	}
+	return nil
+}
+
+// getNodesToBeUpgraded gets the node versions of all nodes in cluster clusterName and checks which
+// nodes needs to have Avalanche Go & SubnetEVM upgraded. It first checks the subnet EVM version -
+// it will install the newest subnet EVM version and install the latest avalanche Go that is still compatible with the Subnet EVM version
+// if the node is not tracking any subnet, it will just install latestAvagoVersion
+// sample return result map[aws_node_i-006e0fbf1df7e2198:avalanchego=v1.10.13,subnet-evm=n8Anw9kErmgk7KHuuxRag42nR4xhhBeePH2MkwSxfGNPiaCJs:v0.5.7
```
felipemadero, line 80: just return a map of node ids to some struct containing all the upgrade info

felipemadero: or if you like so, a list of structs where the node id is.also a field of the struct

felipemadero: or if you like so, a list of structs where the node id is.also a field of the struct

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+// aws_node_i-06af8a2ebbeb1bf01:avalanchego=v1.10.13,subnet-evm=n8Anw9kErmgk7KHuuxRag42nR4xhhBeePH2MkwSxfGNPiaCJs:v0.5.7]
+func getNodesToBeUpgraded(clusterName string) (map[string]string, error) {
```
felipemadero, line 82: name is misleaving, it is not only returning node ids

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+	toUpgradeNodes, err := getNodesToBeUpgraded(clusterName)
+	if err != nil {
+		return err
+	}
+	for node, toUpgrade := range toUpgradeNodes {
+		// toUpgradeNodes either has value of avalancheGo or subnetEvm or avalancheGo,subnetEvm
+		toUpgradeItems := strings.Split(toUpgrade, ",")
+		for _, toUpGradeItem := range toUpgradeItems {
+			// toUpGradeItem either has value format of avalancheGo=<versionNum> or subnetEvm=<versionNum>
+			upgradeInfo := strings.Split(toUpGradeItem, "=")
```
felipemadero, line 54: should we made a parallel update for all the cluster? of avago and subnet-evm. can we suppose
all the nodes in the cluster track the same vms?
if not, any way a parallel call can be made for all the nodes that have the same avago requirement, and also for all the nodes that have the same subnet-evm requirement. let's parallelize this

cmd/nodecmd/upgrade.go
felipemadero: new idea: enable parallelization. have a list of nodes associated to each upgrade requirement. also improve the subnet-evm command to upgrade a list of subnet-evm binaries in one command call (probably give to ansible to vm ids not the full vm path)

pkg/ansible/playbook/upgradeSubnetEVM.yml
```diff
@@ -0,0 +1,13 @@
...
+- hosts: "{{ target }}"
+  tasks:
+    - name: download new subnet EVM release
+      shell: wget "{{ subnetEVMReleaseURL }}"
+    - name: unpack new subnet EVM release
+      shell: tar xvf "{{ subnetEVMArchive}}"
+    - name: stop node - stop avalanchego
+      shell: sudo systemctl stop avalanchego
+    - name: replace subnet EVM binary
+      shell: cp subnet-evm "{{ subnetEVMBinaryPath }}"
```
felipemadero, line 11: enable this to update a list of binary paths

felipemadero: maybe only pass the vm id to the command, and encapsulate the full path here

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+	failedNodes := []string{}
+	nodeErrors := []error{}
+	nodesToUpgrade := make(map[string]string)
+	for _, host := range ansibleNodeIDs {
+		if err := app.CreateAnsibleStatusFile(app.GetAvalancheGoJSONFile()); err != nil {
+			failedNodes = append(failedNodes, host)
+			nodeErrors = append(nodeErrors, err)
+			continue
+		}
+		if err := ansible.RunAnsiblePlaybookCheckAvalancheGoVersion(app.GetAnsibleDir(), app.GetAvalancheGoJSONFile(), app.GetAnsibleInventoryDirPath(clusterName), host); err != nil {
```
felipemadero, line 114: do this once in parallel for all hosts

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+			continue
+		}
+		currentAvalancheGoVersion := vmVersions[constants.PlatformKeyName]
+		var avalancheGoVersionToUpdateTo string
+		var subnetEVMVersionToUpdateTo string
+		for vmName, vmVersion := range vmVersions {
+			// when calling info.getNodeVersion, this is what we get
+			// "vmVersions":{"avm":"v1.10.12","evm":"v0.12.5","n8Anw9kErmgk7KHviddYtecCmziLZTphDwfL1V2DfnFjWZXbE":"v0.5.6","platform":"v1.10.12"}},
+			// we need to get the VM ID of the subnets that the node is currently validating, in the example above it is n8Anw9kErmgk7KHviddYtecCmziLZTphDwfL1V2DfnFjWZXbE
+			if vmName != constants.PlatformKeyName && vmName != constants.EVMKeyName && vmName != constants.AVMKeyName {
```
felipemadero, line 132: can we use some set belong of list contain?

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+				if vmVersion != latestSubnetEVMVersion {
```
felipemadero, line 133: you need to check if the vm is custom vm or subnet-evm

felipemadero: you can either search the remote sidecar or the local sidecar (probably the remote sidecar?) but a check is needed.
for custom vms, another thing shoud be made in this same command. (another PR)
what do you think is best? remote sidecar or local sidecar? probably remote. can we have a new CLI command that
provides all the needed info for all the subnets so as to search for the vm? and call that command. maybe we can use one of the current subnet cli commands like list.

felipemadero: I believe it is better to call a remote CLI command and parse the output locally than to get the remote sidecars

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+					failedNodes = append(failedNodes, host)
+					nodeErrors = append(nodeErrors, err)
+					continue
+				}
+			}
+		}
+		if slices.Contains(failedNodes, host) {
+			continue
+		}
+		if currentAvalancheGoVersion != avalancheGoVersionToUpdateTo {
```
felipemadero, line 150: can we hace a target version to latest first, and the update if needed by the evm req, and if the current differs the target, do the update

pkg/ansible/playbook/upgradeAvalancheGo.yml
```diff
@@ -0,0 +1,5 @@
+---
+- hosts: "{{ target }}"
+  tasks:
+    - name: update avalanche go
+      shell: ./avalanchego-installer.sh
```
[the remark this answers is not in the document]
felipemadero: We have an issue for it https://github.com/ava-labs/avalanche-cli/issues/970 
I believe it is needed so

cmd/nodecmd/create.go
```diff
@@ -322,6 +322,13 @@ func getAvalancheGoVersion() (string, error) {
 		return "", err
 	}
 	if chosenOption != "latest" {
+		if chosenOption == "custom" {
```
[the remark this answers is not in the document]
felipemadero: I believe you used constants.CustomVMDir by mistake on the code

cmd/nodecmd/create.go
```diff
@@ -322,6 +322,13 @@ func getAvalancheGoVersion() (string, error) {
...
+			customVersion, err := app.Prompt.CaptureString("Which version of AvalancheGo would you like to install? (Use format v1.10.13)")
```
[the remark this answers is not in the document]
felipemadero: can you please use golang.org/x/mod/semver semver.IsValid?

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,224 @@
...
+		return nil, fmt.Errorf("failed to upgrade node(s) %s", failedNodes)
+	} else {
+		ux.Logger.PrintToUser("All nodes in cluster %s are successfully upgraded!", clusterName)
+	}
+	return nodesToUpgrade, nil
+}
+
+func upgradeAvalancheGo(clusterName, ansibleNodeID, avaGoVersionToUpdateTo string) error {
+	ux.Logger.PrintToUser("Upgrading Avalanche Go version of node %s to version %s ...", ansibleNodeID, avaGoVersionToUpdateTo)
+	if err := ansible.RunAnsiblePlaybookUpgradeAvalancheGo(app.GetAnsibleDir(), app.GetAnsibleInventoryDirPath(clusterName), ansibleNodeID); err != nil {
```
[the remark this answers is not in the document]
felipemadero: you really not are using avaGoVersionToUpdateTo, so either rm the msg and the param, or fix the ansible script

cmd/nodecmd/upgrade.go
felipemadero: could you please put the struct def here instead of models? it seems it is not going to have global usage for the moment

pkg/ansible/playbook/upgradeSubnetEVM.yml
felipemadero: It is not so nice to stop and start the node many times if the node has many vms. Can you use a bash for loop?

cmd/nodecmd/upgrade.go
```diff
@@ -0,0 +1,255 @@
...
+	"github.com/ava-labs/avalanche-cli/pkg/binutils"
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+	"github.com/ava-labs/avalanche-cli/pkg/models"
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+	"github.com/ava-labs/avalanche-cli/pkg/vm"
+	"github.com/spf13/cobra"
+	"golang.org/x/exp/slices"
+)
+
+type NodeUpgradeInfo struct {
```
felipemadero, line 22: does not need to be exported

felipemadero (approved):

PR #1117, opened 2023-10-27
Monitor nodes

sukantoraymond:
Enables monitoring in nodes by adding ports 3000 and 9090 to security groups in AWS and GCP. 
Installs monitoring scripts in nodes through ansible during creation


cmd/nodecmd/create.go
```diff
@@ -274,7 +274,10 @@ func runAnsible(inventoryPath, avalancheGoVersion, clusterName, ansibleHostIDs s
 	if err != nil {
 		return err
 	}
-	return ansible.RunAnsiblePlaybookSetupNode(app.GetConfigPath(), app.GetAnsibleDir(), inventoryPath, avalancheGoVersion, ansibleHostIDs)
+	if err = ansible.RunAnsiblePlaybookSetupNode(app.GetConfigPath(), app.GetAnsibleDir(), inventoryPath, avalancheGoVersion, ansibleHostIDs); err != nil {
+		return err
+	}
+	return ansible.RunAnsiblePlaybookSetupMonitoring(app.GetConfigPath(), app.GetAnsibleDir(), inventoryPath, avalancheGoVersion, ansibleHostIDs)
```
arturrez, line 280: we need to guard it with some flag

PR #1126, opened 2023-11-01
Add lead time for add primary validator cmd

sukantoraymond:
Since users don't have to manually input bls info, we can stick to 20 seconds for node cmd add primary validator, which also prevents users from having to wait longer before proceeding to adding the node as subnet validator.


cmd/nodecmd/validatePrimary.go
```diff
@@ -248,7 +248,10 @@ func GetTimeParametersPrimaryNetwork(network models.Network, nodeIndex int, vali
 			return time.Time{}, 0, err
 		}
 	} else {
-		start = time.Now().Add(constants.PrimaryNetworkValidatingStartLeadTime)
+		start = time.Now().Add(constants.PrimaryNetworkValidatingStartLeadTimeNodeCmd)
```
arturrez, line 251: maybe just pass time.Duration instead of the bool. 
and use constants when you call it.

arturrez (approved):

PR #1130, opened 2023-11-02
add flags so as node create and node stop can be called without user

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/1129

### DIFF
--- a/cmd/nodecmd/create.go
+++ b/cmd/nodecmd/create.go
@@ -346,7 +346,7 @@ func getIPAddress() (string, error) {
 func getAvalancheGoVersion() (string, error) {
 	version := ""
 	subnet := ""
-	if useLatestAvalanchego {
+	if useLatestAvalanchego { //nolint: gocritic
 		version = "latest"
 	} else if useAvalanchegoFromSubnet != "" {
 		subnet = useAvalanchegoFromSubnet
--- a/cmd/nodecmd/create.go
+++ b/cmd/nodecmd/create.go
@@ -27,6 +27,19 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	useAWS                        bool
+	useGCP                        bool
+	cmdLineRegion                 string
+	authorizeAccess               bool
+	numNodes                      int
+	useLatestAvalanchego          bool
+	useAvalanchegoFromSubnet      string
+	cmdLineGCPCredentialsPath     string
+	cmdLineGCPProjectName         string
+	cmdLineAlternativeKeyPairName string
+)
+
 type CloudConfig struct {
 	InstanceIDs   []string
 	PublicIPs     []string
@@ -57,9 +70,19 @@ and users can call node commands with <clusterName> so that the command
 will apply to all nodes in the cluster`,
 		SilenceUsage: true,
 		Args:         cobra.ExactArgs(1),
-		RunE:         createNode,
+		RunE:         createNodes,
 	}
 	cmd.Flags().BoolVar(&useStaticIP, "use-static-ip", true, "attach static Public IP on cloud servers")
+	cmd.Flags().BoolVar(&useAWS, "aws", false, "create node/s in AWS cloud")
+	cmd.Flags().BoolVar(&useGCP, "gcp", false, "create node/s in GCP cloud")
+	cmd.Flags().StringVar(&cmdLineRegion, "region", "", "create node/s in given region")
+	cmd.Flags().BoolVar(&authorizeAccess, "authorize-access", false, "authorize CLI to create cloud resources")
+	cmd.Flags().IntVar(&numNodes, "num-nodes", 0, "number of nodes to create")
+	cmd.Flags().BoolVar(&useLatestAvalanchego, "latest-avalanchego", false, "install latest avalanchego on node/s")
+	cmd.Flags().StringVar(&useAvalanchegoFromSubnet, "avalanchego-from-subnet", "", "install latest avalanchego, compatible with given subnet, on node/s")
+	cmd.Flags().StringVar(&cmdLineGCPCredentialsPath, "gcp-credentials", "", "use given GCP credentials")
+	cmd.Flags().StringVar(&cmdLineGCPProjectName, "gcp-project", "", "use given GCP project")
+	cmd.Flags().StringVar(&cmdLineAlternativeKeyPairName, "alternative-key-pair-name", "", "key pair name to use if default one generates conflicts")
 
 	return cmd
 }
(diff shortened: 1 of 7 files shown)


cmd/nodecmd/create.go
```diff
@@ -27,6 +27,19 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	useAWS                        bool
+	useGCP                        bool
+	cmdLineRegion                 string
+	authorizeAccess               bool
```
arturrez, line 34: I don't think we need this one as user doesn't need to authorize access if uses command line.
what was the point of adding all this params to command line args  and not authorizing access ?

arturrez: there is no other value than true for command to run
we ask user in UI because of politeness

cmd/nodecmd/create.go
```diff
@@ -57,9 +70,19 @@ and users can call node commands with <clusterName> so that the command
...
 		SilenceUsage: true,
 		Args:         cobra.ExactArgs(1),
-		RunE:         createNode,
+		RunE:         createNodes,
 	}
 	cmd.Flags().BoolVar(&useStaticIP, "use-static-ip", true, "attach static Public IP on cloud servers")
+	cmd.Flags().BoolVar(&useAWS, "aws", false, "create node/s in AWS cloud")
+	cmd.Flags().BoolVar(&useGCP, "gcp", false, "create node/s in GCP cloud")
+	cmd.Flags().StringVar(&cmdLineRegion, "region", "", "create node/s in given region")
+	cmd.Flags().BoolVar(&authorizeAccess, "authorize-access", false, "authorize CLI to create cloud resources")
```
arturrez, line 79: remove this ?

cmd/nodecmd/create.go
```diff
@@ -57,9 +70,19 @@ and users can call node commands with <clusterName> so that the command
 will apply to all nodes in the cluster`,
 		SilenceUsage: true,
 		Args:         cobra.ExactArgs(1),
-		RunE:         createNode,
+		RunE:         createNodes,
 	}
 	cmd.Flags().BoolVar(&useStaticIP, "use-static-ip", true, "attach static Public IP on cloud servers")
+	cmd.Flags().BoolVar(&useAWS, "aws", false, "create node/s in AWS cloud")
```
arturrez, line 76: replace with single option --provider. Options `aws` `gcp`. Fail on any other input

cmd/nodecmd/create.go
```diff
@@ -57,9 +70,19 @@ and users can call node commands with <clusterName> so that the command
...
+	cmd.Flags().BoolVar(&useGCP, "gcp", false, "create node/s in GCP cloud")
+	cmd.Flags().StringVar(&cmdLineRegion, "region", "", "create node/s in given region")
+	cmd.Flags().BoolVar(&authorizeAccess, "authorize-access", false, "authorize CLI to create cloud resources")
+	cmd.Flags().IntVar(&numNodes, "num-nodes", 0, "number of nodes to create")
+	cmd.Flags().BoolVar(&useLatestAvalanchego, "latest-avalanchego", false, "install latest avalanchego on node/s")
```
arturrez, line 81: nit: add version to the name maybe

cmd/nodecmd/create.go
```diff
@@ -57,9 +70,19 @@ and users can call node commands with <clusterName> so that the command
...
+	cmd.Flags().StringVar(&useAvalanchegoFromSubnet, "avalanchego-from-subnet", "", "install latest avalanchego, compatible with given subnet, on node/s")
```
arturrez, line 82: nit: add version to the name maybe

cmd/nodecmd/create.go
```diff
@@ -339,34 +382,40 @@ func GetLatestAvagoVersionForRPC(configuredRPCVersion int) (string, error) {
...
 	versionOption, err := app.Prompt.CaptureList(txt, versionOptions)
 	if err != nil {
-		return "", err
+		return "", "", err
 	}
 
 	switch versionOption {
 	case defaultVersion:
-		return "latest", nil
+		return "latest", "", nil
```
arturrez, line 396: "latest" and "subnet" convert to constants

cmd/nodecmd/stop.go
```diff
@@ -33,8 +35,10 @@ The node stop command stops a running node in cloud server
 Note that a stopped node may still incur cloud server storage fees.`,
 		SilenceUsage: true,
 		Args:         cobra.ExactArgs(1),
-		RunE:         stopNode,
+		RunE:         stopNodes,
 	}
+	cmd.Flags().BoolVar(&authorizeAccess, "authorize-access", false, "authorize CLI to release cloud resources")
```
arturrez, line 40: we don't need this when executed from cli

cmd/nodecmd/stop.go
```diff
@@ -33,8 +35,10 @@ The node stop command stops a running node in cloud server
...
+	cmd.Flags().BoolVar(&authorizeRemove, "authorize-remove", false, "authorize CLI to remove all local files related to cloud nodes")
```
arturrez, line 41: same

cmd/nodecmd/create.go
felipemadero: I am not sure about how to fix this. If not adding a flag, CLI is always going to add interaction and ask the user for authorization. We ask the user due because that was previously asked to law people. To cover ourselves. The flag is for advanced users. cc @sukantoraymond

cmd/nodecmd/create.go
felipemadero: following same pattern as in other CLI commands. See deploy with local, fuji, mainnet options. If options are know, fixed, not too much, the flag already checks for proper option content. What do you think on this ui related stuff @dasconnor

cmd/nodecmd/create.go
felipemadero: ok

cmd/nodecmd/create.go
felipemadero: ok

cmd/nodecmd/create.go
felipemadero: yup

cmd/nodecmd/create.go
felipemadero: ok

cmd/nodecmd/stop.go
felipemadero: we need

cmd/nodecmd/stop.go
felipemadero: we need

cmd/nodecmd/create.go
dasconnor: One general design principle we've followed is flags should never be required. This makes it easier for the user to get started and requires less knowledge of how the tool works. 

So in this case, you probably want to put in a prompt to select which cloud provider you want. Then the flag allows you to skip that prompt if set.

dasconnor: Ahh, I think Felipe clarified this for me. I don't think there's a huge difference here. Personally I like `--aws` better, but I don't feel super strongly. Imo, a tie also goes to the PR author 😄

arturrez: ok let's go with `aws` and `gcp` flags

arturrez (approved):

PR #1155, opened 2023-11-08
Node ssh command

felipemadero:
https://asciinema.org/a/619916


cmd/nodecmd/list.go
```diff
@@ -40,34 +39,32 @@ func list(_ *cobra.Command, _ []string) error {
 	if len(clustersConfig.Clusters) == 0 {
 		ux.Logger.PrintToUser("There are no clusters defined.")
 	}
-	for clusterName, clusterConfig := range clustersConfig.Clusters {
-		ux.Logger.PrintToUser(fmt.Sprintf("Cluster %q", clusterName))
+	for clusterName, clusterConf := range clustersConfig.Clusters {
+		ux.Logger.PrintToUser("Cluster %q (%s)", clusterName, clusterConf.Network.Kind.String())
```
arturrez, line 43: use Name()

cmd/nodecmd/list.go
```diff
@@ -40,34 +39,32 @@ func list(_ *cobra.Command, _ []string) error {
...
 			return err
 		}
 		if err := setupAnsible(clusterName); err != nil {
 			return err
 		}
+		ansibleHostIDs, err := ansible.GetAnsibleHostsFromInventory(app.GetAnsibleInventoryDirPath(clusterName))
+		if err != nil {
+			return err
+		}
 		ansibleHosts, err := ansible.GetHostMapfromAnsibleInventory(app.GetAnsibleInventoryDirPath(clusterName))
```
arturrez, line 54: is ansibleHosts used? should be removed

cmd/nodecmd/ssh.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/ava-labs/avalanche-cli/pkg/models"
+	"github.com/ava-labs/avalanche-cli/pkg/utils"
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+	"github.com/spf13/cobra"
+)
+
+func newSSHCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "ssh [clusterName] [cmd]",
+		Short: "(ALPHA Warning) Execute ssh command on node/s",
```
arturrez, line 19: replace /s with (s)

cmd/nodecmd/ssh.go
```diff
@@ -0,0 +1,93 @@
...
+		var err error
+		clustersConfig := models.ClustersConfig{}
+		if app.ClustersConfigExists() {
+			clustersConfig, err = app.LoadClustersConfig()
+			if err != nil {
+				return err
+			}
+		}
+		if len(clustersConfig.Clusters) == 0 {
+			ux.Logger.PrintToUser("There are no clusters defined.")
```
arturrez, line 44: we should return there
no point of continue

cmd/nodecmd/ssh.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+		for clusterName, clusterConfig := range clustersConfig.Clusters {
+			ux.Logger.PrintToUser("Cluster %q (%s)", clusterName, clusterConfig.Network.Kind.String())
```
arturrez, line 47: use Name()

cmd/nodecmd/ssh.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	ansibleHostIDs, err := ansible.GetAnsibleHostsFromInventory(app.GetAnsibleInventoryDirPath(clusterName))
+	if err != nil {
+		return err
+	}
+	ansibleHosts, err := ansible.GetHostMapfromAnsibleInventory(app.GetAnsibleInventoryDirPath(clusterName))
+	if err != nil {
+		return err
+	}
+	for _, host := range ansibleHostIDs {
```
arturrez, line 74: you execute this commands once after another. 
Do you want to use waitGroup or errorGroup to execute ssh commands in parallel ? 
maybe add flags to make these commands executed in parallel so user has a choice

cmd/nodecmd/ssh.go
```diff
@@ -0,0 +1,93 @@
...
+		_, cloudID, err := models.HostAnsibleIDToCloudID(host)
+		if err != nil {
+			return err
+		}
+		cmdLine := utils.GetSSHConnectionString(ansibleHosts[host].IP, ansibleHosts[host].SSHPrivateKeyPath) + " " + strings.Join(args[1:], " ")
```
arturrez, line 79: use fmt.Sprintf instead

cmd/nodecmd/ssh.go
```diff
@@ -0,0 +1,93 @@
...
+		ux.Logger.PrintToUser("%s[%s] %s", indent, cloudID, cmdLine)
+		if len(args) > 1 {
+			splitCmdLine := strings.Split(cmdLine, " ")
+			cmd := exec.Command(splitCmdLine[0], splitCmdLine[1:]...) //nolint: gosec
```
arturrez, line 83: we should set cmd environment with os.environ to be consistent with what user shell env is when he executed ssh commands from his terminal manually

cmd/nodecmd/list.go
felipemadero: it is used to get IPs

felipemadero: node list is showing IPs anyway

arturrez (approved):

PR #1161, opened 2023-11-09
add source code repo prompts to custom vm create

felipemadero:
- makes binary in subnet create
- after that, ready for the rest of flows (eg deploy local, node deploy)


cmd/subnetcmd/create.go
```diff
@@ -60,12 +60,15 @@ configuration, pass the -f flag.`,
 		PersistentPostRun: handlePostRun,
 	}
 	cmd.Flags().StringVar(&genesisFile, "genesis", "", "file path of genesis to use")
-	cmd.Flags().StringVar(&vmFile, "vm", "", "file path of custom vm to use")
 	cmd.Flags().BoolVar(&useSubnetEvm, "evm", false, "use the Subnet-EVM as the base template")
 	cmd.Flags().StringVar(&vmVersion, "vm-version", "", "version of vm template to use")
 	cmd.Flags().BoolVar(&useCustom, "custom", false, "use a custom VM template")
 	cmd.Flags().BoolVar(&useLatestVersion, latest, false, "use latest VM version, takes precedence over --vm-version")
 	cmd.Flags().BoolVarP(&forceCreate, forceFlag, "f", false, "overwrite the existing configuration if one exists")
+	cmd.Flags().StringVar(&vmFile, "vm", "", "file path of custom vm to use")
```
arturrez, line 68: maybe rename `vm` to `custom-vm-path`  or `custom-vm-to-use` to be consistent with the rest of params

pkg/vm/createCustom.go
```diff
@@ -55,3 +78,104 @@ func loadCustomGenesis(app *application.Avalanche, genesisPath string) ([]byte,
...
+		}
+	}
+	if customVMBuildScript != "" {
+		ux.Logger.PrintToUser("Checking build script %s", customVMBuildScript)
+		if err := prompts.ValidateRepoFile(customVMRepoURL, customVMBranch, customVMBuildScript); err != nil {
+			ux.Logger.PrintToUser("Invalid repository build script %s: %s", customVMBuildScript, err)
+			customVMBuildScript = ""
+		}
+	}
+	if customVMBuildScript == "" {
```
arturrez, line 117: put `if customVMBuildScript == ""` first and replase `if customVMBuildScript != ""` with else to make code more clear and consistent with other if statements

pkg/vm/createCustom.go
```diff
@@ -55,3 +78,104 @@ func loadCustomGenesis(app *application.Avalanche, genesisPath string) ([]byte,
...
+		}
+	}
+	sc.CustomVMRepoURL = customVMRepoURL
+	sc.CustomVMBranch = customVMBranch
+	sc.CustomVMBuildScript = customVMBuildScript
+	return nil
+}
+
+func checkGitIsInstalled() error {
+	if err := exec.Command("git").Run(); errors.Is(err, exec.ErrNotFound) {
```
arturrez, line 130: nit: run `git version` not `git`

pkg/vm/createCustom.go
```diff
@@ -55,3 +78,104 @@ func loadCustomGenesis(app *application.Avalanche, genesisPath string) ([]byte,
...
+	_ = os.RemoveAll(vmPath)
+
+	// build
+	cmd = exec.Command(sc.CustomVMBuildScript, vmPath) //nolint:gosec
+	cmd.Dir = repoDir
+	utils.SetupRealtimeCLIOutput(cmd, true, true)
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("error building custom vm binary using script %s on repo %s: %w", sc.CustomVMBuildScript, sc.CustomVMRepoURL, err)
+	}
+	info, err := os.Stat(vmPath)
```
arturrez, line 173: use `utils.FileExists` for this

pkg/vm/createCustom.go
```diff
@@ -55,3 +78,104 @@ func loadCustomGenesis(app *application.Avalanche, genesisPath string) ([]byte,
...
+	if errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("custom VM binary %s not found. Expected build script to create it as specified on the first script argument", vmPath)
+	}
+	if info.Mode()&0x0100 == 0 {
```
arturrez, line 177: move this code into utils like `isExecutable` so we can reuse this later

pkg/vm/createCustom.go
felipemadero: this is this way on purpose. in the param given in flags is invalid, is it set to "" and then, it is asked to the user in the following if

felipemadero: so, the user is asked with prompt either if:
- no flag was provided (eg customVMBuildScript == "")
- a flag was provided but was invalid, so validate repo file fails, and after that (customVMBuildScript == "")

pkg/vm/createCustom.go
felipemadero: no needed, it is searching for ErrNotFound, it is failing is binary is not found

felipemadero: why do you prefer git version ?

arturrez (approved):

PR #1191, opened 2023-11-13
Monitor Nodes

sukantoraymond:
Enables user to set a separate host for monitoring. 
Installs a custom monitoring installer script in the monitoring instance where it will scrape metrics from the other nodes in the cluster. 
Will be using custom dashboard json to display unified dashboard. 
Ansible host inventory for the separate host is stored in ansible/inventory/<clustername>/monitoring/hosts in .avalanche-cli

### DIFF
--- a/cmd/nodecmd/create.go
+++ b/cmd/nodecmd/create.go
@@ -339,8 +339,8 @@ func createNodes(_ *cobra.Command, args []string) error {
 		avalancheGoPorts := []string{}
 		machinePorts := []string{}
 		for _, publicIP := range cloudConfig.PublicIPs {
-			avalancheGoPorts = append(avalancheGoPorts, fmt.Sprintf("\\'%s:%s\\'", publicIP, strconv.Itoa(constants.AvalanchegoAPIPort)))
-			machinePorts = append(machinePorts, fmt.Sprintf("\\'%s:%s\\'", publicIP, strconv.Itoa(constants.AvalanchegoMachineMetricsPort)))
+			avalancheGoPorts = append(avalancheGoPorts, fmt.Sprintf("'%s:%s'", publicIP, strconv.Itoa(constants.AvalanchegoAPIPort)))
+			machinePorts = append(machinePorts, fmt.Sprintf("'%s:%s'", publicIP, strconv.Itoa(constants.AvalanchegoMachineMetricsPort)))
 		}
 		if err = app.SetupMonitoringEnv(); err != nil {
 			return err
@@ -363,7 +363,7 @@ func createNodes(_ *cobra.Command, args []string) error {
 			wg.Add(1)
 			go func(nodeResults *models.NodeResults, host *models.Host) {
 				defer wg.Done()
-				nodeDirPath := app.GetNodeInstanceDirPath(host.GetCloudID())
+				nodeDirPath := app.GetNodeInstanceAvaGoConfigDirPath(host.NodeID)
 				if err := ssh.RunSSHDownloadNodeConfig(host, nodeDirPath); err != nil {
 					nodeResults.AddResult(host.NodeID, nil, err)
 					return
@@ -378,6 +378,9 @@ func createNodes(_ *cobra.Command, args []string) error {
 				}
 				if err := ssh.RunSSHRestartNode(host); err != nil {
 					nodeResults.AddResult(host.NodeID, nil, err)
+					return
+				}
+				if err := os.RemoveAll(nodeDirPath); err != nil {
 					return
 				}
 			}(&wgResults, host)
--- a/pkg/ssh/ssh.go
+++ b/pkg/ssh/ssh.go
@@ -374,7 +374,6 @@ func RunSSHSetupMachineMetrics(host *models.Host) error {
 }
 
 func RunSSHSetupSeparateMonitoring(host *models.Host, monitoringDashboardPath, avalancheGoPorts, machinePorts string) error {
-	fmt.Printf("running RunSSHSetupSeparateMonitoring on  %s, %s \n", avalancheGoPorts, machinePorts)
 	if err := host.Upload(
 		monitoringDashboardPath,
 		fmt.Sprintf("/home/ubuntu/%s", filepath.Base(monitoringDashboardPath)),
--- a/pkg/application/app.go
+++ b/pkg/application/app.go
@@ -131,6 +131,10 @@ func (app *Avalanche) GetNodeInstanceDirPath(nodeName string) string {
 	return filepath.Join(app.GetNodesDir(), nodeName)
 }
 
+func (app *Avalanche) GetNodeInstanceAvaGoConfigDirPath(nodeName string) string {
+	return filepath.Join(app.GetAnsibleDir(), nodeName)
+}
+
 func (app *Avalanche) GetAnsibleDir() string {
 	return filepath.Join(app.GetNodesDir(), constants.AnsibleDir)
 }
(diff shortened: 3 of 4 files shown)


cmd/nodecmd/create.go
```diff
@@ -88,6 +88,7 @@ will apply to all nodes in the cluster`,
 		RunE:         createNodes,
 	}
 	cmd.Flags().BoolVar(&useStaticIP, "use-static-ip", true, "attach static Public IP on cloud servers")
+	cmd.Flags().BoolVar(&separateMonitoringInstance, "separate-monitoring-instance", false, "host monitoring for all cloud servers on a separate instance")
```
arturrez, line 91: we need another flag to enable monitoring overall. As by default we install bunch of stuff like prometheus, grafana etc to the instances and "steal" resources avaliable for actual avalanchego

cmd/nodecmd/create.go
```diff
@@ -152,6 +153,12 @@ func createNodes(_ *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
+		if !separateMonitoringInstance {
+			separateMonitoringInstance, err = app.Prompt.CaptureYesNo("Do you want to set up a separate instance to host monitoring? (This enables you to monitor all your set up instances in one dashboard)")
```
arturrez, line 157: we should ask if we want to enable monitoring overall

cmd/nodecmd/create.go
```diff
@@ -181,6 +188,11 @@ func createNodes(_ *cobra.Command, args []string) error {
 				publicIPMap[node] = cloudConfig.PublicIPs[i]
 			}
 		}
+		if separateMonitoringInstance {
+			if err = awsAPI.AddSecurityGroupRule(ec2Svc, cloudConfig.PublicIPs[len(cloudConfig.PublicIPs)-1], cloudConfig.SecurityGroup); err != nil {
```
arturrez, line 192: `[len(cloudConfig.PublicIPs)-1` we need a better way to get IP of the monitoring instance. I suggest using separate TF code and separate output.

cmd/nodecmd/create.go
```diff
@@ -208,6 +220,16 @@ func createNodes(_ *cobra.Command, args []string) error {
 	if err = createClusterNodeConfig(network, cloudConfig, clusterName, cloudService); err != nil {
 		return err
 	}
+	monitoringInstanceNodeID := ""
+	monitoringInstancePublicIP := ""
+	if separateMonitoringInstance {
+		// take the last instance ID and public IP and assign them as those of the monitoring instance
+		monitoringInstanceNodeID = cloudConfig.InstanceIDs[len(cloudConfig.InstanceIDs)-1]
```
arturrez, line 227: the same issue. we should keep monitoring as separate entity in cloudConfig. this is very confusing

pkg/ansible/ansible.go
```diff
@@ -18,7 +18,7 @@ import (
 
 // CreateAnsibleHostInventory creates inventory file for ansible
 // specifies the ip address of the cloud server and the corresponding ssh cert path for the cloud server
-func CreateAnsibleHostInventory(inventoryDirPath, certFilePath, cloudService string, publicIPMap map[string]string) error {
+func CreateAnsibleHostInventory(inventoryDirPath, certFilePath, cloudService, monitoringInstanceNodeID string, publicIPMap map[string]string, forMonitoring bool) error {
```
arturrez, line 21: we should have separate func for monitoring inventory IMHO

pkg/ssh/shell/setupMachineMetrics.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+#name:TASK [download monitoring script]
+wget -nd -m https://raw.githubusercontent.com/ava-labs/avalanche-monitoring/main/grafana/monitoring-installer.sh
+#name:TASK [modify permission for monitoring script]
+chmod 755 monitoring-installer.sh
+#name:TASK [set up Prometheus]
+./monitoring-installer.sh --1
```
arturrez, line 7: we only need node_exporter in case of dedicated monitoring host, no need to install full prometheus there

cmd/nodecmd/create.go
arturrez: as we would need difference cpu/mem/disk for this instance

pkg/ssh/shell/setupMachineMetrics.sh
arturrez: another thing that we can have one ssh script here and use golang templates if-else case to control what steps go where

cmd/nodecmd/create.go
```diff
@@ -268,17 +301,92 @@ func createNodes(_ *cobra.Command, args []string) error {
...
+						return
+					}
+				}
 			}
-			if err := ssh.RunSSHSetupCLIFromSource(host, constants.SetupCLIFromSourceBranch); err != nil {
+			if err := ssh.RunSSHSetupBuildEnv(host); err != nil {
 				nodeResults.AddResult(host.NodeID, nil, err)
 				return
 			}
+			// DO WE NEED THIS? We already installed CLI in setUpNode
```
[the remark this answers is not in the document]
arturrez: we need it as per https://github.com/ava-labs/avalanche-cli/blob/main/pkg/ssh/ssh.go#L269

cmd/nodecmd/create.go
```diff
@@ -268,17 +301,92 @@ func createNodes(_ *cobra.Command, args []string) error {
...
+			} else {
+				if setUpMonitoring {
+					if err := ssh.RunSSHSetupMonitoring(host); err != nil {
+						nodeResults.AddResult(host.NodeID, nil, err)
+						return
+					}
+				}
 			}
-			if err := ssh.RunSSHSetupCLIFromSource(host, constants.SetupCLIFromSourceBranch); err != nil {
+			if err := ssh.RunSSHSetupBuildEnv(host); err != nil {
```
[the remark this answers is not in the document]
arturrez: i don't think order is important in this case

pkg/ssh/shell/setupMachineMetrics.sh
arturrez: if we install prometheus, do we start it on each of the monitored nodes ? 
still not clear for me, why prometheus is needed. Node exporter should be able to run standalone

pkg/ssh/ssh.go
```diff
@@ -313,3 +318,90 @@ func RunSSHSubnetSyncStatus(host *models.Host, blockchainID string) ([]byte, err
...
+		constants.SSHDirOpsTimeout,
+	); err != nil {
+		return err
+	}
+	return host.Upload(
+		filepath.Join(nodeInstanceDirPath, constants.NodeFileName),
+		filepath.Join(constants.CloudNodeConfigPath, constants.NodeFileName),
+		constants.SSHFileOpsTimeout,
+	)
+}
```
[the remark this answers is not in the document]
arturrez: done

pkg/ssh/ssh.go
```diff
@@ -313,3 +318,90 @@ func RunSSHSubnetSyncStatus(host *models.Host, blockchainID string) ([]byte, err
...
+	return RunOverSSH(
+		"Setup Machine Metrics",
+		host,
+		constants.SSHScriptTimeout,
+		"shell/setupMachineMetrics.sh",
+		scriptInputs{},
+	)
+}
+
+func RunSSHSetupSeparateMonitoring(host *models.Host, monitoringDashboardPath, avalancheGoPorts, machinePorts string) error {
```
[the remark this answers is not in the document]
arturrez: fixed

PR #1208, opened 2023-11-15
Don't ask about metrics for completion command

### LINKED ISSUE #1206: Problem with installing shell completion on first install
After installing `avalanche-cli` for the first time on a new machine (aka `~/.avalanche-cli` folder doesn't exist), it will always fail. The reason for that is that it tries to install the completion by running the `completion` command of the Avalanche CLI itself. The problem is, since this is a first install, the command actually asks about metrics opt-in before actually outputting the completion script.

https://github.com/ava-labs/avalanche-cli/blob/2428aeb8ec14bef97e38edab5bd2c5b6d0f488a0/scripts/install.sh#L398C43-L398C43

As a result. After a first install, if you use `bash` for example, then you should have the following file:

```
~/.local/share/bash-completion/completions/avalanche.sh
```

with the following content:

```
Avalanche-CLI (the "software") may collect statistical data on how the software is used on an anonymous basis for purposes of product improvement.  This data will not (i) include any passwords, scripts, or data files, (ii) be associated with any particular user or entity, or (iii) include any personally identifiable information or be used to identify individuals or entities using the software.  You can disable such data collection with `avalanche config metrics disable` command, which will result in no data being collected; by using the software without so disabling such data collection you expressly consent to the collection of such data.  You can also read our privacy statement <https://www.avalabs.org/privacy-policy> to learn more.

You can disable data collection with `avalanche config metrics disable` command. You can also read our privacy statement <https://www.avalabs.org/privacy-policy> to learn more.

```

As a workaround, after installation, here's what I did

```
$ avalanche completion bash
```

Then reply to the question, (Yes / No)

Then 

```
$ avalanche completion bash > ~/.local/share/bash-completion/completions/avalanche.sh
$ source .bashrc
```

And now completion works

Mouradif:
Fix #1206

The undocumented `completion` command is used in the installation script in order to output shell completion for the current shell. At that stage, in general, we're in a fresh installation and the user hasn't decided yet about metrics opt-in. In the current version, the command will ask about metrics opt-in making the step `completions()` of the installation process always fail.

This PR fixes this by adding a check using the existing `CheckCommandIsNotCompletion` function (made public) before deciding that it's time to ask the user about metrics opt-in

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -122,7 +122,7 @@ func createApp(cmd *cobra.Command, _ []string) error {
 	if err := migrations.RunMigrations(app); err != nil {
 		return err
 	}
-	if os.Getenv("RUN_E2E") == "" && !app.Conf.ConfigFileExists() && !utils.FileExists(utils.UserHomePath(constants.OldMetricsConfigFileName)) {
+	if os.Getenv("RUN_E2E") == "" && !app.Conf.ConfigFileExists() && !utils.FileExists(utils.UserHomePath(constants.OldMetricsConfigFileName)) && metrics.CheckCommandIsNotCompletion(cmd) {
 		err = metrics.HandleUserMetricsPreference(app)
 		if err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


Mouradif:
@sukantoraymond My commit wasn't GPG signed, I've just amended that commit to sign it, I believe the PR needs to be approved again

PR #1224, opened 2023-11-22
add boolean to consts to control if compiling cli from source

felipemadero:


pkg/constants/constants.go
```diff
@@ -134,6 +134,7 @@ const (
 	OutboundPort                                 = 0
 	Terraform                                    = "terraform"
 	SetupCLIFromSourceBranch                     = "main"
+	EnableSetupCLIFromSource                     = false
```
arturrez, line 137: we need undocumented param for cli to pass

arturrez (approved):

Issue #1246, opened 2023-11-29
check AVAX availability previously executing txs


felipemadero:
with preference in the ledger case

Issue #1185, opened 2023-11-10
use more than one ledger index to pay fees + auth subnet ops


felipemadero:

felipemadero:
Closed by https://github.com/ava-labs/avalanche-cli/pull/1248

Issue #1184, opened 2023-11-10
automatic discovery of the ledger index associated to subnet auth


felipemadero:

felipemadero:
Clossed by https://github.com/ava-labs/avalanche-cli/pull/1248

Issue #1183, opened 2023-11-10
automatic discovery of the ledger index to use to pay tx fees


felipemadero:

felipemadero:
Closed by https://github.com/ava-labs/avalanche-cli/pull/1248

PR #1248, opened 2023-11-30
Search funded ledger index

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/1246 https://github.com/ava-labs/avalanche-cli/issues/1185 https://github.com/ava-labs/avalanche-cli/issues/1184 https://github.com/ava-labs/avalanche-cli/issues/1183 https://github.com/ava-labs/avalanche-cli/issues/1097


cmd/primarycmd/addValidator.go
```diff
@@ -192,10 +187,17 @@ func addValidator(_ *cobra.Command, _ []string) error {
...
 	}
 
-	kc, err := subnetcmd.GetKeychain(false, useLedger, ledgerAddresses, keyName, network)
+	fee := network.GenesisParams().AddPrimaryNetworkValidatorFee
+	kc, err := keychain.GetKeychain(app, false, useLedger, ledgerAddresses, keyName, network, fee)
 	if err != nil {
 		return err
 	}
+
+	// used in E2E to simulate public network execution paths on a local network
```
arturrez, line 196: nit: move this code to utils as we use it in many places

cmd/subnetcmd/addValidator.go
```diff
@@ -85,23 +85,30 @@ func addValidator(_ *cobra.Command, args []string) error {
...
 		useEwoq,
-		&useLedger,
+		useLedger,
 		ledgerAddresses,
+		fee,
 	)
 	if err != nil {
 		return err
 	}
+	// used in E2E to simulate public network execution paths on a local network
```
arturrez, line 102: same nit

cmd/subnetcmd/deploy.go
```diff
@@ -432,22 +411,46 @@ func deploySubnet(cmd *cobra.Command, args []string) error {
...
 		}
-		// use creation key as control key
+		// use first fee-paying key as control key
 		if sameControlKey {
-			controlKeys, err = loadCreationKeys(network, kc)
+			kcKeys, err := kc.PChainFormattedStrAddresses()
 			if err != nil {
 				return err
 			}
+			controlKeys = kcKeys[:1]
```
arturrez, line 448: we need to check if slice is not empty for whatever reason to avoid `slice bounds out of range [:1] with capacity 0`

cmd/subnetcmd/deploy.go
```diff
@@ -587,12 +594,14 @@ func getControlKeys(network models.Network, useLedger bool, kc keychain.Keychain
 	)
 
 	switch listDecision {
-	case creation:
-		keys, err = loadCreationKeys(network, kc)
+	case feePaying:
+		var kcKeys []string
+		kcKeys, err = kc.PChainFormattedStrAddresses()
+		keys = kcKeys[:1]
```
arturrez, line 600: the same check if slice is not empty

pkg/keychain/keychain.go
```diff
@@ -0,0 +1,321 @@
...
+	return &Keychain{
+		Network:       network,
+		Keychain:      keychain,
+		Ledger:        ledger,
+		UsesLedger:    usesLedger,
+		LedgerIndices: ledgerIndices,
+	}
+}
+
+func (kc *Keychain) OnlyOneKey() bool {
```
arturrez, line 60: nit: better name maybe `HasOnlyOneKey`

pkg/keychain/keychain.go
```diff
@@ -0,0 +1,321 @@
...
+		if err != nil {
+			return nil, err
+		}
+		addrsStr = append(addrsStr, addrStr)
+	}
+
+	return addrsStr, nil
+}
+
+func sortUint32(arr []uint32) {
```
arturrez, line 86: move this to utils

pkg/keychain/keychain.go
```diff
@@ -0,0 +1,321 @@
...
+	// for all ledger indices to search for, find if the ledger address belongs to the input
+	// addresses and, if so, add the index pair to indexMap, breaking the loop if
+	// all addresses were found
+	for ledgerIndex := uint32(0); ledgerIndex < numLedgerIndicesToSearch; ledgerIndex++ {
+		ledgerAddress, err := ledgerDevice.Addresses([]uint32{ledgerIndex})
+		if err != nil {
+			return []uint32{}, err
+		}
+		for addressesIndex, addr := range addresses {
+			if addr == ledgerAddress[0] {
```
arturrez, line 248: `if len(ledgerAddress) && addr == ledgerAddress[0]` to make sure ledgerAddress is not empty to avoid crash

pkg/subnet/public.go
```diff
@@ -931,3 +933,10 @@ func IssuePFromXImportTx(
 	}
 	return tx.ID(), err
 }
+
+func multipleTimesMsg(onlyOneKey bool) string {
```
arturrez, line 937: maybe move it to utils

felipemadero:
> One comment
> 
> ```
> Deploying [ledgerSubnet] to Fuji
> ✔ Use ledger
> Looking for ledger indices to pay for 0.200000000 AVAX...
>   Found index 6 with 0.395298250 AVAX
> Ledger addresses: 
>   P-fuji1c8qchzh27navtly08kz5deenm5jvdqayyasuhh (index 0)
>   P-fuji12knpt8x32mhr6zmy23l6rwccgep5l6rqehgddd (index 6)
> Configure which addresses may make changes to the subnet.
> These addresses are known as your control keys. You will also
> set how many control keys are required to make a subnet change (the threshold).
> ✔ Use ledger address
> ```
> 
> So this is my balance
> 
> ```
> +--------+---------+-------------------------+-----------------------------------------------+-------------+---------+
> |  KIND  |  NAME   |          CHAIN          |                    ADDRESS                    |   BALANCE   | NETWORK |
> +--------+---------+-------------------------+-----------------------------------------------+-------------+---------+
> | ledger | index 0 | P-Chain (Bech32 format) | P-fuji1c8qchzh27navtly08kz5deenm5jvdqayyasuhh |           0 | Fuji    |
> +        +---------+                         +-----------------------------------------------+-------------+---------+
> |        | index 1 |                         | P-fuji1kxxyj3ps8mjvjufjdrvl6fwef736ccl00fku2j |           0 | Fuji    |
> +        +---------+                         +-----------------------------------------------+-------------+---------+
> |        | index 2 |                         | P-fuji1qfdag95urh09l64qeacjxrh7x6juykaw8tpshp |           0 | Fuji    |
> +        +---------+                         +-----------------------------------------------+-------------+---------+
> |        | index 3 |                         | P-fuji1ajcfzk2vnjchlpdchhzm7w4mv0er8eyls2dx4l |           0 | Fuji    |
> +        +---------+                         +-----------------------------------------------+-------------+---------+
> |        | index 4 |                         | P-fuji1wht6s3pvmjt0knk6qmv236nrpukgwkzchd5egx |           0 | Fuji    |
> +        +---------+                         +-----------------------------------------------+-------------+---------+
> |        | index 5 |                         | P-fuji1277h88vzntzpmqkhdfmfm3mdfrdgzmjdpzvne0 |           0 | Fuji    |
> +        +---------+                         +-----------------------------------------------+-------------+---------+
> |        | index 6 |                         | P-fuji12knpt8x32mhr6zmy23l6rwccgep5l6rqehgddd | 0.395298250 | Fuji    |
> +        +---------+                         +-----------------------------------------------+-------------+---------+
> |        | index 7 |                         | P-fuji1jjrlhfrx6e975y6e3fass72a09xp73zxd9gqf0 | 2.799000000 | Fuji    |
> +        +---------+                         +-----------------------------------------------+-------------+---------+
> |        | index 8 |                         | P-fuji1dgy2n7leds9g9mvthudy6qp2mt8za92aqy49cy | 0.500000000 | Fuji    |
> +        +---------+                         +-----------------------------------------------+-------------+---------+
> |        | index 9 |                         | P-fuji1tyr5clpry5w42tswknrsqdpwh8u7w2v7f72gc9 |           0 | Fuji    |
> ```
> 
> Wondering why we only show index 0 and index 6 as an option for control keys. Thinking instead we should show the indexes with the funds (in this case index 6,7,8).
> 
> Another comment is maybe we can show more than 1 index with the funds. So inthis case instead of saying found index 6, we can say found these addresses wiht funds (index 6,7,8), which one do uyou want to use to pay for funds

Index 0 is always added as it was chosen as the change index. I mean, It is used in the wallet to consolidate change. I found that makes the behaviour more predictable in general, when different indices are used for fee payment, subnet auth. 
Regarding only index 6, the search stops once it found enough funds. If not, it will continue asking P-Chain for balances until the current max index 100, wasting user time and resources.
I believe it is ok to just pay with the obtained indices, this is really the situation the users are facing issues because they don't want to manage things at the index level.

cmd/primarycmd/addValidator.go
felipemadero: will do it in separate PR to better manage and encapsulate this is the proper way. this needs more that a function, this needs to be called in a specific order and already want to encapsulate all in a bigger function

cmd/subnetcmd/deploy.go
felipemadero: keychain will error is having 0 addresses

cmd/subnetcmd/deploy.go
felipemadero: same response

Issue #1317, opened 2023-12-15
node stop should remove cluster if all nodes are already stopped


felipemadero:
```
go run main.go node stop cluster1
Please note that if your node(s) are validating a Subnet, stopping them could cause Subnet instability and it is irreversible
✔ Yes
Failed nodes: 
Failed to stop node i-02182f6703337a96d due to no running node with instance id i-02182f6703337a96d is found in cluster cluster1
Failed to stop node i-03c47b2bc597422a2 due to no running node with instance id i-03c47b2bc597422a2 is found in cluster cluster1
Failed to stop node i-0bbbd2bb5e053086e due to no running node with instance id i-0bbbd2bb5e053086e is found in cluster cluster1
Failed to stop node i-02f7cdae67d6fbff1 due to no running node with instance id i-02f7cdae67d6fbff1 is found in cluster cluster1
Failed to stop node i-06ff813b332c684d3 due to no running node with instance id i-06ff813b332c684d3 is found in cluster cluster1
Failed to stop node i-07b7282973f652e6d due to no running node with instance id i-07b7282973f652e6d is found in cluster cluster1
Failed to stop node i-0ab27d8a872b98b68 due to no running node with instance id i-0ab27d8a872b98b68 is found in cluster cluster1
Failed to stop node i-0123dc50eddc77cfa due to no running node with instance id i-0123dc50eddc77cfa is found in cluster cluster1
Failed to stop node i-0aa1003bf29b75641 due to no running node with instance id i-0aa1003bf29b75641 is found in cluster cluster1
Failed to stop node i-0be5f52c713cd36d7 due to no running node with instance id i-0be5f52c713cd36d7 is found in cluster cluster1
Error: failed to stop node(s) [i-0ab27d8a872b98b68 i-0123dc50eddc77cfa i-0aa1003bf29b75641 i-0be5f52c713cd36d7 i-07b7282973f652e6d i-03c47b2bc597422a2 i-0bbbd2bb5e053086e i-02f7cdae67d6fbff1 i-06ff813b332c684d3 i-02182f6703337a96d]
exit status 1
```

PR #1319, opened 2023-12-15
add flag to remove cluster info if all nodes are found to be stopped

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/1317

### DIFF
--- a/cmd/nodecmd/stop.go
+++ b/cmd/nodecmd/stop.go
@@ -22,10 +22,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var (
-	authorizeRemove      bool
-	ignoreAlreadyStopped bool
-)
+var authorizeRemove bool
 
 func newStopCmd() *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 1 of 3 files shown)


cmd/nodecmd/stop.go
```diff
@@ -39,6 +42,7 @@ Note that a stopped node may still incur cloud server storage fees.`,
 	}
 	cmd.Flags().BoolVar(&authorizeAccess, "authorize-access", false, "authorize CLI to release cloud resources")
 	cmd.Flags().BoolVar(&authorizeRemove, "authorize-remove", false, "authorize CLI to remove all local files related to cloud nodes")
+	cmd.Flags().BoolVar(&ignoreAlreadyStopped, "ignore-already-stopped-nodes", false, "do not consider already stopped nodes as an error, and remove cluster info anyway")
```
arturrez, line 45: this should be default behavior. If node is stopped and user want all nodes in the cluster to be stopped cli should just silently carry on

arturrez (approved):

PR #1330, opened 2023-12-19
use snapshop compatible with cchain warp

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/1339
Adds new snapshots for pre and post cortina 17, as there is db incompatibility among those for already created local network dbs.


pkg/subnet/local.go
```diff
@@ -697,11 +724,33 @@ func getExpectedDefaultSnapshotSHA256Sum(isSingleNode bool) (string, error) {
 
 // Initialize default snapshot with bootstrap snapshot archive
 // If force flag is set to true, overwrite the default snapshot if it exists
-func SetDefaultSnapshot(snapshotsDir string, forceReset bool, isSingleNode bool) (bool, error) {
-	bootstrapSnapshotArchivePath := filepath.Join(snapshotsDir, constants.BootstrapSnapshotArchiveName)
-	if isSingleNode {
-		bootstrapSnapshotArchivePath = filepath.Join(snapshotsDir, constants.BootstrapSnapshotSingleNodeArchiveName)
+func SetDefaultSnapshot(snapshotsDir string, forceReset bool, avagoVersion string, isSingleNode bool) (bool, error) {
+	resetCurrentSnapshot := forceReset
```
arturrez, line 728: nit: just convert  `SetDefaultSnapshot(snapshotsDir string, resetCurrentSnapshot bool, avagoVersion string, isSingleNode bool) (bool, error) {` to `SetDefaultSnapshot(snapshotsDir string, forceReset bool, avagoVersion string, isSingleNode bool) (bool, error) {`

pkg/subnet/local.go
```diff
@@ -697,11 +724,33 @@ func getExpectedDefaultSnapshotSHA256Sum(isSingleNode bool) (string, error) {
...
+	isPreCortina17 := false
```
arturrez, line 729: nit: `bool` is `false` by default, so it can be just `var isPreCortina17 bool` but I am not sure which way is better

arturrez (approved):

pkg/subnet/local.go
felipemadero: I prefer the current one (I prefer explicit initialization for readability), but using the default init a convention so I wil change

Issue #1283, opened 2023-12-07
make users input GCP region not zone


arturrez:
AWS makes us to specify region and put instance into some random zone in the region
Google makes us to specify region and zone explicitly.

To make things consistent we should make user specify zone for GCP and make `cli` to direct instance(s) into random zone(s) in this region

![Image](https://github.com/ava-labs/avalanche-cli/assets/56270896/7d02c3d2-e682-4941-ae54-06de34246ab5)

Issue #1262, opened 2023-12-04
currently for GCP and AWS we don't verify user input region in case of custom region selection


arturrez:
We need to fix SUBJ and also going forward do the same for all supported clouds

Issue #1138, opened 2023-11-03
add check for valid cloud region, err msg is not so much clear


felipemadero:
Error: RequestError: send request failed
caused by: Post "https://ec2.us-ea.amazonaws.com/": dial tcp: lookup ec2.us-ea.amazonaws.com on 127.0.0.53:53: no such host
exit status 1

PR #1334, opened 2023-12-19
Cloud regions

arturrez:
resolves https://github.com/ava-labs/avalanche-cli/issues/1262 
resolves https://github.com/ava-labs/avalanche-cli/issues/1138
resolves https://github.com/ava-labs/avalanche-cli/issues/1283


cmd/nodecmd/create_aws.go
```diff
@@ -97,8 +98,15 @@ func getAWSCloudConfig(awsProfile string) (map[string]*awsAPI.AwsCloud, map[stri
 	ec2SvcMap := map[string]*awsAPI.AwsCloud{}
 	amiMap := map[string]string{}
 	numNodesMap := map[string]int{}
+	// verify regions are valid
+	if validRegions, err := checkRegions(maps.Keys(finalRegions)); err != nil {
+		return nil, nil, nil, err
+	} else if !validRegions {
+		return nil, nil, nil, fmt.Errorf("invalid regions provided for %s", constants.AWSCloudService)
```
felipemadero, line 105: can you point out to the user the given invalid regions?

cmd/nodecmd/create_aws.go
```diff
@@ -341,3 +349,21 @@ func addCertToSSH(certName string) error {
 	utils.SetupRealtimeCLIOutput(cmd, true, true)
 	return cmd.Run()
 }
+
+// checkRegions checks if the given regions are available in AWS.
+func checkRegions(regions []string) (bool, error) {
+	regionChecker, err := getAWSCloudCredentials(awsProfile, "us-east-1")
```
felipemadero, line 355: the return of getAWSCloudCredentials is not a regionChecker object

felipemadero: us-east-1 can be made a constant

pkg/cloud/gcp/gcp.go
```diff
@@ -407,6 +409,49 @@ func (c *GcpCloud) StopGCPNode(nodeConfig models.NodeConfig, clusterName string)
...
+// ListZonesInRegion returns a list of zones in a specific region for a given project ID.
+func (c *GcpCloud) ListZonesInRegion(region string) ([]string, error) {
+	zoneListCall := c.gcpClient.Zones.List(c.projectID)
+	zoneList, err := zoneListCall.Do()
+	if err != nil {
+		return nil, err
+	}
+	zones := []string{}
+	for _, zone := range zoneList.Items {
+		if zone.Region == fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/%s/regions/%s", c.projectID, region) {
```
felipemadero, line 435: can the fmt string be made a constant

cmd/nodecmd/create_aws.go
arturrez: fixed

pkg/cloud/gcp/gcp.go
arturrez: done

cmd/nodecmd/create_aws.go
arturrez: it's a object that we only use to make AWS API call to check region. I am open to a better name. 
moved us-east-1 to the const

cmd/nodecmd/create_aws.go
felipemadero: I mean, to only print the invalid regions, not all the provided. I mean for checkRegions to not return a boolean, but the list of invalid regions

cmd/nodecmd/create_aws.go
felipemadero: call it awsCloud probably, it is more directly related to the data type

pkg/cloud/gcp/gcp.go
felipemadero: in general when I talk about adding a constant. I suggest to add it to the constants package, or to constants header section of the file, or to a constants file inside the same directoty, but not to a constant defined in the same function
maybe we can agree on which style to use for this

cmd/nodecmd/create_aws.go
arturrez: fixed

cmd/nodecmd/create_aws.go
arturrez: fixed name

pkg/cloud/gcp/gcp.go
arturrez: k sorry. will fix it

felipemadero (approved):

Issue #1368, opened 2024-01-10
for clusters with dynamic IPs, IP information differs between node host config and inventory file


felipemadero:
inventory file is ok in relation with aws site info, but node host configs are a permutation of the IPs

felipemadero:
Closed by https://github.com/ava-labs/avalanche-cli/pull/1369

Issue #1367, opened 2024-01-10
aws session start not using env vars AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN


felipemadero:

felipemadero:
Closed by https://github.com/ava-labs/avalanche-cli/pull/1369

PR #1369, opened 2024-01-10
Update cluster dynamic ips

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/1321 https://github.com/ava-labs/avalanche-cli/issues/1368 https://github.com/ava-labs/avalanche-cli/issues/1367

Add a new node command refresh-ips, to be used by the user when there is suspition a dynamic IP has changed


cmd/nodecmd/dynamic_ips.go
```diff
@@ -5,40 +5,42 @@ package nodecmd
...
-	var gcpCloud *gcpAPI.GcpCloud
-	ux.Logger.PrintToUser("Getting Public IPs for nodes without static IPs ...")
-	for _, node := range nodesWoEIP {
+	var (
+		err        error
+		lastRegion string
+		ec2Svc     *awsAPI.AwsCloud
+		gcpCloud   *gcpAPI.GcpCloud
+	)
+	ux.Logger.PrintToUser("Getting Public IPs for nodes with dynamic IPs ...")
```
arturrez, line 42: nit: should be imho `ux.Logger.PrintToUser("Getting Public IP(s) for node(s) ...")`

cmd/nodecmd/dynamic_ips.go
```diff
@@ -71,6 +72,10 @@ func getPublicIPForNodesWoEIP(nodesWoEIP []models.NodeConfig) (map[string]string
 		} else {
 			publicIP, err = ec2Svc.GetInstancePublicIPs([]string{node.NodeID})
 			if err != nil {
+				if strings.Contains(err.Error(), "RequestExpired: Request has expired") {
```
arturrez, line 75: we need a separate func for this as this is a 3rd copy of the same code now

arturrez: maybe also include `err!=nil` there and just pass err into the func

cmd/nodecmd/dynamic_ips.go
```diff
@@ -79,21 +84,49 @@ func getPublicIPForNodesWoEIP(nodesWoEIP []models.NodeConfig) (map[string]string
 	return publicIPMap, nil
 }
 
-func updateAnsiblePublicIPs(clusterName string) error {
+// update public IPs
+// - in ansible inventory file
+// - in host config file
+func updatePublicIPs(clusterName string) error {
 	clusterNodes, err := getClusterNodes(clusterName)
```
arturrez, line 91: we need to also print user that no changes detected and all IPs are up to date

arturrez: if old IPs are equal to new IPs

cmd/nodecmd/dynamic_ips.go
```diff
@@ -79,21 +84,49 @@ func getPublicIPForNodesWoEIP(nodesWoEIP []models.NodeConfig) (map[string]string
...
+				)
+				changed++
+			}
+			node.ElasticIP = publicIPMap[node.NodeID]
+			if err := app.CreateNodeCloudConfigFile(node.NodeID, &node); err != nil { //nolint:gosec
+				return err
+			}
+		}
+		if changed == 0 {
+			ux.Logger.PrintToUser("All node IPs are the same. No update needed")
```
arturrez, line 122: `No changes to IPs detected` might be a better wording. Open for any other suggestions

cmd/nodecmd/refreshIPs.go
```diff
@@ -0,0 +1,34 @@
...
+		Long: `(ALPHA Warning) This command is currently in experimental mode.
+
+The node refresh-ips command obtains the current IP for all nodes with dynamic IPs in the cluster,
+and updates the local node information used by CLI commands.`,
+		SilenceUsage: true,
+		Args:         cobra.ExactArgs(1),
+		RunE:         refreshIPs,
+	}
+
+	cmd.Flags().StringVar(&awsProfile, "aws-profile", constants.AWSDefaultCredential, "aws profile to use")
```
arturrez, line 23: it should accept clusterName as I would different cluster might be deployed using different AWS profiles

arturrez, line 23: should we ask for clustername and  aws profile if it's not provided ?

cmd/nodecmd/refreshIPs.go
```diff
@@ -0,0 +1,34 @@
...
+package nodecmd
+
+import (
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+	"github.com/spf13/cobra"
+)
+
+func newRefreshIPsCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "refresh-ips [clusterName]",
```
arturrez, line 12: i think just `refresh` is a better name

pkg/cloud/aws/aws.go
```diff
@@ -36,12 +36,25 @@ type AwsCloud struct {
...
-	// Load session from shared config
-	cfg, err := config.LoadDefaultConfig(ctx,
-		config.WithRegion(region),
-		config.WithSharedConfigProfile(awsProfile),
+	var (
+		cfg aws.Config
+		err error
 	)
+	ctx := context.Background()
+	if os.Getenv("AWS_ACCESS_KEY_ID") != "" {
```
arturrez, line 44: do we need this code? 
https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#environment-variables - `By default, the SDK detects AWS credentials set in your environment and uses them to sign requests to AWS.`

arturrez (commented):
commented

cmd/nodecmd/refreshIPs.go
felipemadero: clusterName is the command positional argument

felipemadero: clusterName is the positional argument, we usually fail for that one
regarding awsProfile, we usually defer it to flag for the other commands

pkg/cloud/aws/aws.go
felipemadero: yes we need this. the aws sdk doc seems to be incorrect. at least for linux. please give it a test.

felipemadero: if you use `config.WithSharedConfigProfile` it will ignore env vars on linux at least

cmd/nodecmd/dynamic_ips.go
felipemadero: the code is slightly different depending on the caller. sometimes is returns nil, sometimes it returns err, sometimes it prints an empty line, probably we want to keep the three codes

felipemadero: adding a common function to just check the condition

pkg/cloud/aws/aws.go
arturrez: I can't reproduce it @felipemadero 
```
❯ export AWS_ACCESS_KEY_ID="XXXXXXXX"
❯ export AWS_SECRET_ACCESS_KEY="YYYYYYYYY"
❯ env | grep AWS
AWS_PAGER=
AWS_ACCESS_KEY_ID=XXXXX
AWS_SECRET_ACCESS_KEYYUYYYYYY
❯ ./bin/avalanche node create test3  --use-static-ip=false --latest-avalanchego-version=true --aws  --region us-east-1 --num-nodes=1 --devnet
✔ c5.2xlarge (default)
Creating new EC2 instance(s) on AWS...
Using existing key pair artur-us-east-1-avalanche-cli in AWS[us-east-1]
Using existing security group artur-us-east-1-avalanche-cli-us-east-1-sg in AWS[us-east-1]
Waiting for EC2 instances in AWS[us-east-1] to be provisioned...
New EC2 instance(s) successfully created in AWS!
```

felipemadero: I can, it is a shame (ubuntu 22.10)

```
export AWS_ACCESS_KEY_ID="..."
export AWS_SECRET_ACCESS_KEY="..."
export AWS_SESSION_TOKEN="..."
 env | grep AWS
 AWS_SECRET_ACCESS_KEY=...
 AWS_ACCESS_KEY_ID=...
 AWS_SESSION_TOKEN=...
 ./bin/avalanche node create test3  --use-static-ip=false --latest-avalanchego-version=true --aws  --region us-east-1 --num-nodes=1 --devnet
 ✔ c5.2xlarge (default)
AWS credentials expired
Please update your environment variables AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY
Following https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html#envvars-set
Or fill in ~/.aws/credentials with updated contents following the format below
===========BEGINNING OF FILE===========
[default]
aws_access_key_id=<AWS_ACCESS_KEY>
aws_secret_access_key=<AWS_SECRET_ACCESS_KEY>
===========END OF FILE===========
More info can be found at https://docs.aws.amazon.com/sdkref/latest/guide/file-format.html#file-format-creds

Error: operation error EC2: DescribeImages, https response error StatusCode: 400, RequestID: 0ade8d20-c4b9-4517-abb8-b721d90d1e1a, api error RequestExpired: Request has expired.

```

arturrez: you are correct. It's my macosx having default AWS profile setup

PR #1382, opened 2024-01-17
Monitoring GCP without TF

sukantoraymond:


pkg/cloud/gcp/gcp.go
```diff
@@ -156,7 +159,7 @@ func (c *GcpCloud) SetupNetwork(ipAddress, networkName string) (*compute.Network
 
 // SetFirewallRule creates a new firewall rule in GCP
 func (c *GcpCloud) SetFirewallRule(ipAddress, firewallName, networkName string, ports []string) (*compute.Firewall, error) {
-	if strings.Contains(ipAddress, "/") {
+	if strings.Contains(ipAddress, "/") && ipAddress != "0.0.0.0/0" {
```
arturrez, line 162: do I read it correctly?
it should be `!strings.Contains(ipAddress, "/")` if I understand this code

pkg/cloud/gcp/gcp.go
```diff
@@ -229,6 +232,12 @@ func (c *GcpCloud) SetupInstances(zone, networkName, sshPublicKey, ami string, s
 	eg.SetLimit(parallelism)
 	for i := 0; i < numNodes; i++ {
 		currentIndex := i
+		var cloudDiskSize int64
```
arturrez, line 235: nit: simplify with 
``cloudDiskSize:=1000
if forMonitoring{cloudDiskSize =10}
``

pkg/cloud/gcp/gcp.go
```diff
@@ -407,6 +423,36 @@ func (c *GcpCloud) StopGCPNode(nodeConfig models.NodeConfig, clusterName string)
 	return nil
 }
 
+// AddFirewall adds firewall into an existing project in GCP
+func (c *GcpCloud) AddFirewall(monitoringHostPublicIP, networkName, projectName string) error {
+	firewallName := fmt.Sprintf("%s-%s-monitoring", networkName, strings.ReplaceAll(monitoringHostPublicIP, ".", ""))
```
arturrez, line 428: nit: AddFirewall is generic function to add FW rules, `monitoring` should not be hardcoded in FWname

pkg/cloud/gcp/gcp.go
```diff
@@ -407,6 +423,36 @@ func (c *GcpCloud) StopGCPNode(nodeConfig models.NodeConfig, clusterName string)
 	return nil
 }
 
+// AddFirewall adds firewall into an existing project in GCP
+func (c *GcpCloud) AddFirewall(monitoringHostPublicIP, networkName, projectName string) error {
```
arturrez, line 427: ports, name and etc should be passed to this `func` as params

Issue #1374, opened 2024-01-11
there are no flag settings to create a default subnet evm conf with given chain ID and token name


felipemadero:
it always requires user interaction with prompts

PR #1383, opened 2024-01-17
create a default subnet evm config using flags only

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/1374

Adds flags to subnet evm config generation, so that default configs can be generated without user interaction.

Example of a full devnet + subnet-evm generation:
```
avalanche node devnet wiz cluster1 subnet1 --force-subnet-create --evm-subnet\
  --evm-chain-id 1 --evm-token PEPE --latest-evm-version --evm-defaults=true\
  --default-validator-params --authorize-access --aws --node-type default --num-nodes 5\
  --region us-east-1 --use-static-ip=true
```


cmd/subnetcmd/create.go
```diff
@@ -63,6 +66,9 @@ configuration, pass the -f flag.`,
 	cmd.Flags().StringVar(&genesisFile, "genesis", "", "file path of genesis to use")
 	cmd.Flags().BoolVar(&useSubnetEvm, "evm", false, "use the Subnet-EVM as the base template")
 	cmd.Flags().StringVar(&evmVersion, "vm-version", "", "version of Subnet-Evm template to use")
```
arturrez, line 68: nit: should this name be `evm-version` for consistency

arturrez (approved):

PR #1479, opened 2024-02-21
improve error message for node setup

arturrez:


pkg/ssh/shell/setupNode.sh
```diff
@@ -4,9 +4,10 @@ set -e
 #name:TASK [disable systemctl]
 sudo cp -vf /usr/bin/true /usr/local/sbin/systemctl
 {{end}}
-#name:TASK [update apt data and install dependencies] 
-DEBIAN_FRONTEND=noninteractive sudo apt-get -y update -o DPkg::Lock::Timeout=120
-DEBIAN_FRONTEND=noninteractive sudo apt-get -y install -o DPkg::Lock::Timeout=120 wget curl git 
+#name:TASK [update apt data and install dependencies]
+export DEBIAN_FRONTEND=noninteractive
+until sudo apt-get -y update -o DPkg::Lock::Timeout=120; do sleep 1 && echo "Try again"; done
```
felipemadero, line 9: could you limit the amount of iterations?

felipemadero (commented):
left one comment. would be nice to have an issue that tells which error was observed so as to do this (the until loop, but also the Timeout=120 setting)

pkg/ssh/shell/setupNode.sh
arturrez: we should not start until our  apt pkg information is up to date 
we have timeout on ssh command so it will fail eventually, it will not get stuck

felipemadero (approved):

arturrez:
lets see if we will need this or not. merging it for now as it improves things for sure

Issue #1188, opened 2023-11-13
organize cli output


arturrez:
* make it less verbose
* move node create commands from scripts into debug node and present user with https://github.com/briandowns/spinner 
* make output more organised with less text to read 
* instrument https://github.com/briandowns/spinner all communications with avalanchego or ssh so user is waiting \

good example kind cluster create command

PR #1490, opened 2024-02-23
Node create output improvements

arturrez:
ref: https://github.com/ava-labs/avalanche-cli/issues/1188

image

image


arturrez:
> Screenshot 2024-03-04 at 7 44 30 PM Why not we make the output like this? Only the node id, cloud id, and the public ip

I feel like we should reduce output more, but encourage user to run status command and get this output

arturrez:
> We also need to add:
> 
> to ssh to cluster / node : run `node ssh <clusterName | nodeID>`, go to https://docs.avax.network/tooling/cli-create-nodes/node-ssh for more details

added but in a little bit different wording. Hope that it will be ok. Want to keep things short

arturrez:
> Cloud Instance ID



> So the new output is : Node i-01a5bee72b6082f09[NodeID-GWUiEcMWNJAnoH7j2Ppd8i7cixsyfBweM]|ip:18.217.147.170
> 
> Compared to old output : node aws_node_i-0edbf9a1b361192f1 details: Cloud Instance ID: i-0edbf9a1b361192f1 Public IP: 3.131.40.131 Cloud Region: us-east-2
> 
> I think it should be Cloud Instance ID: i-01a5bee72b6082f09 | Public IP: 18.217.147.170
> 
> Can be confusign with two different node ids in one line

Changed the line. I still feel like we should display NodeID, as we also display it for `node status` and it can be used for `node ssh`. my change leaves it but makes better separation so it's not confusing for the user( I hope 🤞 )

PR #1495, opened 2024-02-27
Loadtest subnetevm

sukantoraymond:


cmd/nodecmd/create.go
```diff
@@ -814,6 +819,9 @@ func addNodeToClustersConfig(network models.Network, nodeID, clusterName string,
 		clustersConfig.Clusters = make(map[string]models.ClusterConfig)
 	}
 	if _, ok := clustersConfig.Clusters[clusterName]; !ok {
+		if isMonitoringInstance {
+			return fmt.Errorf("cluster %s is not found in cluster_config.json", clusterName)
```
arturrez, line 823: not sure if error message is correct.  we check ifMonitoringInstance but return err for cluster not found

cmd/nodecmd/create.go
```diff
@@ -1188,6 +1197,50 @@ func requestCloudAuth(cloudName string) error {
...
+	string,
+	error,
+) {
+	type CloudPrompt struct {
+		defaultLocations []string
+		locationName     string
+		locationsListURL string
+	}
+
+	supportedClouds := map[string]CloudPrompt{
```
arturrez, line 1210: it's the same as `getRegionsNodeNum` - we should move it out from both func so it can be DRY. I suggest moving it to cloud/aws and cloud/gcp as a constants.

another option is to change getRegionsNodeNum func to accomodate both cases when we want to accept slice or single value so no need to loop through prompts

cmd/nodecmd/create.go
```diff
@@ -1188,6 +1197,50 @@ func requestCloudAuth(cloudName string) error {
...
+			locationName:     "Google Region",
+			locationsListURL: "https://cloud.google.com/compute/docs/regions-zones/",
+		},
+	}
+
+	if _, ok := supportedClouds[cloudName]; !ok {
+		return "", fmt.Errorf("cloud %s is not supported", cloudName)
+	}
+
+	awsCustomRegion := fmt.Sprintf("Choose custom %s (list of %ss available at %s)", supportedClouds[cloudName].locationName, supportedClouds[cloudName].locationName, supportedClouds[cloudName].locationsListURL)
```
arturrez, line 1227: this code is very similar to getRegionsNodeNum - let's change this func to work for slice and single input

cmd/nodecmd/create_aws.go
```diff
@@ -95,16 +96,24 @@ func getAWSMonitoringEC2Svc(awsProfile, monitoringRegion string) (map[string]*aw
...
 	finalRegions := map[string]int{}
 	switch {
 	case len(numNodes) != len(utils.Unique(cmdLineRegion)):
 		return nil, nil, nil, fmt.Errorf("number of nodes and regions should be the same")
 	case len(cmdLineRegion) == 0 && len(numNodes) == 0:
 		var err error
-		finalRegions, err = getRegionsNodeNum(constants.AWSCloudService)
-		if err != nil {
-			return nil, nil, nil, err
+		if singleNode {
```
arturrez, line 106: we should not do this as it complicates logic on the func.
let's make getRegionsNodeNum work and keep getAWSCloudConfig unchanged

cmd/nodecmd/create_aws.go
```diff
@@ -122,13 +131,24 @@ func getAWSCloudConfig(awsProfile string) (map[string]*awsAPI.AwsCloud, map[stri
 	}
 	for region := range finalRegions {
 		var err error
-
-		ec2SvcMap[region], err = getAWSCloudCredentials(awsProfile, region)
-		if err != nil {
-			if !strings.Contains(err.Error(), "cloud access is required") {
-				printNoCredentialsOutput(awsProfile)
+		if singleNode {
+			for _, clusterRegion := range clusterSgRegions {
```
arturrez, line 135: why do we need to pass clusterRegions as we have this info already for the cluster we work with

cmd/nodecmd/create_aws.go
```diff
@@ -355,6 +375,29 @@ func AddMonitoringSecurityGroupRule(ec2Svc map[string]*awsAPI.AwsCloud, monitori
 	return nil
 }
 
+func grantAccessToPublicIPViaSecurityGroup(ec2Svc *awsAPI.AwsCloud, publicIP, securityGroupName, region string) error {
```
arturrez, line 378: this is exactly the same func as `AddMonitoringSecurityGroupRule` why not to make it work for both cases

cmd/nodecmd/create_aws.go
```diff
@@ -355,6 +375,29 @@ func AddMonitoringSecurityGroupRule(ec2Svc map[string]*awsAPI.AwsCloud, monitori
...
+	securityGroupExists, sg, err := ec2Svc.CheckSecurityGroupExists(securityGroupName)
+	if err != nil {
+		return err
+	}
+	if !securityGroupExists {
+		return fmt.Errorf("security group %s doesn't exist in region %s", securityGroupName, region)
```
arturrez, line 384: `region` is only used for error msg, is it worth passing it only for this

arturrez: anyway, it's not that important vs not using this function at all as https://github.com/ava-labs/avalanche-cli/pull/1495/files#r1513554306

cmd/nodecmd/create_gcp.go
```diff
@@ -93,16 +93,24 @@ func getGCPCloudCredentials() (*compute.Service, string, string, error) {
...
 	finalRegions := map[string]int{}
 	switch {
 	case len(numNodes) != len(utils.Unique(cmdLineRegion)):
 		return nil, nil, "", "", "", errors.New("number of regions and number of nodes must be equal. Please make sure list of regions is unique")
 	case len(cmdLineRegion) == 0 && len(numNodes) == 0:
 		var err error
-		finalRegions, err = getRegionsNodeNum(constants.GCPCloudService)
-		if err != nil {
-			return nil, nil, "", "", "", err
+		if singleNode {
```
arturrez, line 103: no need to this if statement and `getSeparateHostNodeParam` as `getRegionsNodeNum`can be changed and used

cmd/nodecmd/create_gcp.go
```diff
@@ -365,3 +373,26 @@ func updateClustersConfigGCPKeyFilepath(projectName, serviceAccountKeyFilepath s
 	}
 	return app.WriteClustersConfigFile(&clustersConfig)
 }
+
+func grantAccessToPublicIPViaFirewall(gcpClient *gcpAPI.GcpCloud, projectName string, publicIP string, label string) error {
```
arturrez, line 377: we should also use it to replace ```if separateMonitoringInstance {
					prefix, err := defaultAvalancheCLIPrefix("")
					if err != nil {
						return err
					}
					networkName := fmt.Sprintf("%s-network", prefix)
					firewallName := fmt.Sprintf("%s-%s-monitoring", networkName, strings.ReplaceAll(monitoringNodeConfig.PublicIPs[0], ".", ""))
					ports := []string{
						strconv.Itoa(constants.AvalanchegoMachineMetricsPort), strconv.Itoa(constants.AvalanchegoAPIPort),
						strconv.Itoa(constants.AvalanchegoMonitoringPort), strconv.Itoa(constants.AvalanchegoGrafanaPort),
					}
					if err = gcpClient.AddFirewall(
						monitoringNodeConfig.PublicIPs[0],
						networkName,
						projectName,
						firewallName,
						ports,
						true); err != nil {
						return err
					}
				}```
as we have it in couple places

cmd/nodecmd/loadtest.go
```diff
@@ -0,0 +1,302 @@
...
+			return err
+		}
+		cloudService = separateNodeConfig.CloudService
+	} else {
+		ux.Logger.PrintToUser("Creating a separate instance to run load test...")
+		cloudService, err = setCloudService()
+		if err != nil {
+			return err
+		}
+		nodeType = "default"
```
arturrez, line 123: nit: use `const.DefaultNodeType`

cmd/nodecmd/loadtest.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	separateHostRegion := ""
+	cloudSecurityGroupList, err := getCloudSecurityGroupList(clusterNodes)
+	if err != nil {
+		return err
+	}
+	// we only support endpoint in the same cluster
+	filteredSGList := utils.Filter(cloudSecurityGroupList, func(sg regionSecurityGroup) bool { return sg.cloud == cloudService })
+	if len(filteredSGList) == 0 {
+		return fmt.Errorf("no endpoint found in the  %s", cloudService)
```
arturrez, line 137: nit: rm double space

cmd/nodecmd/loadtest.go
```diff
@@ -0,0 +1,302 @@
...
+	if err := GetLoadTestScript(app); err != nil {
+		return err
+	}
+
+	// waiting for all nodes to become accessible
+	if existingSeparateInstance == "" {
+		failedHosts := waitForHosts(separateHosts)
+		if failedHosts.Len() > 0 {
+			for _, result := range failedHosts.GetResults() {
+				ux.Logger.PrintToUser("Instance %s failed to provision with error %s. Please check instance logs for more information", result.NodeID, result.Err)
```
arturrez, line 259: nit: use ux.Logger.RedXToUser

cmd/nodecmd/loadtest.go
```diff
@@ -0,0 +1,302 @@
...
+			}
+			return fmt.Errorf("failed to provision node(s) %s", failedHosts.GetNodeList())
+		}
+		ux.Logger.PrintToUser("Separate instance %s provisioned successfully", separateHosts[0].NodeID)
```
arturrez, line 263: nit: use ux.Logger.GreenCheckBoxToUser

cmd/nodecmd/loadtest.go
arturrez: I suggest using RedXToUser every time we try to use ux.Logger.PrintToUser before returning err

cmd/nodecmd/loadtest.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	if err := ssh.RunSSHBuildLoadTest(separateHosts[0], loadTestRepoURL, loadTestBuildCmd); err != nil {
```
arturrez, line 265: nit: add spinner there

pkg/ssh/ssh.go
```diff
@@ -467,3 +495,32 @@ func RunSSHSubnetSyncStatus(host *models.Host, blockchainID string) ([]byte, err
...
+	err = t.Execute(&script, templateVars)
+	if err != nil {
+		return err
+	}
+
+	if err := host.StreamSSHCommand(script.String(), nil, timeout); err != nil {
+		return err
+	}
+	return nil
+}
```
arturrez, line 526: nice to have:
add e2e for loadtest

pkg/ssh/ssh.go
```diff
@@ -446,6 +422,29 @@ func RunSSHSetupBuildEnv(host *models.Host) error {
 	)
 }
 
+func RunSSHBuildLoadTest(host *models.Host, loadTestRepo, loadTestPath string) error {
+	loadTestRepoPaths := strings.Split(loadTestRepo, "/")
+	// remove .git
+	loadTestRepoDir := strings.Split(loadTestRepoPaths[len(loadTestRepoPaths)-1], ".")
```
arturrez, line 428: nit: use `name := strings.TrimSuffix(basename, filepath.Ext(basename))`

pkg/ssh/ssh.go
```diff
@@ -446,6 +422,29 @@ func RunSSHSetupBuildEnv(host *models.Host) error {
...
+	)
+}
+
+func RunSSHRunLoadTest(host *models.Host, loadTestCommand string) error {
+	return StreamOverSSH(
+		"Run Load Test",
+		host,
+		constants.SSHScriptTimeout,
+		"shell/runLoadTest.sh",
+		scriptInputs{GoVersion: constants.BuildEnvGolangVersion, LoadTestCommand: loadTestCommand},
```
arturrez, line 444: goversion is not used in the script

patrick-ogrady (approved):

PR #1538, opened 2024-03-06
Add API/Monitoring node info to node list/status + fix certain bugs/nits in node cmds

felipemadero:
Bugs/Nits:
- separate monitogin node was not beign removed if created in devnet
- nodes non beign removed if trying to create nodes and the addrs limit is full
- outdated credentials msg not beign show if calling checkRegions
- if not elastic addresses were generated for addrs limit, recovery code anyway tried to release elastic addresses
- IP was not saved on node config for separate monitoring node if using dynamic IPs


cmd/nodecmd/list.go
```diff
@@ -39,29 +41,41 @@ func list(_ *cobra.Command, _ []string) error {
...
-				return err
+			nodeIDStr := "----------------------------------------"
+			funcs := []string{}
+			if clusterConf.MonitoringInstance != cloudID {
+				nodeID, err := getNodeID(app.GetNodeInstanceDirPath(cloudID))
+				if err != nil {
+					return err
+				}
+				nodeIDStr = nodeID.String()
+				if clusterConf.IsAPIHost(cloudID) {
```
arturrez, line 65: nit: abstract adding `API` `Node` and `Monitor` to some function, so output is consistent everywhere

cmd/nodecmd/ssh.go
```diff
@@ -75,6 +77,11 @@ func sshNode(_ *cobra.Command, args []string) error {
 			// try to detect nodeID
 			for clusterName := range clustersConfig.Clusters {
 				clusterHosts, _ := ansible.GetInventoryFromAnsibleInventoryFile(app.GetAnsibleInventoryDirPath(clusterName))
+				monitoringInventoryPath := filepath.Join(app.GetAnsibleInventoryDirPath(clusterName), constants.MonitoringDir)
```
arturrez, line 80: why do we want to include monitor host to ssh command?
are yiou trying to allow ssh command execution for hostID for monitoring ?

cmd/nodecmd/status.go
```diff
@@ -234,46 +232,95 @@ func printOutput(
...
+				}
 			}
 			row = append(row, syncedStatus)
 		}
 		table.Append(row)
 	}
 	table.Render()
 }
+
+func removeColors(s string) string {
```
arturrez, line 300: you can use https://pkg.go.dev/github.com/pborman/ansi#Strip for this.
no need to hardcode list of colours

pkg/cloud/aws/aws.go
```diff
@@ -309,23 +309,27 @@ func (c *AwsCloud) StopInstance(instanceID, publicIP string, releasePublicIP boo
...
-		if len(addressOutput.Addresses) == 0 {
-			return ErrNoAddressFound
-		}
-		releaseAddressInput := &ec2.ReleaseAddressInput{
-			AllocationId: aws.String(*addressOutput.Addresses[0].AllocationId),
-		}
-		if _, err = c.ec2Client.ReleaseAddress(c.ctx, releaseAddressInput); err != nil {
-			return err
+		if publicIP == "" {
+			ux.Logger.PrintToUser("Unabled to remove public IP for instannce %s: undefined", instanceID)
```
arturrez, line 313: nit: use ux.Logger.RedXToUser

cmd/nodecmd/list.go
```diff
@@ -39,29 +41,41 @@ func list(_ *cobra.Command, _ []string) error {
...
+				} else {
+					funcs = append(funcs, "Node")
+				}
+			}
+			if nodeConfig.IsMonitor {
+				funcs = append(funcs, "Monitor")
+			}
+			funcDesc := strings.Join(funcs, ",")
+			if funcDesc != "" {
+				funcDesc = " [" + funcDesc + "]"
```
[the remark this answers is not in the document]
felipemadero: I will just abstract the list of roles

cmd/nodecmd/ssh.go
felipemadero: yes. I am adding that so can use ssh command also with the monitoring instance ip, cloud id, and whatever

felipemadero: this command should give access to all resources

cmd/nodecmd/ssh.go
```diff
@@ -69,12 +74,31 @@ func sshNode(_ *cobra.Command, args []string) error {
 				if err != nil {
 					return err
 				}
+				monitoringInventoryPath := filepath.Join(app.GetAnsibleInventoryDirPath(clusterNameOrNodeID), constants.MonitoringDir)
```
arturrez, line 77: we should not allow running ssh commands for the cluster on monitoring instance as it's very different.  Everything else is the same so it makes sense to allow it as all boxes are deployed exactly the same so there is value in running one command across cluster.
For monitoring intance as it's not running avalanchego - it should be excluded from cluster wide ssh commands.

But I still see value allowing users to reference monitoring instance per cloud instance ID to run ssh command or open terminal sessions

arturrez (commented):
https://github.com/ava-labs/avalanche-cli/pull/1538/files#r1515124123

arturrez (approved):

PR #1562, opened 2024-03-12
Terminate instances

sukantoraymond:
Changes stop command to terminate

Also changes functionality from stopping node to terminating node


arturrez (approved):
LGTM. Approved.

The only comment I have if we want to keep `StopInstance` and add `Terminate` for cloud  in case we might need it for something in the future

PR #1568, opened 2024-03-12
Git commit loadtest

sukantoraymond:


cmd/nodecmd/loadtest.go
```diff
@@ -404,10 +411,32 @@ func createClusterYAMLFile(clusterName, subnetID, chainID string, separateHost *
 	return enc.Encode(clusterInfoYAML)
 }
 
+func getGitCommit(gitRepoURL string) string {
```
arturrez, line 414: move to utils/common
add unittests

cmd/nodecmd/loadtest.go
```diff
@@ -404,10 +411,32 @@ func createClusterYAMLFile(clusterName, subnetID, chainID string, separateHost *
...
+	if strings.Contains(gitRepoURL, "/commit/") {
```
arturrez, line 415: add check that gitRepoUrl is a valid url
add check len(splitURL)>=1

cmd/nodecmd/loadtest.go
```diff
@@ -404,10 +411,32 @@ func createClusterYAMLFile(clusterName, subnetID, chainID string, separateHost *
...
+		splitURL := strings.Split(gitRepoURL, "/")
+		commitID := splitURL[len(splitURL)-1]
+		return commitID
+	}
+	return ""
+}
+
+func getRepoFromCommitURL(gitRepoURL string) string {
+	splitURL := strings.Split(gitRepoURL, "/")
```
arturrez, line 424: we need to check if url is valid, we need checks for len of the slice to make sure it doesn't crash

cmd/nodecmd/loadtest.go
```diff
@@ -404,10 +411,32 @@ func createClusterYAMLFile(clusterName, subnetID, chainID string, separateHost *
...
+	splitURLWOCommit := splitURL[:len(splitURL)-2]
+	repoDirName = splitURLWOCommit[len(splitURLWOCommit)-1]
+	gitRepo := strings.Join(splitURLWOCommit[:], "/")
+	gitRepo += ".git"
```
arturrez, line 428: nit: we have `const gitExtension = ".git"` maybe we should use it there too so we reuse const

cmd/nodecmd/loadtest.go
```diff
@@ -418,6 +447,10 @@ func GetLoadTestScript(app *application.Avalanche) error {
 		if err != nil {
 			return err
 		}
+		loadTestRepoCommit = getGitCommit(loadTestRepoURL)
```
arturrez, line 450: why do we do it twice ? 
maybe move it up from the if{} statement

pkg/ssh/shell/buildLoadTest.sh
```diff
@@ -2,6 +2,9 @@
 # get most updated load test repo
 echo "getting load test repo ..."
 git -C {{ .LoadTestRepoDir }} pull || git clone {{ .LoadTestRepo }}
+{{if .CheckoutCommit }}
+cd {{ .RepoDirName}}; git checkout {{ .LoadTestGitCommit}}
```
arturrez, line 6: nit: use `pushd` `popd` so `cwd` states the same after the operation

pkg/ssh/ssh.go
```diff
@@ -433,7 +436,7 @@ func RunSSHSetupBuildEnv(host *models.Host) error {
 	)
 }
 
-func RunSSHBuildLoadTest(host *models.Host, loadTestRepo, loadTestPath string) error {
+func RunSSHBuildLoadTest(host *models.Host, loadTestRepo, loadTestPath, loadTestGitCommit, repoDirName string, checkoutCommit bool) error {
 	loadTestRepoPaths := strings.Split(loadTestRepo, "/")
 	// remove .git
 	loadTestRepoDir := strings.Split(loadTestRepoPaths[len(loadTestRepoPaths)-1], ".")
```
arturrez, line 442: nit: check for len of slice to make sure it doesn't crash

pkg/ssh/shell/buildLoadTest.sh
```diff
@@ -2,6 +2,9 @@
 # get most updated load test repo
 echo "getting load test repo ..."
 git -C {{ .LoadTestRepoDir }} pull || git clone {{ .LoadTestRepo }}
+{{if .CheckoutCommit }}
```
arturrez, line 5: do we need `.CheckoutCommit`  if we can check `.LoadTestGitCommit` is not empty ?

cmd/nodecmd/loadtest.go
```diff
@@ -404,10 +411,32 @@ func createClusterYAMLFile(clusterName, subnetID, chainID string, separateHost *
...
+		splitURL := strings.Split(gitRepoURL, "/")
+		commitID := splitURL[len(splitURL)-1]
+		return commitID
+	}
+	return ""
+}
+
+func getRepoFromCommitURL(gitRepoURL string) string {
+	splitURL := strings.Split(gitRepoURL, "/")
+	splitURLWOCommit := splitURL[:len(splitURL)-2]
```
arturrez, line 425: add comment with url example so it's easy to read this logic

patrick-ogrady (approved):

Issue #1596, opened 2024-03-14
speedup node creation


arturrez:
- make go and rust install in parallel
- make monitoring install in parallel

PR #1597, opened 2024-03-14
Speedup setup build env

arturrez:
resolves: https://github.com/ava-labs/avalanche-cli/issues/1596 
speedup: 103s -> 95s for 2 nodes cluster

`./bin/avalanche node create lt0 --use-static-ip=false --latest-avalanchego-version=true --aws  --region us-east-1 --num-validators=2 --devnet --num-apis=0  --node-type=c5.2xlarge --skip-monitoring`

### DIFF
--- a/pkg/ssh/shell/setupBuildEnv.sh
+++ b/pkg/ssh/shell/setupBuildEnv.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
 set -e
 #name:TASK [install gcc if not available]
+sudo apt-get -y -o DPkg::Lock::Timeout=120 update
 gcc --version || DEBIAN_FRONTEND=noninteractive sudo apt-get -y -o DPkg::Lock::Timeout=120 install gcc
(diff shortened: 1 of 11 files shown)


pkg/ssh/ssh.go
```diff
@@ -429,13 +430,36 @@ func RunSSHUpdateSubnet(host *models.Host, subnetName, importPath string) error
...
 		constants.SSHScriptTimeout,
 		"shell/setupBuildEnv.sh",
-		scriptInputs{GoVersion: constants.BuildEnvGolangVersion},
-	)
+		scriptInputs{},
+	); err != nil {
+		return err
+	}
+	eg := &errgroup.Group{}
+	eg.SetLimit(2)
```
patrick-ogrady, line 443: I don't think you need to SetLimit for this?

pkg/ssh/shell/setupBuildEnv.sh
```diff
@@ -3,22 +3,3 @@ set -e
 #name:TASK [install gcc if not available]
 sudo apt-get -y -o DPkg::Lock::Timeout=120 update
```
patrick-ogrady, line 4: I wonder if you can do this concurrently? I don't think it'll be used by the installs?

arturrez: we can delete this actually as we do it in pkg/ssh/shell/setupNode.sh

pkg/ssh/ssh.go
arturrez: sure.

pkg/ssh/shell/setupBuildEnv.sh
arturrez: no we can't unfortunately

arturrez:
closing in favour of https://github.com/ava-labs/avalanche-cli/pull/1608

PR #1631, opened 2024-03-19
Start stop relayer

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/1580


cmd/teleportercmd/startRelayer.go
```diff
@@ -0,0 +1,78 @@
...
+		false,
+		startRelayerNetworkOptions,
+		"",
+	)
+	if err != nil {
+		return err
+	}
+	switch {
+	case network.Kind == models.Local:
+		b, _, _, err := teleporter.RelayerIsUp(
```
arturrez, line 47: nit: var name `b` is too short to understand what it means, hard to read code

cmd/teleportercmd/stopRelayer.go
```diff
@@ -0,0 +1,74 @@
...
+		if err != nil {
+			return err
+		}
+		if err := ssh.RunSSHStopAWMRelayerService(host); err != nil {
+			return err
+		}
+		ux.Logger.PrintToUser("Remote AWM Relayer on %s successfully stopped", host.GetCloudID())
+	}
+	return nil
+}
```
arturrez, line 74: overall nit: use GreenCheckMartToUser instead of PrintToUser for better cli UI.

pkg/node/node.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func GetAWMRelayerHost(app *application.Avalanche, clusterName string) (*models.Host, error) {
+	clusterConfig, err := app.GetClusterConfig(clusterName)
+	if err != nil {
+		return nil, err
+	}
+	relayerCloudID := ""
+	for _, cloudID := range clusterConfig.GetCloudIDs() {
+		nodeConfig, err := app.LoadClusterNodeConfig(cloudID)
```
arturrez, line 40: nit: can be shortened with if nodeConfig, err :=...; err!=nil ... else if nodeConfig.IsAWMRelayer ...

pkg/teleporter/relayer.go
```diff
@@ -115,38 +115,51 @@ func DeployRelayer(
...
+
+func RelayerCleanup(runFilePath string, storageDir string) error {
+	if err := os.RemoveAll(storageDir); err != nil {
+		return err
+	}
+	b, pid, proc, err := RelayerIsUp(runFilePath)
+	if err != nil {
+		return err
+	}
+	if b {
```
arturrez, line 153: what's `b` we need better name

pkg/teleporter/relayer.go
```diff
@@ -115,38 +115,51 @@ func DeployRelayer(
...
+		if err := proc.Signal(os.Interrupt); err != nil {
+			ux.Logger.PrintToUser("failed trying to kill awm relayer with SIGINT. Using SIGKILL instead")
```
arturrez, line 155: do we want to wait for the process to handle os.Interrupt signal before we try to kill it?  err is failure to send a signal, but it doesn't show us that process exited or not.  Also no need to show user internals what signal we use to kill it.  
maybe use https://pkg.go.dev/os#Process.Wait and https://pkg.go.dev/os#Process.Kill.
The question is if we want to be nice to the process or not and can just afford killing it to simplify logic

arturrez (commented):
overall LGTM with some nits. only 1 concern

Issue #1582, opened 2024-03-13
[Teleporter] Add Teleporter options to subnet create


felipemadero:
Avalanche-cli automatically spins up an AWM Relayer that is automatically configured when a subnet is deployed. 

However, some developers would like to modify the Relayer and then use that modified Relayer to relay these messages.

By offering an option to not run a Relayer, they could just run a modified Relayer image manually.

This can be best achieved by adding this as an option to the subnet create flow:

Ask if subnet should be Teleporter  enabled.

1.) Yes (default): deploy Teleporter and registry and run latest Relayer

2.) No: do nothing

3.) Custom:

3a) Ask if Teleporter and registry should be deployed

3b) Ask if AWM Relayer should be run. Maybe also give the options 
- yes, latest
- yes, specific release version (might be tricky if configurations format changes) 
- no

PR #1639, opened 2024-03-20
add option to not deploy relayer on subnet config

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/1582


cmd/subnetcmd/create.go
```diff
@@ -90,7 +92,9 @@ configuration, pass the -f flag.`,
 	cmd.Flags().StringVar(&customVMBranch, "custom-vm-branch", "", "custom vm branch or commit")
 	cmd.Flags().StringVar(&customVMBuildScript, "custom-vm-build-script", "", "custom vm build-script")
 	cmd.Flags().BoolVar(&useRepo, "from-github-repo", false, "generate custom VM binary from github repository")
-	cmd.Flags().BoolVar(&teleporterReady, "teleporter", true, "generate a teleporter-ready vm")
+	cmd.Flags().BoolVar(&useWarp, "warp", true, "generate a vm with warp support (needed for teleporter)")
```
arturrez, line 95: nit: wording should be consistent, right now we have `warp` `teleporter` `run-relayer` and the last one should be `relayer`. Another option do `with-warp` `with-releporter` `with-relayer`. Wondering what @sukantoraymond opinion on this. Pls feel free to ignore it

arturrez, line 95: we need to check for situation when warp is false and teleporter is true. should we return err in this case

arturrez (approved):

cmd/subnetcmd/create.go
felipemadero: using "relayer"

PR #1643, opened 2024-03-21
Remove security group during delete

sukantoraymond:


cmd/nodecmd/destroy.go
```diff
@@ -145,6 +145,9 @@ func destroyNodes(_ *cobra.Command, args []string) error {
 				}
 				ux.Logger.PrintToUser("node %s is already destroyed", nodeConfig.NodeID)
 			}
+			if err = deleteMonitoringSecurityGroupRule(map[string]*awsAPI.AwsCloud{nodeConfig.Region: ec2Svc}, nodeConfig.ElasticIP, nodeConfig.SecurityGroup, nodeConfig.Region); err != nil {
+				return err
```
arturrez, line 149: not sure if we want to return error and stop everything if we can't delete security rule. I would suggest display an error and carry on

arturrez (approved):

Issue #1678, opened 2024-03-26
if cluster create fails before staking keys generation, node list does not work anymore


felipemadero:
fails trying to find staking key files

PR #1679, opened 2024-03-26
on node list, avoid failing on error if there are not staking keys

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/1678

### DIFF
--- a/cmd/nodecmd/list.go
+++ b/cmd/nodecmd/list.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/ava-labs/avalanche-cli/pkg/models"
 	"github.com/ava-labs/avalanche-cli/pkg/ux"
-	"github.com/ava-labs/avalanchego/utils/logging"
 
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/maps"
@@ -52,9 +51,8 @@ func list(_ *cobra.Command, _ []string) error {
 		for _, cloudID := range clusterConf.GetCloudIDs() {
 			nodeIDStr := "----------------------------------------"
 			if clusterConf.IsAvalancheGoHost(cloudID) {
-				nodeID, err := getNodeID(app.GetNodeInstanceDirPath(cloudID))
-				if err != nil {
-					ux.Logger.PrintToUser(logging.Yellow.Wrap("could not obtain node ID for nodes %s: %s"), cloudID, err)
+				if nodeID, err := getNodeID(app.GetNodeInstanceDirPath(cloudID)); err != nil {
+					ux.Logger.RedXToUser("could not obtain node ID for nodes %s: %s", cloudID, err)
 				} else {
 					nodeIDStr = nodeID.String()
 				}
--- a/cmd/nodecmd/list.go
+++ b/cmd/nodecmd/list.go
@@ -3,12 +3,12 @@
 package nodecmd
 
 import (
-	"fmt"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


cmd/nodecmd/list.go
```diff
@@ -45,29 +45,34 @@ func list(_ *cobra.Command, _ []string) error {
...
-			nodeConfig, err := app.LoadClusterNodeConfig(cloudID)
-			if err != nil {
-				return err
-			}
 			nodeIDStr := "----------------------------------------"
 			if clusterConf.IsAvalancheGoHost(cloudID) {
 				nodeID, err := getNodeID(app.GetNodeInstanceDirPath(cloudID))
 				if err != nil {
-					return err
+					ux.Logger.PrintToUser(logging.Yellow.Wrap("could not obtain node ID for nodes %s: %s"), cloudID, err)
```
arturrez, line 57: nit: use RedXToUser instead

cmd/nodecmd/list.go
```diff
@@ -45,29 +45,34 @@ func list(_ *cobra.Command, _ []string) error {
...
 		}
+		nodeIDs := []string{}
 		for _, cloudID := range clusterConf.GetCloudIDs() {
-			nodeConfig, err := app.LoadClusterNodeConfig(cloudID)
-			if err != nil {
-				return err
-			}
 			nodeIDStr := "----------------------------------------"
 			if clusterConf.IsAvalancheGoHost(cloudID) {
 				nodeID, err := getNodeID(app.GetNodeInstanceDirPath(cloudID))
```
arturrez, line 55: can be shortened with if nodeID, err:=..., err!=nil; else nodeIDStr = nodeID.String()

cmd/nodecmd/list.go
```diff
@@ -45,29 +45,34 @@ func list(_ *cobra.Command, _ []string) error {
...
 				if err != nil {
-					return err
+					ux.Logger.PrintToUser(logging.Yellow.Wrap("could not obtain node ID for nodes %s: %s"), cloudID, err)
+				} else {
+					nodeIDStr = nodeID.String()
```
arturrez, line 59: simpler `nodeIDs = append(nodeIDs, nodeID.String())` here and no need `nodeIDs = append(nodeIDs, nodeIDStr)` later

cmd/nodecmd/list.go
```diff
@@ -45,29 +45,34 @@ func list(_ *cobra.Command, _ []string) error {
...
+		ux.Logger.PrintToUser("Cluster %q (%s)", clusterName, clusterConf.Network.Kind.String())
+		for i, cloudID := range clusterConf.GetCloudIDs() {
+			nodeConfig, err := app.LoadClusterNodeConfig(cloudID)
+			if err != nil {
+				return err
 			}
 			roles := clusterConf.GetHostRoles(nodeConfig)
 			rolesStr := strings.Join(roles, ",")
 			if rolesStr != "" {
 				rolesStr = " [" + rolesStr + "]"
```
arturrez, line 73: nit: use fmt.Sprintf

arturrez (approved):

cmd/nodecmd/list.go
felipemadero: it is needed so, for the default case

cmd/nodecmd/list.go
felipemadero: I believe it is ok as it

PR #1695, opened 2024-03-28
Fix Load Test Stop issues

sukantoraymond:


cmd/nodecmd/destroy.go
```diff
@@ -120,10 +115,18 @@ func destroyNodes(_ *cobra.Command, args []string) error {
 		nodesToStop = append(nodesToStop, monitoringNode)
 	}
 	// stop all load test nodes if specified
-	ltHostsToStop, err := getClusterLoadTestNodes(clusterName)
+	ltHosts, err := getLoadTestInstancesInCluster(clusterName)
```
patrick-ogrady, line 118: nit: it seems like this could just return the instances? am I missing something?

cmd/nodecmd/destroy.go
```diff
@@ -120,10 +115,18 @@ func destroyNodes(_ *cobra.Command, args []string) error {
...
 	}
+	ltHostsToStop := []string{}
+	for _, loadTestName := range ltHosts {
+		ltInstance, err := getExistingLoadTestInstance(clusterName, loadTestName)
+		if err != nil {
+			return err
+		}
+		ltHostsToStop = append(ltHostsToStop, ltInstance)
+	}
 	nodesToStop = append(nodesToStop, ltHostsToStop...)
```
patrick-ogrady, line 130: nit: why not just add to `nodesToStop` in the loop above?

patrick-ogrady (approved):

PR #1701, opened 2024-03-29
robust loadtest

sukantoraymond:


pkg/ssh/shell/buildLoadTestDeps.sh
```diff
@@ -1,7 +1,21 @@
 #!/usr/bin/env bash
 # install gcc
 echo "ensuring that gcc is installed ..."
-gcc --version || DEBIAN_FRONTEND=noninteractive sudo apt-get -y -o DPkg::Lock::Timeout=120 install gcc
+while true; do
```
patrick-ogrady, line 4: Found a simpler variant:
```
while ! gcc --version >/dev/null 2>&1; do
    echo "GCC is not installed. Trying to install..."
    DEBIAN_FRONTEND=noninteractive sudo apt-get -y -o DPkg::Lock::Timeout=120 install gcc
    if [ $? -ne 0 ]; then
        echo "Failed to install GCC. Retrying in 10 seconds..."
        sleep 10
    fi
done
```

patrick-ogrady (approved):

PR #1713, opened 2024-04-03
Destroy improvements

sukantoraymond:


cmd/nodecmd/loadtestStop.go
```diff
@@ -94,6 +94,37 @@ func stopLoadTest(_ *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+	cloudSecurityGroupList, err := getCloudSecurityGroupList(clusterNodes)
+	if err != nil {
+		return err
+	}
+	filteredSGList := utils.Filter(cloudSecurityGroupList, func(sg regionSecurityGroup) bool { return sg.cloud == nodeToStopConfig.CloudService })
+	if len(filteredSGList) == 0 {
+		return fmt.Errorf("no endpoint found in the  %s", nodeToStopConfig.CloudService)
```
felipemadero, line 118: pls improve this error message

cmd/nodecmd/loadtestStop.go
```diff
@@ -94,6 +94,37 @@ func stopLoadTest(_ *cobra.Command, args []string) error {
...
+	}
+	ec2SvcMap := make(map[string]*awsAPI.AwsCloud)
+	for _, sg := range filteredSGList {
+		sgEc2Svc, err := awsAPI.NewAwsCloud(awsProfile, sg.region)
```
felipemadero, line 122: can you first check if the region is already present in the map?

cmd/nodecmd/loadtestStop.go
```diff
@@ -129,6 +160,12 @@ func stopLoadTest(_ *cobra.Command, args []string) error {
 			if err = destroyNode(existingSeparateInstance, clusterName, loadTestName, loadTestEc2SvcMap[separateHostRegion], nil); err != nil {
 				return err
 			}
+			for _, sg := range filteredSGList {
+				if err = deleteMonitoringSecurityGroupRule(ec2SvcMap[sg.region], loadTestNodeConfig.PublicIPs[0], sg.securityGroup); err != nil {
```
felipemadero, line 164: can we rename the function name? it seems it is more general than just monitoring sg (load test in this case). also change naming inside the function, like the param monirotingHostPublicIP

cmd/nodecmd/destroy.go
```diff
@@ -126,14 +128,29 @@ func destroyNodes(_ *cobra.Command, args []string) error {
...
+	}
+	// TODO: will need to change this logic if we decide to mix AWS and GCP instances in a cluster
+	filteredSGList := utils.Filter(cloudSecurityGroupList, func(sg regionSecurityGroup) bool { return sg.cloud == nodeToStopConfig.CloudService })
+	if len(filteredSGList) == 0 {
+		return fmt.Errorf("no endpoint found in the  %s", nodeToStopConfig.CloudService)
+	}
 	var gcpCloud *gcpAPI.GcpCloud
+	ec2SvcMap := make(map[string]*awsAPI.AwsCloud)
+	for _, sg := range filteredSGList {
+		sgEc2Svc, err := awsAPI.NewAwsCloud(awsProfile, sg.region)
```
felipemadero, line 148: this assumes sg in filteredSGList are aws, so not so general (what happens if cluster is non AWS?)

cmd/nodecmd/loadtestStop.go
felipemadero: getCloudSecurityGroupList checks for diff security group name also, so there is a chance that a region is repeated

felipemadero (approved):

PR #1716, opened 2024-04-03
[scripts] Build from Source

patrick-ogrady:


cmd/nodecmd/wiz.go
```diff
@@ -172,7 +172,9 @@ func wiz(cmd *cobra.Command, args []string) error {
 
 	if !clusterAlreadyExists {
 		globalNetworkFlags.UseDevnet = true
-		useAvalanchegoVersionFromSubnet = subnetName
+		if len(useCustomAvalanchegoVersion) == 0 {
```
felipemadero, line 175: also check for useLatestAvalanchegoReleaseVersion, useLatestAvalanchegoPreReleaseVersion

pkg/ssh/shell/setupBuildEnv.sh
```diff
@@ -1,29 +1,8 @@
...
-  cd ~
-  sudo rm -rf $GOFILE go
-  wget -q -nv https://go.dev/dl/$GOFILE
-  tar xfz $GOFILE
-  echo >> ~/.bashrc
-  echo export PATH=\$PATH:~/go/bin:~/bin >> ~/.bashrc
-  echo export CGO_ENABLED=1 >> ~/.bashrc
-}
-go version || install_go
+go version || sudo snap install go --classic
```
felipemadero, line 8: 1. we are loosing control of golang version here (.GoVersion). Do we want that? we probably can use --channel=... here
2. It is posible for snap latest version to become obsolete regarding vms/avago compilation needs? in this case a tar give best control.
3. Besides that, probably it is a bit more custom to use apt instead of snap.

PR #1775, opened 2024-04-16
rm sed as external dependency for custom grafana dashboard

arturrez:


pkg/ssh/ssh.go
```diff
@@ -264,7 +276,7 @@ func RunSSHUpdateMonitoringDashboards(host *models.Host, monitoringDashboardPath
 		if err := utils.FileCopy(utils.ExpandHome(customGrafanaDashboardPath), filepath.Join(monitoringDashboardPath, constants.CustomGrafanaDashboardJSON)); err != nil {
 			return err
 		}
-		if err := replaceCustomVarDashboardValues(monitoringDashboardPath, constants.CustomGrafanaDashboardJSON, chainID); err != nil {
+		if err := replaceCustomVarDashboardValues(constants.CustomGrafanaDashboardJSON, chainID); err != nil {
```
felipemadero, line 279: why monitoringDashboardPath is not needed now?

felipemadero: the file seems to be `filepath.Join(monitoringDashboardPath, constants.CustomGrafanaDashboardJSON)` probably it is best to now assume current working dir

felipemadero (commented):
couple comments then will approve

pkg/ssh/ssh.go
arturrez: we don't need working dir anymore

arturrez: but you are correct.
Fixed

arturrez:
seems to be working 
image

felipemadero (approved):
LGTM

PR #1811, opened 2024-04-29
Fix config for relayer new version 1.2.0

bmastc:
AWM Relayer version 1.2.0 was just released today.

It includes a change in config file which is making the relayer process launched by CLI panic on start.

This commit bumps the dependency on `awm-relayer` and fixes the config file creation function.


felipemadero (approved):
LGTM. Thanks!

PR #1846, opened 2024-05-07
Fix metrics

sukantoraymond:


felipemadero (commented):
why not asking the user if doing metrics collection anymore? Maybe the issue can be solved in some other way? if the decision is to avoid asking the user like this, you also should remove the metrics setting from the config command

felipemadero (approved):

PR #1850, opened 2024-05-07
Node Telemetry

sukantoraymond:


pkg/constants/constants.go
```diff
@@ -262,6 +262,13 @@ const (
 	AVMKeyName                 = "avm"
 	EVMKeyName                 = "evm"
 	PlatformKeyName            = "platform"
+	CloudService               = "cloud-service"
```
felipemadero, line 265: can we add a prefix to discriminate this are metrics item constants only?

cmd/nodecmd/create.go
```diff
@@ -1488,6 +1498,29 @@ func defaultAvalancheCLIPrefix(region string) (string, error) {
...
 }
 
+func sendMetrics(cmd *cobra.Command, cloudService, network string, nodes map[string]NumNodes) {
+	flags := make(map[string]string)
+	totalValidatorNodes := 0
+	totalAPINodes := 0
+	for region := range nodes {
+		totalValidatorNodes += nodes[region].numValidators
+		totalAPINodes += nodes[region].numAPI
+		flags[region] = strconv.Itoa(nodes[region].numValidators)
```
felipemadero, line 1508: should you also add a constant taken prefix to discriminate this is a metrics region item?

felipemadero, line 1508: why not the total sum of nodes in the region?

felipemadero (approved):

PR #1960, opened 2024-06-12
use avalanche-tooling-sdk-go for node provision

arturrez:
ref: https://github.com/ava-labs/avalanche-cli/issues/1944


arturrez:
closing this one as it's not needed anymore. sdk scope has been changed and this is not part of it anymore

PR #2026, opened 2024-07-12
remove remote cli dependency for CLI

arturrez:
sdk PR to follow
ref: https://github.com/ava-labs/avalanche-tooling-sdk-go/issues/76

### DIFF
--- a/cmd/nodecmd/sync.go
+++ b/cmd/nodecmd/sync.go
@@ -97,14 +97,6 @@ func trackSubnet(
 		go func(nodeResults *models.NodeResults, host *models.Host) {
 			defer wg.Done()
 			subnetExportPath := filepath.Join("/tmp", filepath.Base(subnetPath))
-			if err := ssh.RunSSHExportSubnet(host, subnetPath, subnetExportPath); err != nil {
-				nodeResults.AddResult(host.NodeID, nil, err)
-				return
-			}
-			if err := ssh.RunSSHUploadClustersConfig(host, app.GetClustersConfigPath()); err != nil {
-				nodeResults.AddResult(host.NodeID, nil, err)
-			}
-
 			if err := ssh.RunSSHTrackSubnet(host, subnetName, subnetExportPath, networkFlag); err != nil {
 				nodeResults.AddResult(host.NodeID, nil, err)
 				return
--- a/cmd/nodecmd/sync.go
+++ b/cmd/nodecmd/sync.go
@@ -9,9 +9,7 @@ import (
 	"github.com/ava-labs/avalanche-cli/cmd/subnetcmd"
 	"github.com/ava-labs/avalanche-cli/pkg/ansible"
 	"github.com/ava-labs/avalanche-cli/pkg/cobrautils"
-	"github.com/ava-labs/avalanche-cli/pkg/constants"
 	"github.com/ava-labs/avalanche-cli/pkg/models"
-	"github.com/ava-labs/avalanche-cli/pkg/plugins"
 	"github.com/ava-labs/avalanche-cli/pkg/ssh"
 	"github.com/ava-labs/avalanche-cli/pkg/utils"
 	"github.com/ava-labs/avalanche-cli/pkg/ux"
@@ -86,13 +84,9 @@ func syncSubnet(_ *cobra.Command, args []string) error {
 	return nil
 }
 
+// prepareSubnetPlugin creates subnet plugin to all nodes in the cluster
 func prepareSubnetPlugin(hosts []*models.Host, subnetName string) error {
-	pluginDir := app.GetTmpPluginDir()
 	sc, err := app.LoadSidecar(subnetName)
-	if err != nil {
-		return err
-	}
-	vmPath, err := plugins.CreatePlugin(app, sc.Name, pluginDir)
 	if err != nil {
 		return err
 	}
@@ -102,7 +96,7 @@ func prepareSubnetPlugin(hosts []*models.Host, subnetName string) error {
 		wg.Add(1)
 		go func(nodeResults *models.NodeResults, host *models.Host) {
 			defer wg.Done()
-			if err := host.Upload(vmPath, constants.CloudNodePluginsPath, constants.SSHFileOpsTimeout); err != nil {
+			if err := ssh.RunSSHCreatePlugin(host, sc); err != nil {
 				nodeResults.AddResult(host.NodeID, nil, err)
 			}
 		}(&wgResults, host)
--- a/cmd/nodecmd/sync.go
+++ b/cmd/nodecmd/sync.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ava-labs/avalanche-cli/pkg/cobrautils"
 	"github.com/ava-labs/avalanche-cli/pkg/models"
 	"github.com/ava-labs/avalanche-cli/pkg/ssh"
+	"github.com/ava-labs/avalanche-cli/pkg/utils"
 	"github.com/ava-labs/avalanche-cli/pkg/ux"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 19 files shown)


cmd/nodecmd/wiz.go
```diff
@@ -73,6 +73,7 @@ var (
 	deployTeleporterMessenger              bool
 	deployTeleporterRegistry               bool
 	replaceKeyPair                         bool
+	depricatedRemoteCLIVersion             string
```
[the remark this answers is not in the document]
arturrez: it would force us to update hypersdk example for morpheusVM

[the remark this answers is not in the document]
arturrez: k. will do

cmd/nodecmd/sync.go
```diff
@@ -78,34 +84,63 @@ func syncSubnet(_ *cobra.Command, args []string) error {
 	return nil
 }
 
+// prepareSubnetPlugin creates subnet plugin to all nodes in the cluster
+func prepareSubnetPlugin(hosts []*models.Host, subnetName string) error {
+	sc, err := app.LoadSidecar(subnetName)
+	if err != nil {
+		return err
+	}
+	ux.Logger.Info("Creating subnet %s plugin for hosts: %s", subnetName, hosts)
```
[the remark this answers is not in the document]
arturrez: fixed

pkg/ssh/installer.go
```diff
@@ -0,0 +1,30 @@
...
+type HostInstaller struct {
+	Host *models.Host
+}
+
+func NewHostInstaller(host *models.Host) *HostInstaller {
+	return &HostInstaller{Host: host}
+}
+
+func (h *HostInstaller) GetArch() (string, string) {
+	goArhBytes, err := h.Host.Command("go env GOARCH", nil, constants.SSHScriptTimeout)
```
felipemadero, line 21: I wonder if we can avoid depending on go binary for this

pkg/ssh/ssh.go
```diff
@@ -502,42 +494,253 @@ func RunSSHUploadStakingFiles(host *models.Host, nodeInstanceDirPath string) err
...
+	// make sure that genesis and bootstrap data is preserved
+	if genesisFileExists(host) {
+		avagoConf.GenesisPath = filepath.Join(constants.DockerNodeConfigPath, constants.GenesisFileName)
+	}
+	remoteAvagoConfFile, err := getAvalancheGoConfigData(host)
+	if err != nil {
 		return err
 	}
-	if _, err := host.Command(fmt.Sprintf("/home/ubuntu/bin/avalanche subnet join %s %s --avalanchego-config /home/ubuntu/.avalanchego/configs/node.json --plugin-dir /home/ubuntu/.avalanchego/plugins --force-write", subnetName, networkFlag), nil, constants.SSHScriptTimeout); err != nil {
+	bootstrapIDs, _ := utils.GetValueString(remoteAvagoConfFile, "bootstrap-ids")
```
felipemadero, line 527: we should check for error here

pkg/ssh/ssh.go
```diff
@@ -502,42 +494,253 @@ func RunSSHUploadStakingFiles(host *models.Host, nodeInstanceDirPath string) err
...
+	bootstrapIPs, _ := utils.GetValueString(remoteAvagoConfFile, "bootstrap-ips")
```
felipemadero, line 528: same

pkg/ssh/ssh.go
```diff
@@ -502,42 +494,253 @@ func RunSSHUploadStakingFiles(host *models.Host, nodeInstanceDirPath string) err
...
+		installURL, _, err := dl.GetDownloadURL(sc.VMVersion, hostInstaller) // extension is tar.gz
+		if err != nil {
+			return err
+		}
+
+		archiveName := "subnet-evm.tar.gz"
+		archiveFullPath := filepath.Join(tmpDir, archiveName)
+
+		// download and install subnet evm
+		if _, err := host.Command(fmt.Sprintf("%s %s -O %s", "busybox wget", installURL, archiveFullPath), nil, constants.SSHLongRunningScriptTimeout); err != nil {
```
felipemadero, line 589: I believe you have scripts for this on shell, eg upgradeSubnetEVM, pkg/ssh/shell/getNewSubnetEVMRelease.sh, why not using them?

felipemadero: why not having one .sh file for all of this?

pkg/ssh/ssh.go
```diff
@@ -502,42 +494,253 @@ func RunSSHUploadStakingFiles(host *models.Host, nodeInstanceDirPath string) err
...
+		}
+		defer os.Remove(subnetConfigFile.Name())
+		if err := os.WriteFile(subnetConfigFile.Name(), subnetConfig, constants.WriteReadUserOnlyPerms); err != nil {
+			return err
+		}
+		subnetConfigPath := filepath.Join(constants.CloudNodeConfigPath, "subnets", subnetIDStr+".json")
+		if err := host.MkdirAll(filepath.Dir(subnetConfigPath), constants.SSHDirOpsTimeout); err != nil {
+			return err
+		}
+		if err := host.Upload(subnetConfigFile.Name(), subnetConfigPath, constants.SSHFileOpsTimeout); err != nil {
```
felipemadero, line 689: can we have an upload that takes as input a slice of bytes, so we dont need to create tmp files?

pkg/ssh/ssh.go
```diff
@@ -702,3 +905,31 @@ func composeFileExists(host *models.Host) bool {
...
+
+func getAvalancheGoConfigData(host *models.Host) (map[string]interface{}, error) {
+	// get remote node.json file
+	nodeJSONPath := filepath.Join(constants.CloudNodeConfigPath, constants.NodeFileName)
+	tmpFile, err := os.CreateTemp("", "avalanchecli-node-*.json")
+	if err != nil {
+		return nil, err
+	}
+	defer os.Remove(tmpFile.Name())
+	if err := host.Download(nodeJSONPath, tmpFile.Name(), constants.SSHFileOpsTimeout); err != nil {
```
felipemadero, line 922: can we have a download that returns a byte slice?

pkg/utils/common.go
```diff
@@ -559,3 +559,11 @@ func Command(cmdLine string, params ...string) *exec.Cmd {
 	c.Env = os.Environ()
 	return c
 }
+
+// GetValueString returns the value of a key in a map as a string.
+func GetValueString(data map[string]interface{}, key string) (string, error) {
```
felipemadero, line 564: can we improve the name of the function?

pkg/utils/json.go
```diff
@@ -25,3 +25,14 @@ func ValidateJSON(path string) ([]byte, error) {
 
 	return contentBytes, nil
 }
+
+// MergeJsonMaps merges two maps of type map[string]interface{}
+func MergeJSONMaps(a, b map[string]interface{}) map[string]interface{} {
```
felipemadero, line 30: can be maps.Copy be used instead of MergeJSONMaps?

pkg/ssh/installer.go
arturrez: good point. agree

pkg/ssh/ssh.go
arturrez: agree. we should have another func for this

pkg/ssh/ssh.go
arturrez: agree. we should have another func for this

pkg/utils/common.go
arturrez: which one do you suggest ?

pkg/ssh/ssh.go
arturrez: it will increase size of `scriptInputs` struct so I though we can get away without it as operation is straightforward

PR #2034, opened 2024-07-16
Cleanup

sukantoraymond:
To clean all previously created instances : 
`avalanche node destroy --all`

otherwise `avalanche node destroy clustername`

This PR also makes `avalanche node destroy --all` silence all logs from destroy 

In addition, it doesn't consider "node config not found" to be an error because if node config is not found, most likely it is already deleted on console but not removed from cluster list in cluster_config.json

GCP still missing. This PR is mostly for AWS

### DIFF
--- a/cmd/nodecmd/destroy.go
+++ b/cmd/nodecmd/destroy.go
@@ -105,7 +105,8 @@ func CallDestroyNode(clusterName string) error {
 // This is because some nodes might have had their node_config.json file deleted as part of
 // deletion process but if an error occurs during deletion process, the node might still exist
 // as part of the cluster in cluster_config.json
-// getFirstAvailableNode
+// If all nodes in the cluster no longer have their node_config.json files, getFirstAvailableNode
+// will return false in its second return value
 func getFirstAvailableNode(nodesToStop []string) (string, bool) {
 	firstAvailableNode := nodesToStop[0]
 	noAvailableNodesFound := false


cmd/nodecmd/create_aws.go
```diff
@@ -523,7 +523,9 @@ func createAWSInstances(
 		if len(failedNodes) > 0 {
 			ux.Logger.PrintToUser("Failed nodes: ")
 			for node, err := range failedNodes {
-				ux.Logger.PrintToUser(fmt.Sprintf("Failed to destroy node %s due to %s", node, err))
+				if enableLogs {
```
arturrez, line 526: i feel like we should still display logs so it will be up to the user to still investigate.

cmd/nodecmd/destroy.go
```diff
@@ -24,6 +24,8 @@ import (
 var (
 	authorizeRemove bool
 	authorizeAll    bool
+	enableLogs      bool
```
arturrez, line 27: see above. not needed

cmd/nodecmd/destroy.go
```diff
@@ -93,12 +97,62 @@ func removeClustersConfigFiles(clusterName string) error {
 	return removeNodeFromClustersConfig(clusterName)
 }
 
-func CallDestroyNode(clusterName string) error {
+func CallDestroyNode(clusterName string, logs bool) error {
```
arturrez, line 100: see above

cmd/nodecmd/destroy.go
```diff
@@ -93,12 +97,62 @@ func removeClustersConfigFiles(clusterName string) error {
...
 	authorizeAll = true
+	enableLogs = logs
```
arturrez, line 102: see above

pkg/cloud/aws/aws.go
```diff
@@ -347,10 +347,15 @@ func (c *AwsCloud) checkInstanceIsRunning(nodeID string) (bool, error) {
 }
 
 // DestroyAWSNode terminates an EC2 instance with the given ID.
-func (c *AwsCloud) DestroyAWSNode(nodeConfig models.NodeConfig, clusterName string) error {
+func (c *AwsCloud) DestroyAWSNode(nodeConfig models.NodeConfig, clusterName string, enableLogs bool) error {
```
arturrez, line 350: same see above. we need logs to my opinion

cmd/nodecmd/whitelist.go
```diff
@@ -286,6 +288,12 @@ func whitelistSSHPubKey(clusterName string, pubkey string) error {
 func getCloudSecurityGroupList(clusterNodes []string) ([]regionSecurityGroup, error) {
 	cloudSecurityGroupList := []regionSecurityGroup{}
 	for _, node := range clusterNodes {
+		_, err := os.Stat(app.GetNodeConfigPath(node))
```
arturrez, line 291: utils.FileExists can be used

cmd/nodecmd/destroy.go
```diff
@@ -165,21 +223,31 @@ func destroyNodes(_ *cobra.Command, args []string) error {
 	}
 	for _, node := range nodesToStop {
 		if !isExternalCluster {
+			// if we can't find node config path, that means node already deleted on console
+			// but we didn't get to delete the node from cluster config file
+			_, err := os.Stat(app.GetNodeConfigPath(node))
```
arturrez, line 228: use utils.FileExists

cmd/nodecmd/destroy.go
```diff
@@ -93,12 +97,62 @@ func removeClustersConfigFiles(clusterName string) error {
...
+			if errors.Is(err, os.ErrNotExist) {
+				continue
+			}
+		}
+		firstAvailableNode = node
+	}
+	return firstAvailableNode, noAvailableNodesFound
+}
+
+func Cleanup() error {
```
arturrez, line 125: i think logic for destroy can be improved by loading all clusters and then filtering this list by cluster name provided by users or using this list for `--all` case

cmd/nodecmd/destroy.go
```diff
@@ -93,12 +97,62 @@ func removeClustersConfigFiles(clusterName string) error {
...
 }
 
-func CallDestroyNode(clusterName string) error {
+func CallDestroyNode(clusterName string, logs bool) error {
 	authorizeAll = true
+	enableLogs = logs
 	return destroyNodes(nil, []string{clusterName})
 }
 
+func getFirstAvailableNode(nodesToStop []string) (string, bool) {
```
arturrez, line 106: not sure why we even need this

cmd/nodecmd/destroy.go
arturrez: to make one code to work in all cases

cmd/nodecmd/wiz.go
```diff
@@ -135,6 +136,7 @@ The node wiz command creates a devnet and deploys, sync and validate a subnet in
 	cmd.Flags().IntVar(&throughput, "aws-volume-throughput", constants.AWSGP3DefaultThroughput, "AWS throughput in MiB/s (for gp3 volume type only)")
 	cmd.Flags().StringVar(&volumeType, "aws-volume-type", "gp3", "AWS volume type")
 	cmd.Flags().IntVar(&volumeSize, "aws-volume-size", constants.CloudServerStorageSize, "AWS volume size in GB")
+	cmd.Flags().BoolVar(&cleanupNodes, "cleanup", false, "clean up all existing instances created by Avalanche CLI")
```
arturrez, line 139: I am not sure what it means in the scope of the `wiz`. delete all clusters registered  with avalanche-cli in case of `wiz` command failure?

cmd/nodecmd/destroy.go
```diff
@@ -98,7 +100,56 @@ func CallDestroyNode(clusterName string) error {
...
+			}
+		}
+	}
+	ux.Logger.PrintToUser("all existing instances created by Avalanche CLI successfully destroyed")
+	return nil
+}
+
 func destroyNodes(_ *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		if !destroyAll {
```
arturrez, line 148: nit: reverse logic `if destroyAll -> Cleanup else return err`

cmd/nodecmd/destroy.go
```diff
@@ -98,7 +100,56 @@ func CallDestroyNode(clusterName string) error {
...
 }
 
+func getFirstAvailableNode(nodesToStop []string) (string, bool) {
+	firstAvailableNode := nodesToStop[0]
+	noAvailableNodesFound := false
+	for index, node := range nodesToStop {
+		nodeConfigPath := app.GetNodeConfigPath(node)
+		_, err := os.Stat(nodeConfigPath)
+		if err != nil {
+			if index == len(nodesToStop)-1 {
```
arturrez, line 110: can you pls add a comment with explanation of the logic

arturrez (approved):

PR #2047, opened 2024-07-22
address felipe feedback

arturrez:


pkg/utils/json.go
```diff
@@ -28,11 +29,6 @@ func ValidateJSON(path string) ([]byte, error) {
 
 // MergeJsonMaps merges two maps of type map[string]interface{}
 func MergeJSONMaps(a, b map[string]interface{}) map[string]interface{} {
```
felipemadero, line 31: do you believe it may be best to just remove the function, and use maps.Copy directly in pkg/ssh/ssh.go instead of MergeJSONmaps?

pkg/models/host.go
```diff
@@ -148,6 +164,19 @@ func (h *Host) Download(remoteFile string, localFile string, timeout time.Durati
 	return err
 }
 
+// ReadFileBytes downloads a file from the remote server to a byte array
+func (h *Host) ReadFileBytes(remoteFile string, timeout time.Duration) ([]byte, error) {
```
felipemadero, line 168: ohh I see https://github.com/melbahja/goph/blob/master/client.go does not support uploading or downloading bytes. such a shame. maybe we can propose a PR to goph.

felipemadero (approved):

pkg/utils/json.go
arturrez: I still feel like it makes sense to keep a separate func

arturrez: it makes code more readable imho

pkg/models/host.go
arturrez: agree

PR #2055, opened 2024-07-24
Allowlist: Add red x to error message

siljapetasch:
When user wants to add an address to the allowlist that already has a role, an error message is printed. This one is hard to see, so I suggest adding a red X to the message

### DIFF
--- a/pkg/vm/allowlist.go
+++ b/pkg/vm/allowlist.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/ava-labs/avalanche-cli/pkg/application"
 	"github.com/ava-labs/avalanche-cli/pkg/utils"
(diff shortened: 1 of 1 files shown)


arturrez:
please merge latest main and resolve conflict
once done I should be able to approve it @siljapetasch  //cc @felipemadero

siljapetasch:
@arturrez merged recent changes as requested

felipemadero:
@siljapetasch hi, thanks for the collaboration. we need all commits to be signed. once that is done, we will merge

arturrez:
please amend your commit to add signature to it following https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

meaghanfitzgerald:
@siljapetasch bumping this

PR #2137, opened 2024-09-08
SDK refactor from tooling sdk repo

sukantoraymond:


pkg/vm/create_evm.go
```diff
@@ -121,29 +113,33 @@ func CreateEVMGenesis(
 		)
 	}
 
-	getPrecompiles(conf, params, &genesis.Timestamp)
+	precompiles := getPrecompiles(params, utils.TimeToNewUint64(time.Now()))
```
felipemadero, line 116: can we use a variable for genesis block 0 timestamp and set it to utils.TimeToNewUint64(time.Now()),
then pass it to getPrecompiles and also as a new field of subnetSDK.SubnetParams so all timestamps can
be specified (and in this case, genesis block 0 timestamp to equal precompile timestamps)

pkg/tooling-sdk/subnet/subnet.go
```diff
@@ -0,0 +1,305 @@
...
+
+func (c *Subnet) SetSubnetID(subnetID ids.ID) {
+	c.SubnetID = subnetID
+}
+
+func createEvmGenesis(
+	subnetEVMParams *SubnetEVMParams,
+) ([]byte, error) {
+	genesis := core.Genesis{}
+	genesis.Timestamp = *utils.TimeToNewUint64(time.Now())
```
felipemadero, line 199: can we set it from a subnetEVMParams field?

felipemadero (commented):
two comments:

1) can we use top level `sdk/` instead of `pkg/tooling-sdk/`. I believe it is best to deeply differenciate sdk from pkg, both for external users and for CLI internal usage

2) probably also rename `sdk/subnet/` to `sdk/blockchain/`

3) I aggre with changes on `pkg/vm/` and added a comment on `sdk/blockchain/`, but I see the dependencies grow. The problem is that you are not introducing replacements for all this packages. so we are having duplicate code for many of them. I believe the idea was to avoid duplicate code as a requirements. so, why not we start from bottom up but making PRs for the most basic ones, and replacing whenever needed? and them move up?

arturrez:
it seems like this is full sdk code without CLI refactoring to use it. 
Wondering if we should also split this into separate pieces(PRs) with CLI refactoring using it.

arturrez (approved):
please fix lint

PR #2138, opened 2024-09-08
Subnet sdk refactor

sukantoraymond:


pkg/tooling-sdk/avalanchesdk/network.go
```diff
@@ -28,18 +21,6 @@ const (
 	MainnetAPIEndpoint = "https://api.avax.network"
 )
 
-func (nk NetworkKind) String() string {
```
arturrez, line 31: why do we remove these functions?

arturrez (approved):
why do we remove some functions that seems useful from sdk ?

PR #2147, opened 2024-09-10
separate home and remote decimals when deploying erc20 remote

felipemadero:


cmd/interchaincmd/tokentransferrercmd/deploy.go
```diff
@@ -578,12 +577,11 @@ func CallDeploy(_ []string, flags DeployFlags) error {
 		remoteSupply  *big.Int
 	)
 
-	tokenRemoteDecimals := tokenHomeDecimals
-	if flags.remoteFlags.tokenDecimals != 0 {
-		tokenRemoteDecimals = flags.remoteFlags.tokenDecimals
-	}
-
 	if !flags.remoteFlags.native {
+		tokenRemoteDecimals := tokenHomeDecimals
```
michaelkaplan13, line 581: So for ERC20TokenRemote, we default the `tokenDecimals` to be the same as the `tokenHomeDecimals`, but allow this to be overridden by a user provided flag? This seems perfectly reasonable to me, could we just add a comment here documenting the behavior?

arturrez (approved):

michaelkaplan13 (approved):
Makes sense to me.

Just want to confirm, it is always the case that either:
- `flags.homeFlags.homeAddress != ""`
- `flags.homeFlags.erc20Address != ""`
- `flags.homeFlags.native`

Otherwise, it appears the `tokenHomeSymbol`, `tokenHomeName`, and `tokenHomeDecimals` are never properly set.

felipemadero:
> Makes sense to me.
> 
> Just want to confirm, it is always the case that either:
> 
> * `flags.homeFlags.homeAddress != ""`
> * `flags.homeFlags.erc20Address != ""`
> * `flags.homeFlags.native`
> 
> Otherwise, it appears the `tokenHomeSymbol`, `tokenHomeName`, and `tokenHomeDecimals` are never properly set.

Yes, we check for this on https://github.com/ava-labs/avalanche-cli/blob/c8cfe5ebfdf9132021a804463d211a9dde94da56/cmd/interchaincmd/tokentransferrercmd/deploy.go#L129

michaelkaplan13:
> > Makes sense to me.
> > Just want to confirm, it is always the case that either:
> > 
> > * `flags.homeFlags.homeAddress != ""`
> > * `flags.homeFlags.erc20Address != ""`
> > * `flags.homeFlags.native`
> > 
> > Otherwise, it appears the `tokenHomeSymbol`, `tokenHomeName`, and `tokenHomeDecimals` are never properly set.
> 
> Yes, we check for this on
> 
> https://github.com/ava-labs/avalanche-cli/blob/c8cfe5ebfdf9132021a804463d211a9dde94da56/cmd/interchaincmd/tokentransferrercmd/deploy.go#L129

It looks like that function checks that no more than 1 of the conditions is true, but doesn't ensure that at least 1 of them are true. It may not be a big issue either way though.

PR #2150, opened 2024-09-11
Prompts for Subnet create acp77

sukantoraymond:
Prompts for get new native token minter and validatormanager contract controller

### DIFF
--- a/cmd/blockchaincmd/deploy.go
+++ b/cmd/blockchaincmd/deploy.go
@@ -6,10 +6,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/ethereum/go-ethereum/common"
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/ethereum/go-ethereum/common"
 
 	"github.com/ava-labs/avalanche-cli/pkg/binutils"
 	"github.com/ava-labs/avalanche-cli/pkg/cobrautils"


cmd/blockchaincmd/create.go
```diff
@@ -63,6 +66,8 @@ var (
 	errMutuallyExlusiveVersionOptions             = errors.New("version flags --latest,--pre-release,vm-version are mutually exclusive")
 	errMutuallyExclusiveVMConfigOptions           = errors.New("--genesis flag disables --evm-chain-id,--evm-defaults,--production-defaults,--test-defaults")
 	errMutuallyExlusiveValidatorManagementOptions = errors.New("validator management type flags --proof-of-authority,--proof-of-stake are mutually exclusive")
+	errTokenMinterAddressConflict                 = errors.New("--validator-manager-mint-only means that no additional addresses can be provided in --token-minter-address")
```
felipemadero, line 69: we can just leave the current precompile prompts for native minter

felipemadero: I mean, no flags.

felipemadero: if wanting to add flags for precompile settings, let do a separate PR for that

cmd/blockchaincmd/create.go
```diff
@@ -108,6 +113,9 @@ configuration, pass the -f flag.`,
 	cmd.Flags().BoolVar(&createFlags.useExternalGasToken, "external-gas-token", false, "use a gas token from another blockchain")
 	cmd.Flags().BoolVar(&createFlags.proofOfAuthority, "proof-of-authority", false, "use proof of authority for validator management")
 	cmd.Flags().BoolVar(&createFlags.proofOfStake, "proof-of-stake", false, "use proof of stake for validator management")
+	cmd.Flags().BoolVar(&createFlags.validatorManagerMintOnly, "validator-manager-mint-only", false, "only enable validator manager contract to mint new native tokens")
```
felipemadero, line 116: lets just remove this flags:
- both managers should be always minter enabled addresses in native minter precompile
- let's the user decide if adding more minters (double check with meag but It seems to me restrictive to force only one minter for PoS)

cmd/blockchaincmd/create.go
```diff
@@ -197,6 +205,19 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
 		return errMutuallyExlusiveValidatorManagementOptions
 	}
 
+	if createFlags.proofOfAuthority {
+		return errMutuallyExlusiveValidatorManagementOptions
+	}
+
+	if len(createFlags.tokenMinterAddress) > 0 {
```
felipemadero, line 212: lets remove this checks

cmd/blockchaincmd/create.go
```diff
@@ -375,6 +425,57 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func getValidatorContractManagerAddr() ([]string, bool, error) {
```
felipemadero, line 428: lets have this address as a constant and not ask it to the user

cmd/blockchaincmd/create.go
```diff
@@ -375,6 +425,57 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
...
+		}
+		if len(controlAddr) != 0 {
+			return controlAddr, false, nil
+		}
+		ux.Logger.RedXToUser("An address to control Validator Manage Contract is required before proceeding")
+	}
+}
+
+// Configure which addresses may make mint new native tokens
+func getTokenMinterAddr() ([]string, error) {
```
felipemadero, line 447: no need for this

cmd/blockchaincmd/create.go
```diff
@@ -358,6 +379,35 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
 	if err = promptValidatorManagementType(app, sc); err != nil {
```
felipemadero, line 379: we just need logic to always enable native minter precompile is SoV, and add the manager as enabled, when generating evm params

pkg/models/sidecar.go
```diff
@@ -38,8 +38,10 @@ type Sidecar struct {
 	TeleporterVersion string
 	RunRelayer        bool
 	// SubnetEVM based VM's only
-	SubnetEVMMainnetChainID uint
-	ValidatorManagement     ValidatorManagementType
+	SubnetEVMMainnetChainID    uint
+	ValidatorManagement        ValidatorManagementType
+	ValidatorManagerController []string
```
felipemadero, line 43: not needed, one is a constant, the other is in the genesis (really anyway both are at the genesis)

cmd/blockchaincmd/create.go
```diff
@@ -358,6 +359,19 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
...
 		return err
 	}
+	if createFlags.validatorManagerController == nil {
```
felipemadero, line 362: this seems to be done only for PoA. Can we just check is PoA is selected previous to doing this?

cmd/blockchaincmd/create.go
```diff
@@ -358,6 +359,19 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
...
+		var cancelled bool
+		createFlags.validatorManagerController, cancelled, err = getValidatorContractManagerAddr()
+		if err != nil {
+			return err
+		}
+		if cancelled {
+			return fmt.Errorf("user cancelled operation")
+		}
```
felipemadero, line 370: we should check that this address control funds, say, that it has some allocations, cause it needs to pay tx fees

cmd/blockchaincmd/create.go
```diff
@@ -375,6 +389,11 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func getValidatorContractManagerAddr() ([]string, bool, error) {
+	// TODO: replace this with implementation in validator prompt PR
+	return nil, false, nil
```
felipemadero, line 394: implement this giving user key options besides custom. also, I believe you only need one address, so, lets
return string not []string

cmd/blockchaincmd/create.go
```diff
@@ -49,6 +49,7 @@ type CreateFlags struct {
 	useExternalGasToken           bool
 	proofOfStake                  bool
 	proofOfAuthority              bool
+	validatorManagerController    []string
```
felipemadero, line 52: this should be just string

cmd/blockchaincmd/create.go
felipemadero: maybe automatically add funds to it? or probably better, ask the user to give funds or if he desires to add funds

PR #2154, opened 2024-09-16
Genesis icm

felipemadero:
Closes https://github.com/ava-labs/avalanche-cli/issues/2142
Setups icm and registry in subnet-vm genesis.


pkg/contract/allocations.go
```diff
@@ -235,3 +235,39 @@ func GetEVMSubnetGenesisNativeMinterAdmin(
...
+) (bool, error) {
+	genesisData, err := GetBlockchainGenesis(
+		app,
+		network,
+		chainSpec,
+	)
+	if err != nil {
+		return false, err
+	}
+	if !utils.ByteSliceIsSubnetEvmGenesis(genesisData) {
```
arturrez, line 253: nit: can we fix inverse logic and return ContractAddressIsInGenesisData for the happy path and `else` return `false, fmt.Errorf`

pkg/contract/allocations.go
```diff
@@ -235,3 +235,39 @@ func GetEVMSubnetGenesisNativeMinterAdmin(
...
+		return false, fmt.Errorf("checking for contract belonging to genesis is only supported on EVM based vms")
```
arturrez, line 254: nit: error wording is confusing. it seems like `only EVM based vms support genesis contract checks` is better  or some maybe some other way

cmd/blockchaincmd/describe.go
```diff
@@ -276,6 +277,12 @@ func printAllocations(sc models.Sidecar, genesis core.Genesis) error {
 		t.SetTitle("Initial Token Allocation")
 		t.AppendHeader(table.Row{"Description", "Address and Private Key", "Amount (10^18)", "Amount (wei)"})
 		for address := range genesis.Alloc {
```
arturrez, line 279: nit: use `address, allocation := range genesis.Alloc` in this case following `genesis.Alloc[address]` can be replaced with `allocation` in the code

pkg/subnet/local.go
```diff
@@ -81,11 +82,12 @@ type setDefaultSnapshotFunc func(string, bool, string, bool) (bool, error)
 type ICMSpec struct {
 	SkipICMDeploy                bool
 	SkipRelayerDeploy            bool
-	Version                      string
+	ICMVersion                   string
```
arturrez, line 85: nit: not sure that ICMVersion is a good alternative to Version in `ICMSPec` struct. pls ignore this comment

pkg/teleporter/genesis/genesis.go
```diff
@@ -0,0 +1,139 @@
...
+
+func setSimpleStorageValue(
+	storage map[common.Hash]common.Hash,
+	slot string,
+	value string,
+) {
+	storage[common.HexToHash(slot)] = common.HexToHash(value)
+}
+
+func trimHexa(s string) string {
```
arturrez, line 41: we should move these funcs into utils as these just string funcs

pkg/vm/create_evm.go
```diff
@@ -103,6 +106,14 @@ func CreateEVMGenesis(
 		params.initialTokenAllocation[common.HexToAddress(teleporterInfo.FundedAddress)] = core.GenesisAccount{
 			Balance: balance,
 		}
+		if addICMMessengerToGenesis || addICMRegistryToGenesis {
```
arturrez, line 109: logic is not clear for me here. Can we deal with messenger and registry separate. so we have 2 if cases where we check each and deploy each

pkg/contract/allocations.go
felipemadero: this will change the way this is done in other parts of the package. this is a check for a precondition
on the input data, and sometimes there are more checks before or after the processing itself, and
returning here is more related to the condition check.

cmd/blockchaincmd/describe.go
```diff
@@ -275,8 +276,14 @@ func printAllocations(sc models.Sidecar, genesis core.Genesis) error {
...
 		t.AppendHeader(table.Row{"Description", "Address and Private Key", "Amount (10^18)", "Amount (wei)"})
-		for address := range genesis.Alloc {
-			amount := genesis.Alloc[address].Balance
+		for address, allocation := range genesis.Alloc {
+			if len(allocation.Code) > 0 {
+				continue
+			}
+			if address == common.HexToAddress(icmgenesis.MessengerDeployerAddress) {
+				continue
+			}
```
michaelkaplan13, line 285: Why do we skip the deployer address and any allocation with code here? Could be helpful to add a comment explaining

pkg/contract/allocations.go
```diff
@@ -235,3 +235,39 @@ func GetEVMSubnetGenesisNativeMinterAdmin(
 	}
 	return getGenesisNativeMinterAdmin(app, network, genesisData)
 }
+
+func ContractAddressIsInBlockchainGenesis(
```
michaelkaplan13, line 239: Could we add unit tests for these two new functions?

pkg/teleporter/genesis/genesis.go
```diff
@@ -0,0 +1,136 @@
...
+	allocs[common.HexToAddress(RegistryContractAddress)] = core.GenesisAccount{
+		Balance: big.NewInt(0),
+		Code:    deployedRegistryBytes,
+		Storage: storage,
+		Nonce:   1,
+	}
+	return nil
+}
+
+func ICMAtBlockchainGenesis(
```
michaelkaplan13, line 112: Could you add comments documenting what these new functions are and how they are intended to be used?

cmd/blockchaincmd/describe.go
felipemadero: probably we should add a new table for smart contracts.

felipemadero: changing to a 0 balance check, and adding a section on deployed contracts

pkg/contract/allocations.go
felipemadero: one is not used, adding to the other

pkg/teleporter/genesis/genesis.go
felipemadero: done

cmd/blockchaincmd/describe.go
```diff
@@ -298,6 +304,29 @@ func printAllocations(sc models.Sidecar, genesis core.Genesis) error {
...
+			if len(allocation.Code) > 0 {
+				if address == common.HexToAddress(icmgenesis.MessengerContractAddress) {
+					description = "ICM Messenger"
+					deployer = icmgenesis.MessengerDeployerAddress
+				}
+				t.AppendRow(table.Row{description, address.Hex(), deployer})
+			}
+		}
+		ux.Logger.PrintToUser(t.Render())
+	}
```
michaelkaplan13, line 327:
```suggestion
	if len(genesis.Alloc) == 0 {
		return
	}
	
	ux.Logger.PrintToUser("")
	t := table.NewWriter()
	t.Style().Title.Align = text.AlignCenter
	t.Style().Title.Format = text.FormatUpper
	t.Style().Options.SeparateRows = true
	t.SetTitle("Smart Contracts")
	t.AppendHeader(table.Row{"Description", "Address", "Deployer"})
	for address, allocation := range genesis.Alloc {
		var description, deployer string
		if len(allocation.Code) > 0 {
			if address == common.HexToAddress(icmgenesis.MessengerContractAddress) {
				description = "ICM Messenger"
				deployer = icmgenesis.MessengerDeployerAddress
			}
			t.AppendRow(table.Row{description, address.Hex(), deployer})
		}
	}
	ux.Logger.PrintToUser(t.Render())
```

Inverting the if statement like this to return early can help increase code readability by reducing the amount of nesting 👍

michaelkaplan13 (approved):

PR #2156, opened 2024-09-17
make --devnet and --endpoint to work properly with `key list`

felipemadero:


pkg/networkoptions/network_options.go
```diff
@@ -334,6 +335,11 @@ func GetNetworkFromCmdLineFlags(
 		}
 	}
 
+	if networkFlags.Endpoint != "" {
+		re := regexp.MustCompile(`/+$`)
```
arturrez, line 339: why do you need `regex` for something that can be done via `strings.TrimSuffix(s, "/")`

felipemadero: I want to remove any amount of trailing / just in case

arturrez: in my opinion this is excessive but approved it anyway

PR #2157, opened 2024-09-18
Key transfer P to P, C to P, C to C, EVM to EVM

felipemadero:
- add options for P->C and C->P, but also for C->C and, in general, Chain->Chain for EVM chains
- can take only the --endpoint flag, will try to correctly map the network, being `devnet` the default
- adds options for x/p to chain spec so flags and prompts can included them if desired so
- cleanup of key transfer code into one separate function per transfer flow
- correctly map blockchain deploys to the underlying network endpoint, so `key transfer`
  shows more deployed blockchains, and `key list` associates
  more rpcs when using the `--blockchains` flag.


cmd/contractcmd/deploy_erc20.go
```diff
@@ -46,7 +46,8 @@ func newDeployERC20Cmd() *cobra.Command {
 	}
 	networkoptions.AddNetworkFlagsToCmd(cmd, &deployERC20Flags.Network, true, deployERC20SupportedNetworkOptions)
 	deployERC20Flags.PrivateKeyFlags.AddToCmd(cmd, "as contract deployer")
-	deployERC20Flags.chainFlags.AddToCmd(cmd, "deploy the ERC20 contract", true)
+	deployERC20Flags.chainFlags.SetEnabled(true, true, false, false, true)
```
arturrez, line 49: nit: it's hard to read true true false false true to get what it means. pls add comment describing it
pls feel free to ignore this comment

pkg/contract/chain.go
```diff
@@ -100,30 +182,39 @@ func GetBlockchainEndpoints(
...
+		if err != nil {
+			return "", "", fmt.Errorf("failed to load sidecar: %w", err)
+		}
+		networkName := network.Name()
+		if sc.Networks[networkName].BlockchainID == ids.Empty {
+			// look into the cluster deploys
+			for k := range sc.Networks {
+				sidecarNetwork, err := app.GetNetworkFromSidecarNetworkName(k)
+				if err == nil {
+					if sidecarNetwork.Kind == network.Kind && sidecarNetwork.Endpoint == network.Endpoint {
```
arturrez, line 196: nit: we should make a func that we can compare if 2 networks are equal and make it part of the the network.go

arturrez (approved):

PR #2164, opened 2024-09-20
Genesis poa

felipemadero:
## Why this should be merged
Add basic setting for PoA manager:
- add contract code to genesis
- add manager address to minters
- allocate some amount to manager owner

It also enables user to check automatic minting and allocation settings.
Bit more work needs to be done so user may also check teleporter addresses and
token transferrer minting settings. Other PR


cmd/blockchaincmd/create.go
```diff
@@ -220,7 +223,28 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
...
+			if createFlags.poaValidatorManagerOwner == "" {
+				createFlags.poaValidatorManagerOwner, err = getValidatorContractManagerAddr()
+				if err != nil {
+					return err
+				}
+			}
+			sc.PoAValidatorManagerOwner = createFlags.poaValidatorManagerOwner
+			ux.Logger.GreenCheckmarkToUser("PoA Validator Manager Contract owner %s", createFlags.poaValidatorManagerOwner)
+		}
+
```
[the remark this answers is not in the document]
felipemadero: why can't this be before?
this is made before here so as the user can later on (during create) inspect and edit, if desired so:
- manager owner address and the assigned balance (most probably change the assigned balance if desired so)
- native minter precompile settings, including the contract address enabled role

cmd/blockchaincmd/create.go
```diff
@@ -220,7 +223,28 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	sc := &models.Sidecar{}
+
+	if err = promptValidatorManagementType(app, sc); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
felipemadero: same answer. there was a request from devrel over letting the user modify the balance (if so desired) of
the automatic teleporter allocation. why this not implement that, this implementation sets the base to do
it. in principle all prompts that affect allocations and precompiles should be done before, so the user can
inspect and edit settings.

pkg/vm/evm_prompts.go
```diff
@@ -173,6 +170,18 @@ func PromptSubnetEVMGenesisParams(
 		err    error
 		params SubnetEVMGenesisParams
 	)
+	params.initialTokenAllocation = core.GenesisAlloc{}
+
+	if sc.ValidatorManagement == models.ProofOfAuthority {
+		params.UsePoAValidatorManager = true
+		params.enableNativeMinterPrecompile = true
+		params.nativeMinterPrecompileAllowList.EnabledAddresses = []common.Address{
```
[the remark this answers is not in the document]
felipemadero: yes I made a mistake, the PoA is not the minter... changing and adding todo

cmd/blockchaincmd/create.go
```diff
@@ -222,7 +223,28 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
...
+
+	if sc.ValidatorManagement != models.ProofOfAuthority && createFlags.poaValidatorManagerOwner != "" {
+		return errors.New("--poa-manager-owner flag cannot be used when blockchain validator management type is not Proof of Authority")
+	}
+
 	if vmType == models.SubnetEvm {
+		if sc.ValidatorManagement == models.ProofOfAuthority {
```
arturrez, line 237: nit: we use it so many time we should add a func to sidecar `sc.PoA` returning bool and `sc.PoS` with this comparison

pkg/validatormanager/validatormanager.go
```diff
@@ -0,0 +1,67 @@
...
+	"strings"
+
+	"github.com/ava-labs/avalanche-cli/pkg/contract"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/subnet-evm/core"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+const (
+	PoAValidarorMessengerContractAddress = "0x5F584C2D56B4c356e7d82EC6129349393dc5df17"
```
arturrez, line 17: nit: why we want 2 separate const for the same address

pkg/validatormanager/validatormanager.go
```diff
@@ -0,0 +1,67 @@
...
+	rpcURL string,
+	remoteAddress common.Address,
+	privateKey string,
+	subnetID ids.ID,
+	initialOwner common.Address,
+) error {
+	pChainBlockchainID := ids.Empty
+	churnPeriodSeconds := uint64(0)
+	maximumChurnPercentage := uint8(20)
+	type Params struct {
```
arturrez, line 45: nit: should we move this to the top level?

pkg/vm/create_custom.go
```diff
@@ -16,7 +16,8 @@ import (
 	"github.com/ava-labs/avalanche-cli/pkg/ux"
 )
 
-func CreateCustomSidecar(
+func FillCustomSidecar(
```
arturrez, line 19: nit: I feel like CreateCustomSidecar was  fine as it was returning sidecar struct. as with this function it's not obvious that the result of this function is saved in the sidecar passed by pointer.
We should keep CreateCustomSIdecar model , just add a param with old sidecar to be modified and returned.
We can change a name to something like RecieveCustomSidecar or something better

arturrez: this would work like CreateCustomSIdecar if passed old sidecar is null

pkg/vm/evm_prompts.go
```diff
@@ -173,6 +170,21 @@ func PromptSubnetEVMGenesisParams(
 		err    error
 		params SubnetEVMGenesisParams
 	)
+	params.initialTokenAllocation = core.GenesisAlloc{}
+
+	if sc.ValidatorManagement == models.ProofOfAuthority {
```
arturrez, line 175: same

pkg/vm/evm_prompts.go
```diff
@@ -173,6 +170,21 @@ func PromptSubnetEVMGenesisParams(
...
+		params.UsePoAValidatorManager = true
+		params.initialTokenAllocation[common.HexToAddress(sc.PoAValidatorManagerOwner)] = core.GenesisAccount{
+			Balance: defaultPoAOwnerBalance,
+		}
+	}
+
+	if sc.ValidatorManagement == models.ProofOfStake {
```
arturrez, line 182: same

arturrez (approved):

pkg/validatormanager/validatormanager.go
felipemadero: lets keep here for the moment

PR #2179, opened 2024-09-24
ictt deploy fixes

felipemadero:
## Why this should be merged
Closes https://github.com/ava-labs/avalanche-cli/issues/2182 https://github.com/ava-labs/avalanche-cli/issues/2183 https://github.com/ava-labs/avalanche-cli/issues/2184 https://github.com/ava-labs/avalanche-cli/issues/2185 

Besides this, it:
- turn to false the default `ictt deploy` option to remove the native minter admin after deploying a native remote, so there is not an unexpected admin removal. But the flag --remove-minter-admin is kept
- enable also using native minter manager on `ictt deploy`


michaelkaplan13:
@felipemadero, when you have few minutes, could you add a description of what the issues were that are fixed here? It's a bit difficult to tell from the code without knowing beforehand.

felipemadero:
> @felipemadero, when you have few minutes, could you add a description of what the issues were that are fixed here? It's a bit difficult to tell from the code without knowing beforehand.

added description and issues

PR #2192, opened 2024-09-27
Update go 1.22

sukantoraymond:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/ava-labs/avalanche-cli
 
-go 1.22.5
+go 1.22.7
 
 require (
 	github.com/ava-labs/apm v1.0.0


go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/ava-labs/avalanche-cli
 
-go 1.21.12
+go 1.22
```
michaelkaplan13, line 3: We should specify a patch version here also now that it's supported by go.mod

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/ava-labs/avalanche-cli
 
-go 1.22
-
-toolchain go1.22.5
+go 1.22.5
```
michaelkaplan13, line 3: Just double checking the release versions, we typically use the latest patch version of the minor release one behind latest.

This is because we want to include all of the security patches that have been published, and want to give the latest minor version time to become more stable.

In this case, we should be on `1.22.7`
https://go.dev/doc/devel/release#go1.22.minor

arturrez:
updated to 1.22.7

michaelkaplan13 (approved):

michaelkaplan13:
There are a few workflows that are still using go `1.21.X`, but I'm planning on update that in https://github.com/ava-labs/avalanche-cli/compare/main...split-lint-and-test-workflows?expand=1

PR #2196, opened 2024-09-27
Add lint script

michaelkaplan13:
## Why this should be merged
Adds the lint script and configuration used in AvalancheGo, and leverages it in CI.

Note that this script identifies a large number of formatting/style warnings. These should be addressed in follow up PRs, and we will enable requiring the `lint` check to pass to merge PRs once they are.

## How this works
Copies the script from [here](https://github.com/ava-labs/avalanchego/blob/master/scripts/lint.sh).

## How this was tested
Running `./scripts/lint.sh`

## How is this documented
N/A


arturrez (approved):

arturrez:
approved this one. my  only concern is the added complexity of the script

PR #2198, opened 2024-09-27
golang bump to v1.22.7

felipemadero:
## Why this should be merged
A golang bump is needed to be compatible with awm-relayer libraries.
Also, a dependabot dep already bumped a toolchain, which is forcing automatic 
conversion of go.mod each time a compilation is made.

## How this works
Modifies go.mod. It removes the toolchain req. It bumps go version.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,6 @@
 module github.com/ava-labs/avalanche-cli
 
-go 1.21.12
-toolchain go1.22.5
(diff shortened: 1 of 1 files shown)


michaelkaplan13:
Looks like this is a duplicate of https://github.com/ava-labs/avalanche-cli/pull/2192

PR #2202, opened 2024-09-30
Evm debug config

felipemadero:
## Why this should be merged
Users seems to need an easy way to setup subnet evm blockchains, so as evm traces can be obtained.
This PR is adding a flag `--debug` to `blockchain create` that will configure the blockchain for that.
The PR is not trying to make it default, even for developing, but that can be done depending on
PR review comments.
Closes https://github.com/ava-labs/avalanche-cli/issues/1616

## How this works
It adds a --debug flag to blockchain create. When used, it automatically generates a blockchain conf that enables
debugging.

## How this was tested
A blockchain was started on local machine with the given conf, and trace call was made on a tx.

### DIFF
--- a/cmd/blockchaincmd/create.go
+++ b/cmd/blockchaincmd/create.go
@@ -45,6 +45,7 @@ type CreateFlags struct {
 	useLatestReleasedVMVersion    bool
 	useLatestPreReleasedVMVersion bool
 	useExternalGasToken           bool
+	enableDebugging               bool
 }
 
 var (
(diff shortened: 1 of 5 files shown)


cmd/blockchaincmd/create.go
```diff
@@ -358,6 +360,22 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// subnet-evm check based on genesis
+	// covers both subnet-evm vms and custom vms
+	if hasSubnetEVMGenesis, rawErr, err := app.HasSubnetEVMGenesis(blockchainName); rawErr != nil {
```
michaelkaplan13, line 365: Looking at `HasSubnetEVMGenesis`, it looks like the `rawErr` is the last return argument. Are these mixed up here?

pkg/vm/evm_settings.go
```diff
@@ -4,11 +4,15 @@
...
 
 import (
+	_ "embed"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 )
 
+//go:embed evm_debug_config.json
+var EvmDebugConfig []byte
```
michaelkaplan13, line 14: Nit: To follow [standard conventions](https://go.dev/doc/effective_go), abbreviations should all be the same case. Since we want this exported here, I think we should have it be `EVMDebugConfig` to conform.

PR #2211, opened 2024-10-02
Rename sdk test to example

sukantoraymond:
## Why this should be merged
To pass CI without /sdk/ filter

## How this was tested
CI should be green without /sdk/ filter in go test

## How is this documented
no documentation neccessary


michaelkaplan13:
Could we rename the functions in the file something other than `TestXXX()`, and add some comments about what the examples show/how to run them?

PR #2212, opened 2024-10-02
Poa integration

felipemadero:
## Why this should be merged
Adds function to setup PoA manager with an initial validator set, after calling ConvertSubnetTx

## How this works
Adds functions to init PoA with an owner, sign a warp message for convert subnet on p-chain.
init PoA validator set

## How this was tested
A modified version of avalanchego was used that always signs convert subnet warp msgs
A modified version of the manager was used that fix minor issues with latest message format


pkg/evm/evm.go
```diff
@@ -221,6 +225,47 @@ func FundAddress(
...
+		types.AccessTuple{
+			Address:     warp.ContractAddress,
+			StorageKeys: subnetEvmUtils.BytesToHashSlice(predicate.PackPredicate(warpMessage.Bytes())),
+		},
+	}
+	tx := types.NewTx(&types.DynamicFeeTx{
+		ChainID:    chainID,
+		Nonce:      nonce,
+		To:         &contract,
+		Gas:        2_000_000,
```
arturrez, line 258: why this is hardcoded ?

pkg/teleporter/relayer.go
```diff
@@ -362,6 +362,8 @@ func CreateBaseRelayerConfig(
 		SourceBlockchains:      []*config.SourceBlockchain{},
 		DestinationBlockchains: []*config.DestinationBlockchain{},
 		MetricsPort:            metricsPort,
+		DBWriteIntervalSeconds: 10,
```
arturrez, line 365: do you think we need to hardcode this? 
there is a separate PR to use only released version?

pkg/validatormanager/validatormanager.go
```diff
@@ -31,36 +42,238 @@ func AddPoAValidatorManagerContractToAllocations(
...
+func PoAValidatorManagerInitializeValidatorsSet(
+	rpcURL string,
+	managerAddress common.Address,
+	privateKey string,
+	subnetID ids.ID,
+	managerBlockchainID ids.ID,
+	convertSubnetValidators []txs.ConvertSubnetValidator,
+	subnetConversionSignedMessage *warp.Message,
+) (*types.Transaction, *types.Receipt, error) {
+	type InitialValidator struct {
```
arturrez, line 159: we should move types out of the func

pkg/validatormanager/validatormanager.go
```diff
@@ -31,36 +42,238 @@ func AddPoAValidatorManagerContractToAllocations(
...
+		managerAddress,
+		privateKey,
+		subnetID,
+		blockchainID,
+		convertSubnetValidators,
+		subnetConversionSignedMessage,
 	)
-	return err
+	if err != nil {
+		txHashMsg := ""
```
arturrez, line 272: nit:  create func for this and move to utils. something like AddHashToErr(tx), so it can be used like
`fmt.Errorf("failure initializing poa validator manager: %w%s", err, utils.AddHashToErr(tx))`

sdk/interchain/signature-aggregator.go
```diff
@@ -171,11 +171,26 @@ func (s *SignatureAggregator) AggregateSignatures(
...
+	)
+	return hex.EncodeToString(signedMessage.Bytes()), err
+}
+
+// Sign aggregates signatures for a given message and justification.
+//
+// msg is the message to be signed
+// justification is the justification for the signature.
+// Returns the signed message, and an error if the operation fails.
+func (s *SignatureAggregator) Sign(
```
arturrez, line 186: good find. I like it this way

pkg/evm/evm.go
felipemadero: adding a gas limit estimation scheme

pkg/teleporter/relayer.go
felipemadero: added constants. yes, finally I found that increasing the relayer dep on go.mod, updates this struct,
which cause the conf file to be badly generated with incorrect default values

pkg/validatormanager/validatormanager.go
felipemadero: no. the types are only there to be able to be able to properly call TxToMethodWithWarpMessage, but
it is no part of the function call, prefer not to receive a struct but the sequence of params

pkg/validatormanager/validatormanager.go
```diff
@@ -31,36 +42,238 @@ func AddPoAValidatorManagerContractToAllocations(
...
+	}
+	blockchainID, err := contract.GetBlockchainID(
+		app,
+		network,
+		chainSpec,
+	)
+	if err != nil {
+		return err
+	}
+	managerAddress := common.HexToAddress(ValidatorContractAddress)
```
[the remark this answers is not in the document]
felipemadero: it is the same address for both managers (was requested some time ago)

pkg/validatormanager/validatormanager.go
```diff
@@ -31,36 +42,238 @@ func AddPoAValidatorManagerContractToAllocations(
...
 
-func InitializePoAValidatorManager(
+// initializes contract [managerAddress] at [rpcURL], to
+// manage validators on [subnetID], with
+// owner given by [ownerAddress]
+func PoAValidatorManagerInitialize(
 	rpcURL string,
-	remoteAddress common.Address,
+	managerAddress common.Address,
 	privateKey string,
```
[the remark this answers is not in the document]
felipemadero: no, it is the validator contract address

PR #2213, opened 2024-10-02
Integrate ConvertSubnetTx AvalancheGo 

sukantoraymond:
## Why this should be merged
Integrates ConvertSubnetTx from AvalancheGo

## How this works
call subnet deploy command

## How this was tested
PR cannot be built as relayer needs to support this version of AvalancheGo.
 
## How is this documented
Document not implemented yet


cmd/blockchaincmd/deploy.go
```diff
@@ -607,34 +606,16 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
 	}
 
 	if !sidecar.NotSOV {
```
arturrez, line 608: nit: not sure if it's in the scope of the PR but why we have an inverse logic here, should `!sidecar.NotSOV` be just something like sidecar.SOV?

felipemadero (commented):
please fix lint

arturrez (approved):

PR #2216, opened 2024-10-03
node local

arturrez:
## Why this should be merged
it adds commands to support avalanche-cli controlling avalanchego on local
* `node local start`
* `node local stop`
* `node local cleanup`
it also adds `--etna-devnet` flag to allow connection avalanchego running on local to etna devnet 
constants is updated with etna devnet parameters

## How this works
it uses latest ANR to start avalanchego
state is saved to keep data between restarts

## How this was tested
buy running new commands

## How is this documented
not yet


felipemadero (approved):
LGTM

PR #2225, opened 2024-10-08
small nit on `avalanche key transfer` command

meaghanfitzgerald:
## Why this should be merged

When prompted for the sender address, the user currently sees this improper prompt:

```zsh
? Which key source should be used to  for the sender address?:
  ▸ Use stored key
    Use ledger
```

## How this works

Changes the phrasing and corrects spacing so that the user sees:

```zsh
? Which key should be used as the sender address?:
  ▸ Use stored key
    Use ledger
```

## How this was tested

`avalanche key transfer`

## How is this documented

I am currently writing a doc on how to use this command: https://github.com/ava-labs/avalanche-docs/pull/1910

### DIFF
--- a/cmd/keycmd/transfer.go
+++ b/cmd/keycmd/transfer.go
@@ -427,9 +427,9 @@ func transferF(*cobra.Command, []string) error {
 		var useLedger bool
 		goalStr := ""
 		if send {
-			goalStr = " for the sender address"
+			goalStr = "as the sender address"
 		} else {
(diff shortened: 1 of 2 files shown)


felipemadero (commented):
can you use main as the base?

meaghanfitzgerald:
used `main` as base branch https://github.com/ava-labs/avalanche-cli/pull/2226

PR #2227, opened 2024-10-08
ICTT Deployment Fixes

michaelkaplan13:
## Why this should be merged
- Fixes a bug where `homeKey` was left uninitialized prior to use when using an existing `TokenHome` contract and deploying a new `NativeTokenRemote`
- Fixes a bug in balance displays for ERC20 tokens with a decimal count other than 18
- Fixes a bug where the collateral provided to a `TokenHome` contract was previously in the remote denomination. Now it uses the correct amount in the home denomination, as reported by the home contract.
- Adds additional logging

## How this works
- ERC20 balances are now displayed using their specific `decimals()` count if `useGwei` is false.
- The amount of collateral to be added is fetched from the token home rather than using the remote supply.
- Always prompts for the key to use on the home chain for deployment or collateralization.

## How this was tested
Going through the flow of running `avalanche ictt deploy --deploy-native-remote` for a Fuji L1, and deploying a `NativeTokenRemote` connected to the existing USDC `TokenHome` contract on the C-Chain

## How is this documented
N/A

### DIFF
--- a/cmd/interchaincmd/tokentransferrercmd/deploy.go
+++ b/cmd/interchaincmd/tokentransferrercmd/deploy.go
@@ -492,7 +492,6 @@ func CallDeploy(_ []string, flags DeployFlags) error {
 	if err := ictt.BuildContracts(app); err != nil {
 		return err
 	}
-	ux.Logger.PrintToUser("")
 
 	// Home Deploy
 	icttSrcDir, err := ictt.RepoDir(app)
@@ -676,8 +675,10 @@ func CallDeploy(_ []string, flags DeployFlags) error {
 			return err
 		}
 	}
+	ux.Logger.PrintToUser("Remote Deployed to %s", remoteRPCEndpoint)
+	ux.Logger.PrintToUser("Remote Address: %s", remoteAddress)
 
-	if err := ictt.RegisterERC20Remote(
+	if err := ictt.RegisterRemote(
 		remoteRPCEndpoint,
 		remoteKey,
 		remoteAddress,
@@ -688,6 +689,7 @@ func CallDeploy(_ []string, flags DeployFlags) error {
 	checkInterval := 100 * time.Millisecond
 	checkTimeout := 10 * time.Second
 	t0 := time.Now()
+	var collateralNeeded *big.Int
 	for {
 		registeredRemote, err := ictt.TokenHomeGetRegisteredRemote(
 			homeRPCEndpoint,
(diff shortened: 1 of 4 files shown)


arturrez (approved):

pkg/ictt/operate.go
```diff
@@ -458,6 +459,7 @@ func TokenHomeAddCollateral(
 	remoteAddress common.Address,
 	amount *big.Int,
 ) error {
+	ux.Logger.PrintToUser("Collateralizing remote contract on the home chain")
```
felipemadero, line 462: should we add action description logs to this lib? if this is the case, should we pass the logger as arg?

pkg/ictt/deploy.go
```diff
@@ -26,11 +27,12 @@ type TokenRemoteSettings struct {
 	TokenHomeDecimals         uint8
 }
 
-func RegisterERC20Remote(
+func RegisterRemote(
 	rpcURL string,
 	privateKey string,
 	remoteAddress common.Address,
 ) error {
+	ux.Logger.PrintToUser("Registering remote contract with home contract")
```
felipemadero, line 35: same comment on logs

cmd/keycmd/list.go
```diff
@@ -558,28 +558,44 @@ func getEvmBasedChainAddrInfo(
...
+			// Get the raw balance for the given token.
+			balance, err := token.BalanceOf(nil, common.HexToAddress(cChainAddr))
+			if err != nil {
+				return addressInfos, err
+			}
+
+			// Get the decimal count for the token to format the balance.
+			// Note: decimals() is not officially part of the IERC20 interface, but is a common extension.
+			decimals, err := token.Decimals(nil)
+			if err != nil {
```
felipemadero, line 577: should we default to some standard decimal in case of error for not being official?

cmd/interchaincmd/tokentransferrercmd/deploy.go
```diff
@@ -304,41 +343,11 @@ func CallDeploy(_ []string, flags DeployFlags) error {
...
-			}
-		}
-		pk, err := crypto.HexToECDSA(homeKey)
-		if err != nil {
-			return err
-		}
-		homeKeyAddress = crypto.PubkeyToAddress(pk.PublicKey).Hex()
+	// Get the key to be used to deploy the token home contract and collateralize the remote
+	// in the case that it is a native token remote.
+	homeKey, homeKeyAddress, err := getHomeKeyAndAddress(app, network, flags.homeFlags)
```
felipemadero, line 348: previously the bug was not asking for it when needed, but now it may also ask for it when not needed,
as in the case the home preexists and there is no collateralization needed.

PR #2228, opened 2024-10-08
Track subnet refactor 

sukantoraymond:
## Why this should be merged

Refactoring node sync functionality from nodecmd to pkd/node so that we can call node sync from subnet deploy command without encountering import cycle error

## How this works

Moving code from nodecmd to pkd/node

## How this was tested
./scripts/build.sh works

## How is this documented
NA


pkg/node/helper.go
```diff
@@ -0,0 +1,214 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package node
```
arturrez, line 3: all these func already exist in `cmd/nodecmd`

arturrez: did you forgot to remove them from the source when you moved them?

arturrez (approved):

PR #2233, opened 2024-10-10
Acp77 remote etna devnet

arturrez:
## Why this should be merged
adds `--etna-devnet` flag to `node create` command.
adds bootstrap and upgrade/genesis options

## How this works
passes required params to avalanchego configured with `node create`
also does param verification

## How this was tested
`./bin/avalanche node create --etna-devnet  etna0 --aws  --use-static-ip=false`

```
ubuntu@ip-172-31-20-121:~/.avalanchego$ cat configs/node.json 
{
        "http-host": "127.0.0.1",
        "api-admin-enabled": false,
        "index-enabled": false,
        "proposervm-use-current-height-bool": true,
        "network-id": "network-76",
        "bootstrap-ids": "NodeID-WrLWMK5sJ4dBUAsx1dP2FUyTqrYwbFA1,NodeID-bojBKDrpt81bYhxYKQfLw89V7CpoH2m7,NodeID-8LbTmmGsDC991SbD8Nkx88VULT3XYzYXC,NodeID-DDhXtFm6Q9tCq2yiFRmcSMKvHgUgh8yQC,NodeID-QDYnWDQd6g4cQ5H6yiWNqSmfRMBqEH9AG,NodeID-P5QGH4EXddrcyNAzkqyZKHXgEpVX6HExL,NodeID-78ibWpjtZz5ZGT6EyTEdu8VKmboUHTuGT,NodeID-7eRvnfs2a2PvrPHUuCRRpPVAoVjbWxaFG,NodeID-gpXWBExQSZXqJPQt6L6MnveUfgr7HJ4q,NodeID-L4CY8B5uVSDe4cnN1BpeDsHacMp4q4q8q",
        "bootstrap-ips": "107.21.11.213:9651,34.233.248.130:9651,52.201.126.172:9651,35.170.144.5:9651,98.82.41.186:9651,34.228.34.127:9651,44.205.136.166:9651,52.6.31.40:9651,54.197.98.148:9651,18.211.108.228:9651",
        "genesis-file": "/.avalanchego/configs/genesis.json",
        "upgrade-file": "/.avalanchego/configs/upgrade.json",
        "public-ip": "35.175.178.29",
        "db-dir": "/.avalanchego/db/",
        "log-dir": "/.avalanchego/logs/"
}
```

### DIFF
--- a/cmd/nodecmd/create.go
+++ b/cmd/nodecmd/create.go
@@ -5,6 +5,7 @@ package nodecmd
 import (
 	"fmt"
 	"math"
+	"net/netip"
 	"os"
 	"os/user"
 	"path/filepath"
@@ -76,6 +77,12 @@ var (
 	grafanaPkg           string
 	wizSubnet            string
 	publicHTTPPortAccess bool
+
+	bootstrapIDs  []string
+	bootstrapIPs  []string
+	genesisPath   string
+	upgradePath   string
+	useEtnaDevnet bool
 )
 
 func newCreateCmd() *cobra.Command {
@@ -128,6 +135,11 @@ will apply to all nodes in the cluster`,
 	cmd.Flags().IntVar(&volumeSize, "aws-volume-size", constants.CloudServerStorageSize, "AWS volume size in GB")
 	cmd.Flags().BoolVar(&replaceKeyPair, "auto-replace-keypair", false, "automatically replaces key pair to access node if previous key pair is not found")
 	cmd.Flags().BoolVar(&publicHTTPPortAccess, "public-http-port", false, "allow public access to avalanchego HTTP port")
+	cmd.Flags().StringArrayVar(&bootstrapIDs, "bootstrap-id", []string{}, "nodeIDs of bootstrap nodes")
+	cmd.Flags().StringArrayVar(&bootstrapIPs, "bootstrap-ip", []string{}, "IP:port pairs of bootstrap nodes")
+	cmd.Flags().StringVar(&genesisPath, "genesis", "", "path to genesis file")
+	cmd.Flags().StringVar(&upgradePath, "upgrade", "", "path to upgrade file")
+	cmd.Flags().BoolVar(&useEtnaDevnet, "etna-devnet", false, "use Etna devnet. Prepopulated with Etna DevNet bootstrap configuration along with genesis and upgrade files")
 	return cmd
 }
 
@@ -199,6 +211,25 @@ func preCreateChecks(clusterName string) error {
 	if err := failForExternal(clusterName); err != nil {
 		return err
 	}
+	// bootsrap checks
+	if useEtnaDevnet && (len(bootstrapIDs) != 0 || len(bootstrapIPs) != 0 || genesisPath != "" || upgradePath != "") {
+		return fmt.Errorf("etna devnet uses predefined bootsrap configuration")
+	}
+	if len((bootstrapIDs)) != len(bootstrapIPs) {
+		return fmt.Errorf("number of bootstrap ids and ip:port pairs must be equal")
+	}
+	if genesisPath != "" && !utils.FileExists(genesisPath) {
+		return fmt.Errorf("genesis file %s does not exist", genesisPath)
+	}
+	if upgradePath != "" && !utils.FileExists(upgradePath) {
+		return fmt.Errorf("upgrade file %s does not exist", upgradePath)
+	}
+	// check ip:port pairs
+	for _, ip := range bootstrapIPs {
+		if _, err := netip.ParseAddrPort(ip); err != nil {
+			return fmt.Errorf("invalid ip:port pair %s", ip)
+		}
+	}
 
 	return nil
 }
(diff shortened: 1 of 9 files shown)


cmd/nodecmd/create.go
```diff
@@ -76,6 +77,12 @@ var (
 	grafanaPkg           string
 	wizSubnet            string
 	publicHTTPPortAccess bool
+
+	bootstrapIDs  []string
+	bootstrapIPs  []string
+	genesisPath   string
+	upgradePath   string
+	useEtnaDevnet bool
```
michaelkaplan13, line 85: I think `useEtnaDevnet` should be put into the existing `NetworkFlags` rather than being treated differently here.

cmd/nodecmd/create.go
```diff
@@ -128,6 +135,11 @@ will apply to all nodes in the cluster`,
 	cmd.Flags().IntVar(&volumeSize, "aws-volume-size", constants.CloudServerStorageSize, "AWS volume size in GB")
 	cmd.Flags().BoolVar(&replaceKeyPair, "auto-replace-keypair", false, "automatically replaces key pair to access node if previous key pair is not found")
 	cmd.Flags().BoolVar(&publicHTTPPortAccess, "public-http-port", false, "allow public access to avalanchego HTTP port")
+	cmd.Flags().StringArrayVar(&bootstrapIDs, "bootstrap-id", []string{}, "nodeIDs of bootstrap nodes")
+	cmd.Flags().StringArrayVar(&bootstrapIPs, "bootstrap-ip", []string{}, "IP:port pairs of bootstrap nodes")
```
michaelkaplan13, line 139:
```suggestion
	cmd.Flags().StringArrayVar(&bootstrapIDs, "bootstrap-ids", []string{}, "nodeIDs of bootstrap nodes")
	cmd.Flags().StringArrayVar(&bootstrapIPs, "bootstrap-ips", []string{}, "IP:port pairs of bootstrap nodes")
```

cmd/nodecmd/create.go
```diff
@@ -199,6 +211,25 @@ func preCreateChecks(clusterName string) error {
...
+	}
+	if upgradePath != "" && !utils.FileExists(upgradePath) {
+		return fmt.Errorf("upgrade file %s does not exist", upgradePath)
+	}
+	// check ip:port pairs
+	for _, ip := range bootstrapIPs {
+		if _, err := netip.ParseAddrPort(ip); err != nil {
+			return fmt.Errorf("invalid ip:port pair %s", ip)
+		}
+	}
```
michaelkaplan13, line 232: Can we unit test these cases?

Also should we allow bootstrap IDs/IPs in the case Fuji/Mainnet?

cmd/nodecmd/create.go
```diff
@@ -240,21 +271,45 @@ func stringToAWSVolumeType(input string) types.VolumeType {
...
-		globalNetworkFlags,
-		false,
-		true,
-		createSupportedNetworkOptions,
-		"",
-	)
-	if err != nil {
-		return err
+	// etna devnet constants
+	if useEtnaDevnet {
```
michaelkaplan13, line 281: It will also be cleaner here to support the Etna case within `GetNetworkFromCmdLineFlags` rather than needing to have custom handling for it throughout.

cmd/nodecmd/create.go
arturrez: etnadevnet is a shortcut for the devnet
we already have devnet here so I don't think we should

pkg/constants/etna.go
```diff
@@ -0,0 +1,45 @@
...
+		"34.233.248.130:9651",
+		"52.201.126.172:9651",
+		"35.170.144.5:9651",
+		"98.82.41.186:9651",
+		"34.228.34.127:9651",
+		"44.205.136.166:9651",
+		"52.6.31.40:9651",
+		"54.197.98.148:9651",
+		"18.211.108.228:9651",
+	}
```
michaelkaplan13, line 44: There should only be 5 bootstrap nodes I believe:
```
  --bootstrap-ids="NodeID-8LbTmmGsDC991SbD8Nkx88VULT3XYzYXC,NodeID-bojBKDrpt81bYhxYKQfLw89V7CpoH2m7,NodeID-WrLWMK5sJ4dBUAsx1dP2FUyTqrYwbFA1,NodeID-DDhXtFm6Q9tCq2yiFRmcSMKvHgUgh8yQC,NodeID-QDYnWDQd6g4cQ5H6yiWNqSmfRMBqEH9AG"
    --bootstrap-ips="52.201.126.172:9651,34.233.248.130:9651,107.21.11.213:9651,35.170.144.5:9651,98.82.41.186:9651"
```

pkg/docker/ssh.go
```diff
@@ -25,7 +25,17 @@ func ValidateComposeFile(host *models.Host, composeFile string, timeout time.Dur
...
 // ComposeSSHSetupNode sets up an AvalancheGo node and dependencies on a remote host over SSH.
-func ComposeSSHSetupNode(host *models.Host, network models.Network, avalancheGoVersion string, withMonitoring bool, publicAccessToHTTPPort bool) error {
+func ComposeSSHSetupNode(
+	host *models.Host,
+	network models.Network,
+	avalancheGoVersion string,
+	avalanchegoBootstrapIDs []string,
+	avalanchegoBootstrapIPs []string,
+	avalanchegoGenesisFilePath string,
+	avalanchegoUpgradeFilePath string,
```
michaelkaplan13, line 35: Would it be worth encapsulating all of these options in an `AvalancheGoConfigOptions` struct so that they don't need to be passed around individually?

pkg/ssh/ssh.go
```diff
@@ -182,8 +182,18 @@ func RunSSHUpgradeAvalanchego(host *models.Host, network models.Network, avalanc
 	if err != nil {
 		return err
 	}
-
-	if err := docker.ComposeSSHSetupNode(host, network, avalancheGoVersion, withMonitoring, publicAccessToHTTPPort); err != nil {
+	if err := docker.ComposeOverSSH("Compose Node",
```
michaelkaplan13, line 185: Looks like we can rid of the `network` parameter here now if it's no longer used?

cmd/nodecmd/create.go
arturrez: agree

cmd/nodecmd/create.go
arturrez: good point.
etna devnet should be treated like fuji, the only difference is lack of prebuilt genesis/boostrapper info that is included into avalanchego pkg. For enta devnet we need to specify bootstrap config in avalanchego config. We just have it in constants to help user

arturrez: in my opinion it should be something like `custom`

cmd/nodecmd/create.go
arturrez: it's allowed. these params are available for user to use.
I feel like I should move it into `pkg/utils` so I can add unittest for verification `func`

cmd/nodecmd/create.go
arturrez: etnadevnet is shiortcut with predefined bootstrap configuration for custom external network.
let me think about appropriate name for this. External or custom might be a correct name.
Once it's merged we should be able to support any other remote networks like etnadevnet or whatever we will need going forward

cmd/nodecmd/create.go
michaelkaplan13: More generally, it seems like `network_options.go` currently contains logic for supported networks to connect to. I think it would be error prone and less organized to put other network option logic in different places. This way it's all contained in the same places so it's easier to reason about.

pkg/constants/etna.go
arturrez: I used all validators available.
let me fix this. and also open a PR for ANR

PR #2234, opened 2024-10-10
Update avalanche go ledger

sukantoraymond:
## Why this should be merged
Uses newest ledger avalanche version through avalanche go version in https://github.com/ava-labs/avalanchego/pull/3456 to fix issue https://github.com/ava-labs/ledger-avalanche/issues/326

## How this works
Updates avalanche go in go mod

## How this was tested
NA

## How is this documented
NA

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.22.8
 require (
 	github.com/ava-labs/apm v1.0.0
 	github.com/ava-labs/avalanche-network-runner v1.8.4-0.20241003175856-6b1bcdf33e7a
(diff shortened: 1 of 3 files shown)


cmd/keycmd/transfer.go
```diff
@@ -598,6 +599,20 @@ func transferF(*cobra.Command, []string) error {
 			return fmt.Errorf("error signing tx: %w", err)
 		}
 
+		pContext := wallet.P().Builder().Context()
```
arturrez, line 602: not clear for me why we need this additional code

arturrez (approved):

PR #2235, opened 2024-10-10
Integrate local machine into subnet deploy 

sukantoraymond:
## Why this should be merged
Integrates using local machine as bootstrap validator into subnet deploy command. 

## How this works
// create the local node connected to etna
go run main.go node local start <EtnaInstanceName> --etna-devnet --avalanchego-path <AVALANCHEGO_PATH>

// create a blockchain conf
go run main.go blockchain create <BlockchainName>

// create subnet+blockchain+convert on the blokchain
go run main.go blockchain deploy <BlockchainName> --cluster <EtnaInstanceName> 

## How this was tested
./bin/avalanche subnet deploy will have to achieve L1 set up with local machine as sole bootstrap validator in Etna Devnet

## How is this documented
NA

### DIFF
--- a/cmd/blockchaincmd/deploy.go
+++ b/cmd/blockchaincmd/deploy.go
@@ -818,7 +818,12 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
 			return err
 		}
 		ux.Logger.GreenCheckmarkToUser("Subnet is successfully converted into Subnet Only Validator")
+	} else {
+		if err := app.UpdateSidecarNetworks(&sidecar, network, subnetID, blockchainID, "", "", nil); err != nil {
+			return err
+		}
 	}
+
 	flags := make(map[string]string)
 	flags[constants.MetricsNetwork] = network.Name()
 	metrics.HandleTracking(cmd, constants.MetricsSubnetDeployCommand, app, flags)
--- a/cmd/nodecmd/local.go
+++ b/cmd/nodecmd/local.go
@@ -468,11 +468,3 @@ func notImplementedForLocal(what string) error {
 	ux.Logger.PrintToUser("Unsupported cmd: %s is not supported by local clusters", logging.LightBlue.Wrap(what))
 	return nil
 }
-
-func getRPCEndpoint(endpoint string, blockchainID string) string {
-	return models.NewDevnetNetwork(endpoint, 0).BlockchainEndpoint(blockchainID)
-}
-
-func getWSEndpoint(endpoint string, blockchainID string) string {
-	return models.NewDevnetNetwork(endpoint, 0).BlockchainWSEndpoint(blockchainID)
-}
--- a/pkg/node/sync.go
+++ b/pkg/node/sync.go
@@ -322,13 +322,9 @@ func TrackSubnetWithLocalMachine(app *application.Avalanche, clusterName, blockc
 }
 
 func CheckClusterIsLocal(app *application.Avalanche, clusterName string) (bool, error) {
-	clustersConfig := models.ClustersConfig{}
-	if app.ClustersConfigExists() {
-		var err error
-		clustersConfig, err = app.LoadClustersConfig()
-		if err != nil {
-			return false, err
-		}
+	clustersConfig, err := app.GetClustersConfig()
+	if err != nil {
+		return false, err
 	}
 	clusterConf, ok := clustersConfig.Clusters[clusterName]
 	return ok && clusterConf.Local, nil


cmd/blockchaincmd/deploy.go
```diff
@@ -464,6 +466,18 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
...
+	ux.Logger.PrintToUser("You can use your local machine as a bootstrap validator on the blockchain")
+	ux.Logger.PrintToUser("This means that you don't have to to set up a remote server on a cloud service (e.g. AWS / GCP) to be a validator on the blockchain.")
+
+	useLocalMachine, err := app.Prompt.CaptureYesNo("Do you want to use your local machine as a bootstrap validator?")
+	if err != nil {
+		return err
+	}
+
+	if useLocalMachine {
+		bootstrapEndpoints = []string{"http://127.0.0.1:9650"}
```
arturrez, line 478: this might be tricky to assume this endpoint as port can be not 9650 but rather dynamic 
cc @felipemadero

arturrez: maybe we should ask user for the endpoint and not yes/no for local machine.
we can hint that it can be 127.0.0.1:9650 for the local machine

pkg/node/helper.go
```diff
@@ -0,0 +1,298 @@
...
+func CheckCluster(app *application.Avalanche, clusterName string) error {
+	_, err := GetClusterNodes(app, clusterName)
+	return err
+}
+
+func GetClusterNodes(app *application.Avalanche, clusterName string) ([]string, error) {
+	if exists, err := CheckClusterExists(app, clusterName); err != nil || !exists {
+		return nil, fmt.Errorf("cluster %q not found", clusterName)
+	}
+	clustersConfig, err := app.LoadClustersConfig()
```
arturrez, line 40: nit: use 
```
clusterConfig, err := app.GetClusterConfig(clusterName)
	if err != nil {
		return err
	}
	```
	and
	`clusterConfig.Nodes` instead

**Review by arturrez — APPROVED** [review]

File: cmd/nodecmd/local.go [hunk 3]
```diff
@@ -415,14 +416,17 @@ func localStopNode(_ *cobra.Command, _ []string) error {
 func localDestroyNode(_ *cobra.Command, args []string) error {
 	clusterName := args[0]
 
-	localStopNode(nil, nil)
+	if err := localStopNode(nil, nil); err != nil {
+		return fmt.Errorf("failed to destroy local node: %w", err)
```
felipemadero, line 420: node can be already stopped. for the moment on ignore the error or verify is the node is already stopped,
we want to continue removing all the directorioes that is the reason the error is ignored

cmd/nodecmd/local.go
```diff
@@ -452,107 +456,10 @@ func addLocalClusterConfig(network models.Network) error {
...
-	}
-	networkInfo.RPCEndpoints = rpcEndpoints.List()
-	networkInfo.WSEndpoints = wsEndpoints.List()
-	sc.Networks[clusterConfig.Network.Name()] = networkInfo
-	if err := app.UpdateSidecar(&sc); err != nil {
-		return err
-	}
-	ux.Logger.GreenCheckmarkToUser("%s successfully tracking %s", clusterName, blockchainName)
-	return nil
+	return node.TrackSubnetWithLocalMachine(app, clusterName, blockchainName)
```
felipemadero, line 462: I am not sure it is the moment to move this to node pgk. we keep a part of logic here and a part in node.
also, maybe node is not the final dest package.

pkg/node/helper.go
```diff
@@ -0,0 +1,298 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package node
```
felipemadero, line 3: this is for another PR?

felipemadero: anyway, no problem keeping here

cmd/blockchaincmd/deploy.go
```diff
@@ -725,65 +739,81 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
...
-			chainSpec := contract.ChainSpec{
-				BlockchainName: blockchainName,
-			}
-			genesisAddress, genesisPrivateKey, err := contract.GetEVMSubnetPrefundedKey(
-				app,
-				network,
-				chainSpec,
-			)
-			if err != nil {
+		clusterName, err := node.GetClusterNameFromList(app)
```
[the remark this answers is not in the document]
arturrez: models.Networks has ClusterName

PR #2238, opened 2024-10-11
Fix subnet describe for etna

sukantoraymond:
## Why this should be merged
Fix blockchain describe for etna

## How this works
call blockchain describe 

## How this was tested
call blockchain describe 

## How is this documented
NA

### DIFF
--- a/pkg/models/network.go
+++ b/pkg/models/network.go
@@ -20,6 +20,7 @@ const (
 	Fuji
 	Local
 	Devnet
+	Etna
 )
 
 func (nk NetworkKind) String() string {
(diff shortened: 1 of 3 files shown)


arturrez:
this should be reviewed after https://github.com/ava-labs/avalanche-cli/pull/2233 as it implements the same thing in a different way

pkg/models/network.go
```diff
@@ -32,6 +33,8 @@ func (nk NetworkKind) String() string {
 		return "Local Network"
 	case Devnet:
 		return "Devnet"
+	case Etna:
```
arturrez, line 36: pls wait for https://github.com/ava-labs/avalanche-cli/pull/2233 to be merged as this PR implements this in a different way

PR #2239, opened 2024-10-11
add checks for local cluster. minor refactor

arturrez:
## Why this should be merged
it adds checks for unsupported  by local cluster commands

## How this works
adds checks
also refactors getclustersconfig functionality so it's less verbose to use 

## How this was tested
❯ ./bin/avalanche node validate primary local1
Local cluster local1 does not support validate primary cmd

## How is this documented
n/a

### DIFF
--- a/cmd/nodecmd/create.go
+++ b/cmd/nodecmd/create.go
@@ -833,28 +833,22 @@ func saveExternalHostConfig(externalHostConfig models.RegionConfig, hostRegion, 
 }
 
 func getExistingMonitoringInstance(clusterName string) (string, error) {
-	if app.ClustersConfigExists() {
-		clustersConfig, err := app.LoadClustersConfig()
-		if err != nil {
-			return "", err
-		}
-		if _, ok := clustersConfig.Clusters[clusterName]; ok {
-			if clustersConfig.Clusters[clusterName].MonitoringInstance != "" {
-				return clustersConfig.Clusters[clusterName].MonitoringInstance, nil
-			}
+	clustersConfig, err := app.GetClustersConfig()
+	if err != nil {
+		return "", err
+	}
+	if _, ok := clustersConfig.Clusters[clusterName]; ok {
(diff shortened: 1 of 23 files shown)


cmd/nodecmd/create.go
```diff
@@ -833,28 +833,22 @@ func saveExternalHostConfig(externalHostConfig models.RegionConfig, hostRegion,
...
-	if app.ClustersConfigExists() {
-		clustersConfig, err := app.LoadClustersConfig()
-		if err != nil {
-			return "", err
-		}
-		if _, ok := clustersConfig.Clusters[clusterName]; ok {
-			if clustersConfig.Clusters[clusterName].MonitoringInstance != "" {
-				return clustersConfig.Clusters[clusterName].MonitoringInstance, nil
-			}
+	clustersConfig, err := app.GetClustersConfig()
```
felipemadero, line 836: I believe node create can be called over an existing cluster. So it may need the local check

cmd/nodecmd/local.go
```diff
@@ -511,3 +508,8 @@ func localTrack(_ *cobra.Command, args []string) error {
 	ux.Logger.GreenCheckmarkToUser("%s successfully tracking %s", clusterName, blockchainName)
 	return nil
 }
+
+func notImplementedForLocal(clusterName string, what string) error {
+	ux.Logger.PrintToUser("Local cluster %s does not support %s cmd", logging.LightBlue.Wrap(clusterName), logging.LightBlue.Wrap(what))
```
felipemadero, line 513: I would rather use "Unsupported cmd: %s is not supported by local clusters". A nit feel free to think on this
and find out a suitable variation

felipemadero (approved):
preapproving

cmd/nodecmd/create.go
arturrez: fixed

cmd/nodecmd/local.go
arturrez: agree. changed

PR #2242, opened 2024-10-13
support urls without scheme

felipemadero:
## Why this should be merged
It is desirable for cmdline ux to be able to provide urls without scheme as for example 127.0.0.1:9650/ext/bc/2TaAjq8oF5PRrGDzX7DvWxsdei1AZvHRZ7TAdNpqrbCEZKcQ55/rpc
This capability is presented in tools like forge.

### DIFF
--- a/pkg/evm/evm.go
+++ b/pkg/evm/evm.go
@@ -6,6 +6,8 @@ import (
 	"crypto/ecdsa"
 	"fmt"
 	"math/big"
+	"net/url"
+	"strings"
 	"time"
 
 	"github.com/ava-labs/avalanche-cli/pkg/utils"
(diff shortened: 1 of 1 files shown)


pkg/evm/evm.go
```diff
@@ -259,15 +261,80 @@ func SendTransaction(
...
+				return client, scheme, nil
+			}
+		}
+	}
+	return nil, "", notDeterminedErr
+}
+
+func HasScheme(rpcURL string) (bool, error) {
+	if parsedURL, err := url.Parse(rpcURL); err != nil {
+		if !strings.Contains(err.Error(), "first path segment in URL cannot contain colon") {
```
arturrez, line 311: 1. not clear for me why we can't return what url.Parse() gives us and need additional check for the type of error. 
2. I feel like we need unittests for this to explain functionality like example where this `first path segment in URL` is relavant
3.  should be moved to utils/net.go

so the main question is why it has to be more complicated then to rely on  url.Parse  to check if it's a url and has some scheme and then check if Scheme is not empty

pkg/evm/evm.go
```diff
@@ -259,15 +261,80 @@ func SendTransaction(
...
+		return nil, err
+	}
 	for i := 0; i < repeatsOnFailure; i++ {
 		ctx, cancel := utils.GetAPILargeContext()
 		defer cancel()
-		client, err = ethclient.DialContext(ctx, rpcURL)
+		if hasScheme {
+			client, err = ethclient.DialContext(ctx, rpcURL)
+		} else {
+			client, _, err = FindOutScheme(rpcURL)
```
arturrez, line 336: nit: not sure FIndOutScheme is a good name for this as it returns a EthClient.  I feel like FindOutScheme should be a helper function to detect endpoint protocol and correct corresponing rpcURL by adding scheme  so `ethclient.DialContext(ctx, rpcURL)` can be still used to get a client

arturrez: pls feel  free to ignore this comment

pkg/evm/evm.go
```diff
@@ -259,15 +261,80 @@ func SendTransaction(
 	return err
 }
 
+func FindOutScheme(rpcURL string) (ethclient.Client, string, error) {
```
arturrez, line 264: I feel like it will be more flexible if this detects and returns scheme not `ethclient.Client`
in this case it will have more generic use

PR #2249, opened 2024-10-15
Version 1.7.6

sukantoraymond:
## Why this should be merged
Update version

## How this works
NA

## How this was tested
nA

## How is this documented
NA


felipemadero (approved):
OFC LGTM

PR #2256, opened 2024-10-15
Docker acp77 latest

arturrez:
## Why this should be merged
it adds a additional step for acp-77 branch to build docker image  avaplatform/avalanche-cli:acp77-latest on every push

## How this works
adds a new step for github CI

## How this was tested
built an image and run it from Linux OS
`docker run avaplatform/avalanche-cli:acp77-latest config metrics  disable`

## How is this documented
n/a


felipemadero (approved):
LGTM

PR #2281, opened 2024-10-18
Acp77 remote enta cluster fixes

arturrez:
## Why this should be merged
adds automatic bls info if cluster name provided 

## How this works
it reads nodeiD and bls info from staking and node bls keys as we created the, in advance and provide them to cluster nodes 

## How this was tested
```
./bin/avalanche node create remote0
 ./bin/avalanche blockchain deploy remote0 --cluster remote0
./bin/avalanche node sync remote0 remote0

### DIFF
--- a/cmd/blockchaincmd/deploy.go
+++ b/cmd/blockchaincmd/deploy.go
@@ -498,7 +498,10 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
 						}
 					}
 					network = models.NewNetworkFromCluster(network, clusterName)
+				} else {
+					// get bootrstrap endpoints from  remote cluster
 				}
+
 			}
 			// ask user if we wants to use local machine if cluster is not provided
 			if !useLocalMachine && globalNetworkFlags.ClusterName == "" {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cmd/blockchaincmd/deploy.go [hunk 1]
```diff
@@ -575,7 +576,15 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
 					ChangeOwnerAddr:      changeAddr,
 				})
 			}
-		} else {
+
+		case globalNetworkFlags.ClusterName != "":
```
felipemadero, line 580: are we sure here this is not local cluster?

arturrez: yes we are sure as above this we have a code 
```
if len(bootstrapEndpoints) == 0 {
					bootstrapEndpoints, err = getLocalBootstrapEndpoints()
```
so bootstrapEndoints will be already set and slice len will be >0  for local case

arturrez: added comment for this just in case

felipemadero (approved):

PR #2284, opened 2024-10-20
Acp 77 fix

felipemadero:
## Why this should be merged
Avoids usage of global AVALANCHEGO path env var it set


arturrez (approved):
pls fix lint

PR #2303, opened 2024-10-31
Use network name instead of cluster name on sidecar

sukantoraymond:
## Why this should be merged
Currently if a blockchain is created with cluster flag and user uses additional commands like addValidator
```
avalanche-cli/bin/avalanche blockchain addValidator <blockchainName>
✔ Devnet
✔ Custom
✔ Devnet Endpoint: https://etna.avax-dev.network/█
✔ Use stored key
✔ mytestkey
What is the NodeID of the node you want to add as a blockchain validator?: NodeID-8XXvVGYhuQTgDpU5xQrAAYHA96o5tNSiR
Next, we need the public key and proof of possession of the node's BLS
Check https://docs.avax.network/api-reference/info-api#infogetnodeid for instructions on calling info.getNodeID API
✔ What is the node's BLS public key?: 0x8e1de3a59cc61657a120ed330d77bb48439b5b9084ed51ae2571a8c34c8da35e21a07df5bb0a1ffd3c18b41af63d8e09█
✔ What is the node's BLS proof of possession?: 0xb54be09f8c8d2e5e5effe99e40ca141d186baa01e7a1b7c81376e2590b2c264084b9e80cecc61c80a59d43133533cd38146e2b533912b0ac33d3c44d79e7057ce5a161a0605b267fa477a7740bbd17a41e7db9bea82096587c5a2397502cf570█
PoA manager owner 0x6b603F6B045D89beE25dbA0963B1D40687602F00 pays for the initialization of the validator's registration (Blockchain gas token)
Error: blockchain has not been deployed to Devnet https://etna.avax-dev.network/
```
This is because we are storing where the blockchain is deployed in in cluster name instead of which network it was deployed in.

## How this works
Sidecar should show network name instead of cluster name to enable interchangeability of using --cluster vs --network (e.g. --fuji / --etna-devnet). This PR enable interchangeability of --cluster and --network flags.

## How this was tested
Should pass CI 

## How is this documented
NA

### DIFF
--- a/pkg/models/network.go
+++ b/pkg/models/network.go
@@ -4,10 +4,11 @@ package models
 
 import (
 	"fmt"
-	"github.com/ava-labs/avalanche-cli/pkg/utils"
-	"github.com/ava-labs/avalanchego/api/info"
 	"os"
 	"strings"
+
+	"github.com/ava-labs/avalanche-cli/pkg/utils"
+	"github.com/ava-labs/avalanchego/api/info"
 
 	"github.com/ava-labs/avalanche-cli/pkg/constants"
 	"github.com/ava-labs/avalanchego/genesis"


arturrez:
I feel like it adds a confusion if we do it, if I understand this PR correctly.
if we feel like `--cluster` flags is confusing users for blockchain deploy cmd, we should remove support for this flag and only support `--network` flag

pkg/models/network.go
```diff
@@ -75,6 +76,25 @@ func NewDevnetNetwork(endpoint string, id uint32) Network {
 	return NewNetwork(Devnet, id, endpoint, "")
 }
 
+// ConvertClusterToNetwork converts a cluster network into a non cluster network
+func ConvertClusterToNetwork(clusterNetwork Network) (Network, error) {
+	if clusterNetwork.ClusterName == "" {
+		return UndefinedNetwork, errors.New("no cluster network is provided")
+	}
+	switch {
```
felipemadero, line 84: probably you can just set cluster name to "" and return

pkg/node/node.go
```diff
@@ -3,12 +3,33 @@
...
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
 	"github.com/ava-labs/avalanche-cli/pkg/models"
 	"github.com/ava-labs/avalanche-cli/pkg/utils"
+	avagoconstants "github.com/ava-labs/avalanchego/utils/constants"
 )
 
+// GetNetworkFromCluster gets the network that a cluster is on
+func GetNetworkFromCluster(clusterConfig models.ClusterConfig) (models.Network, error) {
+	network := clusterConfig.Network
+	switch {
```
felipemadero, line 19: I believe this is models

felipemadero (commented):
I believe it is ok to add cluster name to sidecar network info.
But probably best would had been to directly change network.Name
instead of removing cluster name info from the network model. 
Also, there are functions that offer or try to offer cluster options for
a given blockchain, and that was based on network name on
sidecar, so now it should be based on cluster name info of
sidecar network info.

pkg/models/network.go
```diff
@@ -75,6 +76,25 @@ func NewDevnetNetwork(endpoint string, id uint32) Network {
...
+	case clusterNetwork.ID == constants.LocalNetworkID:
+		return NewLocalNetwork(), nil
+	case clusterNetwork.ID == avagoconstants.FujiID:
+		return NewFujiNetwork(), nil
+	case clusterNetwork.ID == avagoconstants.MainnetID:
+		return NewMainnetNetwork(), nil
+	case clusterNetwork.ID == constants.EtnaDevnetNetworkID:
+		return NewEtnaDevnetNetwork(), nil
+	default:
```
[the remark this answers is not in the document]
felipemadero: you can query the network id from the endpoint, such a networkoptions does

felipemadero (approved):
preaproving. will appreciate if you tackle the open conversion.

PR #2309, opened 2024-11-01
fix contexts and signals

felipemadero:
## Why this should be merged
- add partial sync and proposer vm height flags for local clusters
- fix diff contexts issues on local clusters (different context on client and server)
- fix context timeout for fuji
- recover cursos after interrupting a spinner with ctrlC


cmd/root.go
```diff
@@ -354,8 +357,18 @@ func initConfig() {
...
 	err := rootCmd.Execute()
 	cobrautils.HandleErrors(err)
 }
+
+func handleInterrupt() {
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+	sig := <-sigChan
+	fmt.Println()
+	fmt.Println("received signal:", sig.String())
```
arturrez, line 372: is it debug println? Pls feel free to ignore this comment

pkg/utils/common.go
```diff
@@ -73,6 +73,11 @@ func SplitKeyValueStringToMap(str string, delimiter string) (map[string]string,
 	return kvMap, nil
 }
 
+// Context for bootstrapping a Fuji Node
+func GetFujiBoostrappingContext() (context.Context, context.CancelFunc) {
```
arturrez, line 77: nit: have GetBootstrappingContext(network) where you can return context for switch/case for each network.
also typo in `GetFujiBoost`

arturrez: or change GetANRContext to add network param there

pkg/utils/common.go
```diff
@@ -589,3 +594,22 @@ func LogLevelToEmoji(logLevel string) (string, error) {
 	}
 	return levelEmoji, nil
 }
+
+// Set k=v in JSON string
+// e.g., "track-subnets" is the key and value is "a,b,c".
+func SetJSONKey(jsonBody string, k string, v interface{}) (string, error) {
```
arturrez, line 600: add unittest for this generic func

cmd/root.go
felipemadero: it is not debug println. but no other output mechanism (eg logging) is generally available here

pkg/utils/common.go
felipemadero: added a method to network

pkg/utils/common.go
felipemadero: added

pkg/node/local.go
```diff
@@ -180,6 +182,27 @@ func StartLocalNode(
 		localDataExists = false
 	}
 
+	// set node configs
```
michaelkaplan13, line 185: I think it would be cleaner for the caller of `StartLocalNode` to set the `nodeConfig` parameter properly for their use case rather than always overriding it with specific settings here. This allows the function to remain more general use for the future, for instance if we want to a start a local node that is going to become a primary network validator.

pkg/node/local.go
```diff
@@ -180,6 +182,27 @@ func StartLocalNode(
...
+	}
+	nodeConfig, err = utils.SetJSONKey(
+		nodeConfig,
+		config.PartialSyncPrimaryNetworkKey,
+		true,
+	)
+	if err != nil {
+		return err
+	}
+	nodeConfig, err = utils.SetJSONKey(
```
michaelkaplan13, line 197: Is `ProposerVMUseCurrentHeightKey` required here? I think it shouldn't be necessary with the ACP-151. If it is still required, could you add a comment explaining why?

pkg/node/local.go
michaelkaplan13: I think we should also only ever serialize the `config` struct to start a node, rather than creating a `Config` instance, serializing it as JSON, then deserializing it to override certain keys, etc. Using the AvalancheGo defined type will be less error prone in the event of any future changes to the config structure.

pkg/node/local.go
felipemadero: I just included it because it was not included previously and reminded the flag was asked to be set.
Will just not include as I am not pretending it is still required.

pkg/node/local.go
felipemadero: changed the Start() param to be a map[string]interface{} , and serialize it inside
I believe the avalanchego defined type (if it is the one define under avalanchego/node/config.go)
is not a good fit as we are just trying to set flags for the avalanchego command line, and it seems 
not direct to map from the internal Config type to the desired flags. Unless I am missing something.

cmd/blockchaincmd/deploy.go
```diff
@@ -534,14 +535,14 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
...
+				nodeConfig := map[string]interface{}{}
 				if app.AvagoNodeConfigExists(blockchainName) {
-					nodeConfigBytes, err := os.ReadFile(app.GetAvagoNodeConfigPath(blockchainName))
+					nodeConfig, err = utils.ReadJSON(app.GetAvagoNodeConfigPath(blockchainName))
 					if err != nil {
 						return err
 					}
-					nodeConfig = string(nodeConfigBytes)
 				}
+				nodeConfig[config.PartialSyncPrimaryNetworkKey] = true
```
michaelkaplan13, line 545: Taking a step back, is there a more general place for us to set the partial sync? 

We should set it for any node that will be an L1 validator only, no matter if that node is run locally or remotely. And conversely, if we want to set up a local node to be a primary network validator, then we won't want this flag set.

michaelkaplan13 (commented):
Left one question, and otherwise am just wonder where we need to set/check the `requirePrimaryNetworkSigners` flag for the `subnet-evm` config? Is that done in the `blockchain create` command?

felipemadero:
> Left one question, and otherwise am just wonder where we need to set/check the `requirePrimaryNetworkSigners` flag for the `subnet-evm` config? Is that done in the `blockchain create` command?

blockchain create forces(set) that setting for all subnet-evm confs generated using CLI. there is no check atm 
for the case the user provides a custom genesis.

cmd/blockchaincmd/deploy.go
felipemadero: I can add flags to `blockchain deploy` and `node local start` to specify if primary network validation is desired
(partial sync being default). that flags to be applied both to cluster creation, and to cluster restart. That following
current scheme.
I did not considered too much on generalities as remote node setup is currently under discussion, and fuji
primary setup is going to take other bootstrap time and was not a target.

felipemadero: but if generality is desired so, we can add either a boolean field for partial sync to the cluster definition struct (if we want to focus on that setting being a relevant part of our configuration), or a field for flags in general to the cluster definition struct (if we want to enable general flexible changes to a cluster definition). I tend to prefer later one as it enables changes in the future with less code changes (we just set/change flags).
in that case, previous to starting the local cluster for the first time, the cluster def is created in a previous, 
separate step. so probably the main function called from inside `blockchain deploy` and `node local start` will first call a func to create the cluster def and then a second func to execute the cluster. the second function, probably being also the function that restarts a cluster, will check if the cluster is partial sync or not, and set the correct flag.
if the use later on wants this cluster to be full sync, for example or restart, a flag should be set will should be used to change the cluster conf probably previous restart.

felipemadero: partial sync for a remote node should probably be target of another PR

michaelkaplan13: Sorry I'm not completely following your suggestions, but I think we should always strive to ensure things are as general/modular as possible. 🙂 

In this case, whether or not to set the `--partial-sync-primary-network` flag should be a determination that is made whenever we're starting/configuring an AvalancheGo node to be run, no matter where it's going to be executed.

I'm not exactly sure how that fits into the commands it should apply to, etc, but having it only apply to local nodes I think is likely to result in either duplicating it in the future in other place or needing to rework it to ensure it applies to all of the correct cases.

felipemadero: added cmdline flags for setting partial sync both for local and remote. as can be verified on https://github.com/ava-labs/avalanche-cli/pull/2309/commits/44b6dab72817a2eeddd92965f86c7069fa2ea057 the flows and ways to configure nodes are pretty different among local and remote. also remote has some issues related to node configuration (two different pieces need to be modified when changing something, one for start and another for restart). refactoring needed to fix and unify this most probably excess the scope of this PR.

felipemadero: cmdline flags are set for partial sync turned on by default

PR #2313, opened 2024-11-02
add node conf to acp-77-pos

felipemadero:
## Why this should be merged
adds node config to acp-77-pos, as suggested by ash team


felipemadero:
already implemented

PR #2314, opened 2024-11-04
make network start to not download new snapshots

felipemadero:
## Why this should be merged
An issue in avacloud devnets. When a new snapshot is released, and a devnet is restarted, the new snapshot 
is installed and the persistent state is lost.

## How this works
It ensures network start will not download new snapshots (but network clean and blockchain deploy will)

### DIFF
--- a/pkg/subnet/local.go
+++ b/pkg/subnet/local.go
@@ -42,16 +42,17 @@ import (
 )
 
 type LocalDeployer struct {
-	procChecker        binutils.ProcessChecker
-	binChecker         binutils.BinaryChecker
-	getClientFunc      getGRPCClientFunc
-	binaryDownloader   binutils.PluginBinaryDownloader
-	app                *application.Avalanche
-	backendStartedHere bool
-	setDefaultSnapshot setDefaultSnapshotFunc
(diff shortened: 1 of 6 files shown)


pkg/subnet/local.go
```diff
@@ -807,10 +823,10 @@ func SetDefaultSnapshot(snapshotsDir string, resetCurrentSnapshot bool, avagoVer
 		if err != nil {
 			ux.Logger.PrintToUser("Warning: failure verifying that the local snapshot is the latest one: %s", err)
 		} else if gotSum != expectedSum {
-			downloadSnapshot = true
+			checksumError = true
 		}
 	}
-	if downloadSnapshot {
+	if missingArchive || (checksumError && installUpdates) {
```
arturrez, line 829: should it be `missingArchive || checksumError || installUpdates`? if I understand logic correctly

arturrez (approved):

pkg/subnet/local.go
felipemadero: it is ok as it is. if the archive is not found, will install. if there is a new update (checksum error) and update install is enabled, will install.

Issue #2318, opened 2024-11-06
Unified `avalanche blockchain deploy` flow 


meaghanfitzgerald:
### Goal

Support PoS and PoA L1 deployment via `avalanche blockchain deploy`. 

### Details

On `acp-77` branch, the flow to create either a PoA or PoS L1 with the default Validator Manager Contracts should be the essentially identical. When a user is asked which mechanism they would like to use for validator management, their input will determine whether the genesis file is fed either PoA or PoS proxy-implementation contracts to be encoded. 

1. If PoA, the user should be prompted to provide an "owner" address that is required by the PoA contract for future validator addition and removal. 

2. PoS example supported by CLI should be [NativeTokenStakingManager](https://github.com/ava-labs/teleporter/blob/validator-manager-readme/contracts/validator-manager/NativeTokenStakingManager.sol). If PoS, the contract is given defaults for the following values:

- minimum stake amount
- maximum stake amount
- minimum stake duration
- minimum delegation fee
- maximum stake multiplier
- weight to value factor

### To Note

regardless of the implementation, the initial validators added as part of `ConvertSubnetTx` are treated as PoA and are [not eligible](https://github.com/ava-labs/teleporter/blob/validator-manager-readme/contracts/validator-manager/README.md#deploying) for staking rewards, so no additional prompting needed.

meaghanfitzgerald:
Reference implementation, [POC](https://github.com/ava-labs/avalanche-cli/pull/2301)

meaghanfitzgerald:
closed by https://github.com/ava-labs/avalanche-cli/pull/2140

PR #2321, opened 2024-11-07
ACP77 PoS Validator removal

arturrez:
## Why this should be merged
it adds removeValidator command
it also improves UX for falied tx traces as it display original error to user
## How this works
if Valildator is bootstrap Validatro it forces removal. POS uses more parameters for validatorRemoval. We rely on network for uptime calc for now
## How this was tested
added e2e. 
also
you can  use `test-defaults` and `ewoq` key everywhere for `etna-devnet` 
```
 ./bin/avalanche blockchain create pos7 --proof-of-stake  --force
 ./bin/avalanche blockchain deploy pos7
 ./bin/avalanche blockchain removeValidator pos7  --cluster pos7-local-node --node-endpoint http://127.0.0.1:9654 --key ewoq
 ```
## How is this documented
not yet


meaghanfitzgerald:
https://github.com/ava-labs/avalanche-cli/issues/2318

PR #2322, opened 2024-11-07
Refactor node provision

sukantoraymond:
## Why this should be merged
Separates node provisioning from node infrastructure creation in node create command. This will enable node provisioning function to be used in other packages.

New command node provision is also created, which will take the node ip and the node ssh key path to provision specified nodes with AvalancheGo and Avalanche CLI installed

## How this works
Modularize node provision out of node create

## How this was tested
Create cloud servers (e.g. AWS / GCP) without them being provisioned and provision them using node provision command. 

## How is this documented
NA

### DIFF
--- a/cmd/nodecmd/create.go
+++ b/cmd/nodecmd/create.go
@@ -231,10 +231,6 @@ func preCreateChecks(clusterName string) error {
 			return fmt.Errorf("invalid ip:port pair %s", ipPortPair)
 		}
 	}
-	if globalNetworkFlags.UseDevnet {
-		partialSync = false
-		ux.Logger.PrintToUser("disabling partial sync default for devnet")
-	}
 
 	return nil
 }
--- a/cmd/nodecmd/setup.go
+++ b/cmd/nodecmd/setup.go
@@ -53,10 +53,15 @@ Currently, only ubuntu-based operating system is supported.`,
 	cmd.Flags().BoolVar(&useSSHAgent, "use-ssh-agent", false, "use ssh agent(ex: Yubikey) for ssh auth")
 	cmd.Flags().StringVar(&genesisPath, "genesis", "", "path to genesis file")
 	cmd.Flags().StringVar(&upgradePath, "upgrade", "", "path to upgrade file")
+	cmd.Flags().BoolVar(&partialSync, "partial-sync", true, "primary network partial sync")
 	return cmd
 }
 
 func setup(hosts []*models.Host, avalancheGoVersion string, network models.Network) error {
+	if globalNetworkFlags.UseDevnet {
+		partialSync = false
+		ux.Logger.PrintToUser("disabling partial sync default for devnet")
+	}
 	ux.Logger.PrintToUser("Starting bootstrap process on the newly created Avalanche node(s)...")
 	wg := sync.WaitGroup{}
 	wgResults := models.NodeResults{}
@@ -93,6 +98,7 @@ func setup(hosts []*models.Host, avalancheGoVersion string, network models.Netwo
 				avalancheGoVersion,
 				bootstrapIDs,
 				bootstrapIPs,
+				partialSync,
 				genesisPath,
 				upgradePath,
 				addMonitoring,


cmd/nodecmd/create.go
```diff
@@ -694,124 +692,100 @@ func createNodes(cmd *cobra.Command, args []string) error {
...
 				if err = ssh.RunSSHSetupDockerService(monitoringHost); err != nil {
-					nodeResults.AddResult(monitoringHost.NodeID, nil, err)
+					nodeResults.AddResult(monitoringHost.IP, nil, err)
 					ux.SpinFailWithError(spinner, "", err)
 					return
 				}
 				ux.Logger.Info("SetupMonitoringEnv RunSSHSetupDockerService completed")
 				if err = ssh.RunSSHSetupMonitoringFolders(monitoringHost); err != nil {
-					nodeResults.AddResult(monitoringHost.NodeID, nil, err)
+					nodeResults.AddResult(monitoringHost.IP, nil, err)
```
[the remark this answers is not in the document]
meaghanfitzgerald: Why is create.go using IP if the node infrastructure is created on a local machine by default?

cmd/nodecmd/create.go
```diff
@@ -694,124 +692,100 @@ func createNodes(cmd *cobra.Command, args []string) error {
...
-				if err != nil {
-					nodeResults.AddResult(host.NodeID, nil, err)
-					ux.SpinFailWithError(spinner, "", err)
-					return
-				}
-				if err = ssh.RunSSHSetupPromtailConfig(host, monitoringNodeConfig.PublicIPs[0], constants.AvalanchegoLokiPort, cloudID, nodeID.String(), ""); err != nil {
-					nodeResults.AddResult(host.NodeID, nil, err)
-					ux.SpinFailWithError(spinner, "", err)
-					return
+		publicAccessToHTTPPort := slices.Contains(cloudConfigMap.GetAllAPIInstanceIDs(), host.GetCloudID()) || publicHTTPPortAccess
```
arturrez, line 747: as we are not using CloudIDs anymore we should rename and change logic GetAllAPIInstanceIDs to GetAllAPIInstanceIPs. 
host.GetCloudID is not needed

arturrez: we rely on IPs now

cmd/nodecmd/create.go
```diff
@@ -694,124 +692,100 @@ func createNodes(cmd *cobra.Command, args []string) error {
...
+	}
+	if addMonitoring {
+		spinSession := ux.NewUserSpinner()
+		for _, host := range hosts {
+			wg.Add(1)
+			go func(nodeResults *models.NodeResults, host *models.Host) {
+				defer wg.Done()
+				spinner := spinSession.SpinToUser(utils.ScriptLog(host.IP, "Add Monitoring"))
+				if addMonitoring {
+					cloudID := host.GetCloudID()
```
arturrez, line 761: we should remove host.GetCloudID() and change logic behind RunSSHSetupPromtailConfig and 
GetNodeInstanceDirPath(cloudID)

cmd/nodecmd/create.go
```diff
@@ -1043,18 +1016,29 @@ func generateNodeCertAndKeys(stakerCertFilePath, stakerKeyFilePath, blsKeyFilePa
 }
 
 func provideStakingCertAndKey(host *models.Host) error {
-	instanceID := host.GetCloudID()
-	keyPath := filepath.Join(app.GetNodesDir(), instanceID)
+	keyPath := filepath.Join(app.GetNodesDir(), "staking", host.IP)
```
arturrez, line 1019: I suggest we create a separate func that uses `keyPath := filepath.Join(app.GetNodesDir(), "staking", host.IP)` so IP is used by default when we want to save stuff. In terms of the reading stuff I suggest func that used `keyPath := filepath.Join(app.GetNodesDir(), "staking", host.IP)` but falls back to keyPath := filepath.Join(app.GetNodesDir(), "staking", NodeID) if this folder doesn't exists to support backward compatibility with nodes created with prev versions of CLI

cmd/nodecmd/create.go
```diff
@@ -1043,18 +1016,29 @@ func generateNodeCertAndKeys(stakerCertFilePath, stakerKeyFilePath, blsKeyFilePa
...
 	nodeID, err := generateNodeCertAndKeys(
 		filepath.Join(keyPath, constants.StakerCertFileName),
 		filepath.Join(keyPath, constants.StakerKeyFileName),
 		filepath.Join(keyPath, constants.BLSKeyFileName),
 	)
+	instanceID := host.GetCloudID()
```
arturrez, line 1025: we should depricate host.GetCloudID()

cmd/nodecmd/setup.go
```diff
@@ -0,0 +1,257 @@
...
+				ux.SpinFailWithError(spinner, "", err)
+				return
+			}
+			ux.SpinComplete(spinner)
+		}(&wgResults, host)
+	}
+	wg.Wait()
+	spinSession.Stop()
+	for _, node := range hosts {
+		if wgResults.HasNodeIDWithError(node.NodeID) {
```
arturrez, line 110: HasNodeIPWithError 
we should not use NodeID as we only have NodeIP now

cmd/nodecmd/setup.go
```diff
@@ -0,0 +1,257 @@
...
+	if err = setup(hosts, avalancheGoVersion, network); err != nil {
+		return err
+	}
+	printSetupResults(hosts)
+	return nil
+}
+
+func printSetupResults(hosts []*models.Host) {
+	for _, host := range hosts {
+		nodePath := filepath.Join(app.GetNodesDir(), "staking", host.IP)
```
arturrez, line 213: should be abstracted so we can hide if we access staking data from IP or NodeID(for backwards compatibility

cmd/nodecmd/setup.go
```diff
@@ -0,0 +1,257 @@
...
+	cmd.Flags().BoolVar(&publicHTTPPortAccess, "public-http-port", false, "allow public access to avalanchego HTTP port")
+	cmd.Flags().StringArrayVar(&nodeIPs, "node-ips", []string{}, "IP addresses of nodes")
+	cmd.Flags().StringArrayVar(&sshKeyPaths, "ssh-key-paths", []string{}, "ssh key paths")
+	cmd.Flags().BoolVar(&useSSHAgent, "use-ssh-agent", false, "use ssh agent(ex: Yubikey) for ssh auth")
+	cmd.Flags().StringVar(&genesisPath, "genesis", "", "path to genesis file")
+	cmd.Flags().StringVar(&upgradePath, "upgrade", "", "path to upgrade file")
+	return cmd
+}
+
+func setup(hosts []*models.Host, avalancheGoVersion string, network models.Network) error {
```
arturrez, line 59: as we exposed this function to the users, we need to check if provided slice of hosts in compatible with CLI
i.e it has ubuntu user and it's supported OS
i.e it has list of pkgs needed 

I would expect people to input weird stuff Ip/ssh key for some user and redhat box, or raspery pi or whatever.

cmd/nodecmd/setup.go
```diff
@@ -0,0 +1,257 @@
...
+		numNodes, err = app.Prompt.CaptureInt(
+			"How many Avalanche nodes do you want to setup?",
+			prompts.ValidatePositiveInt,
+		)
+	}
+	if err != nil {
+		return err
+	}
+	for len(nodeIPs) < numNodes {
+		ux.Logger.PrintToUser("Getting info for node %d", len(nodeIPs)+1)
```
arturrez, line 241: we need to tell the user what OS we expect

cmd/nodecmd/setup.go
```diff
@@ -0,0 +1,257 @@
...
+		if err != nil {
+			return err
+		}
+		sshKeyPath, err := app.Prompt.CaptureString("What is the key path of the private key that can be used to ssh into this node?")
+		if err != nil {
+			return err
+		}
+		nodeIPs = append(nodeIPs, ipAddress)
+		sshKeyPaths = append(sshKeyPaths, sshKeyPath)
+		ux.Logger.GreenCheckmarkToUser("Bootstrap Validator %d:", len(nodeIPs))
```
arturrez, line 252: bootstrap validator?

tests/e2e/testcases/node/create/suite.go
```diff
@@ -46,7 +47,9 @@ var _ = ginkgo.Describe("[Node create]", func() {
 		re := regexp.MustCompile(`Generated staking keys for host (\S+)\[NodeID-(\S+)\]`)
 		match := re.FindStringSubmatch(output)
 		if len(match) >= 3 {
-			hostName = match[1]
+			var err error
```
arturrez, line 50: do we want to  provide basic e2e for new setup cmd?

tests/e2e/utils/helpers.go
```diff
@@ -1071,3 +1073,12 @@ func GetKeyTransferFee(output string) (uint64, error) {
 func GetAPILargeContext() (context.Context, context.CancelFunc) {
 	return context.WithTimeout(context.Background(), constants.APIRequestLargeTimeout)
 }
+
+func GetE2EHostInstanceID() (string, error) {
+	hosts, err := ansible.GetInventoryFromAnsibleInventoryFile(path.Join(GetBaseDir(), constants.NodesDir, constants.AnsibleInventoryDir, constants.E2EClusterName))
+	if err != nil {
+		return "", err
+	}
+	_, cloudHostID, _ := models.HostAnsibleIDToCloudID(hosts[0].NodeID)
```
arturrez, line 1082: as we are only getting IP address/ SSH key pairs from the user we should not use CloudID and use IP addresses instead

arturrez (commented):
I suggest switching to avalanchego NodeID from staking cert to replace CloudID everywhere as `node setup` doesn't require cloud nodes anymore and can be some bare metal

cmd/nodecmd/setup.go
arturrez: I think even better way is to depricate ansible nodeiD in favour of NodeID from avalanchego staking cert.
We will need a special nodeID for monitoring box as it doesn't have avalanchego

cmd/nodecmd/create.go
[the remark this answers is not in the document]
arturrez: I suggest replacing ansible nodeiD with avalanchego NodeID

meaghanfitzgerald:
Does this also conflict with #2337 ? @sukantoraymond

PR #2348, opened 2024-11-20
Cleanup commands

sukantoraymond:
Cleanup unused cli commands


felipemadero (commented):
I believe delete is useful. Maybe validators implementatio. can be added to describe instead of just delete it. e2e needs fixes

PR #2351, opened 2024-11-21
Install binary for avalanchego instead of providing local path

sukantoraymond:
## Why this should be merged
Currently we are asking for avalanche go binary path. Since avalanchego release currently contains binary, we can just download from git.

### DIFF
--- a/tests/e2e/testcases/subnet/sov/etna/suite.go
+++ b/tests/e2e/testcases/subnet/sov/etna/suite.go
@@ -31,7 +31,8 @@ func createEtnaSubnetEvmConfig(poa, pos bool) {
 	gomega.Expect(err).Should(gomega.BeNil())
 	gomega.Expect(exists).Should(gomega.BeFalse())
 
-	cmdArgs := []string{"blockchain",
+	cmdArgs := []string{
+		"blockchain",
 		"create",
 		subnetName,
 		"--evm",


cmd/blockchaincmd/deploy.go
```diff
@@ -599,13 +599,18 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
 				avagoVersionSettings := node.AvalancheGoVersionSettings{}
 				useEtnaDevnet := network.Kind == models.EtnaDevnet
 				if avagoBinaryPath == "" {
-					ux.Logger.PrintToUser("Local build of Avalanche Go is required to create an Avalanche node using local machine")
-					ux.Logger.PrintToUser("Please download Avalanche Go repo at https://github.com/ava-labs/avalanchego and build from source through ./scripts/build.sh")
-					ux.Logger.PrintToUser("Please provide the full path to Avalanche Go binary in the build directory (e.g, xxx/build/avalanchego)")
-					avagoBinaryPath, err = app.Prompt.CaptureString("Path to Avalanche Go build")
+					avaGoVersionSetting := node.AvalancheGoVersionSettings{
+						UseLatestAvalanchegoPreReleaseVersion: true,
```
arturrez, line 603: do we need to include `userProvidedAvagoVersion` into the `node.AvalancheGoVersionSettings`?

cmd/nodecmd/create.go
```diff
@@ -137,8 +137,15 @@ will apply to all nodes in the cluster`,
 func handlePostRun(_ *cobra.Command, _ []string) {}
 
 func preCreateChecks(clusterName string) error {
-	if !flags.EnsureMutuallyExclusive([]bool{useLatestAvalanchegoReleaseVersion, useLatestAvalanchegoPreReleaseVersion, useAvalanchegoVersionFromSubnet != "", useCustomAvalanchegoVersion != ""}) {
-		return fmt.Errorf("latest avalanchego released version, latest avalanchego pre-released version, custom avalanchego version and avalanchego version based on given subnet, are mutually exclusive options")
+	if useCustomAvalanchegoVersion != "" || useAvalanchegoVersionFromSubnet != "" {
```
arturrez, line 140: I am confused with this one. Why do we want to set flags to false `useLatestAvalanchegoReleaseVersion` and `useLatestAvalanchegoPreReleaseVersion` as user input them and we silently change them to something else based on other flags VS how it was before and we just return error to user if we tries to use mutually exclusive flags

cmd/nodecmd/create.go
```diff
@@ -137,8 +137,15 @@ will apply to all nodes in the cluster`,
...
+		useLatestAvalanchegoReleaseVersion = false
+		useLatestAvalanchegoPreReleaseVersion = false
+	}
+	if !flags.EnsureMutuallyExclusive([]bool{useLatestAvalanchegoReleaseVersion, useLatestAvalanchegoPreReleaseVersion}) {
```
arturrez, line 144: why do we allow users to use useLatestAvalanchegoReleaseVersion and useCustomAvalanchegoVersion at the same time ?

cmd/nodecmd/create.go
arturrez: I guess I understand now why, but we should display user a message explaining why we ignore their input.

cmd/nodecmd/create.go
arturrez: pls ignore comment above. We need to tell users that `useLatestAvalanchegoReleaseVersion` and `useLatestAvalanchegoPreReleaseVersion` have lower priority than `useLatestAvalanchegoReleaseVersion` and `useCustomAvalanchegoVersion` and will be override.

pkg/node/local.go
```diff
@@ -208,9 +208,13 @@ func StartLocalNode(
 		avalancheGoVersion, err = GetAvalancheGoVersion(app, avaGoVersionSetting)
 		if err != nil {
 			return err
-		} else {
-			ux.Logger.PrintToUser("Using AvalancheGo version: %s", avalancheGoVersion)
 		}
+		_, avagoDir, err := binutils.SetupAvalanchego(app, avalancheGoVersion)
```
arturrez, line 212: nit: I checked all usage of `binutils.SetupAvalanchego` and avagoDir is always used to only to get path to avalanchego. Can we change this func to return path to avalanchego right away and not dir ?

tests/e2e/testcases/subnet/sov/etna/suite.go
```diff
@@ -25,7 +25,7 @@ const (
 	testLocalNodeName = "e2eSubnetTest-local-node"
 )
 
-func createEtnaSubnetEvmConfig() {
+func createEtnaSubnetEvmConfigPOA() {
```
arturrez, line 28: can we just pass variable to the `createEtnaSubnetEvmConfig()` instead so we can chose between "--proof-of-authority" and "--proof-of-stake"  so we can keep single func to create blockchain config VS having 2 very similar func( 1 line difference)

arturrez: another option is to use `CreateEtnaSubnetEvmConfig` from `tests/e2e/commands/etna.go` 
I think it already supports both PoS and PoA usecase

cmd/nodecmd/local.go
```diff
@@ -147,11 +149,14 @@ func localStartNode(_ *cobra.Command, args []string) error {
 		StakingCertKeyPath:   stakingCertKeyPath,
 		StakingTLSKeyPath:    stakingTLSKeyPath,
 	}
+	if useCustomAvalanchegoVersion != "" {
```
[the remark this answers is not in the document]
arturrez: it could be confusing for the user to get not what he set via cmd line.
I don't have strong opinion on this, so I would just resolve it

pkg/binutils/release.go
```diff
@@ -96,5 +96,5 @@ func InstallBinary(
 
 	binDir, err := installBinaryWithVersion(app, version, installDir, binPrefix, downloader, installer)
 
-	return version, binDir, err
+	return version, filepath.Join(binDir, "avalanchego"), err
```
arturrez, line 99: this func in used to install evm, so it should return binDir not avalanchego path

arturrez: I was talking about SetupAvalanchego func change not InstallBinary func

tests/e2e/testcases/subnet/sov/etna/suite.go
```diff
@@ -25,29 +25,32 @@ const (
 	testLocalNodeName = "e2eSubnetTest-local-node"
 )
 
-func createEtnaSubnetEvmConfig() {
+func createEtnaSubnetEvmConfig(poa, pos bool) {
```
arturrez, line 28: this one I am not sure about.
Do we need 2 boolean vars? 

I think we assume that if not PoA, it's PoS 
so only 1 boolean var should be sufficient imho

pls feel free to ignore this comment :)

arturrez (approved):

PR #2354, opened 2024-11-21
Custom endpoint etna

sukantoraymond:
## Why this should be merged
When users enters https://etna.avax-dev.network/ in custom endpoint for network prompt, it should lead to etna


arturrez (approved):
I guess this is the fix for something.
I feel like we are mixing up Devnet and EtnaDevnet somewhere

PR #2356, opened 2024-11-22
Filter network

sukantoraymond:
## Why this should be merged

Filters list of network options to network that subnet is deployed on based on sidecar.json

```
AVL-7H2W7V:avalanche-cli raymondsukanto$ ./bin/avalanche subnet addValidator poaSubent
  ▸ Etna Devnet
```

### DIFF
--- a/pkg/networkoptions/network_options.go
+++ b/pkg/networkoptions/network_options.go
@@ -257,6 +257,7 @@ func GetNetworkFromCmdLineFlags(
 		}
 	}
 	// unsupported option
+	// allow cluster because we can extract underlying network from cluster
 	if networkOption != Undefined && !slices.Contains(supportedNetworkOptions, networkOption) && networkOption != Cluster {
 		errMsg := fmt.Errorf("network flag %s is not supported. use one of %s", networkFlagsMap[networkOption], supportedNetworksFlags)
 		if subnetName != "" {


pkg/networkoptions/network_options.go
```diff
@@ -64,6 +64,12 @@ func NetworkOptionFromString(s string) NetworkOption {
 		return Devnet
 	case "Cluster":
 		return Cluster
+	default:
+		if strings.Contains(s, "Devnet") {
```
arturrez, line 68: replace case Devnet with this. no need to put it in default

pkg/networkoptions/network_options.go
```diff
@@ -64,6 +64,12 @@ func NetworkOptionFromString(s string) NetworkOption {
...
+			return Devnet
+		} else if strings.Contains(s, "Cluster") {
```
arturrez, line 70: replace case "Cluster" with this. No need to add to default

cmd/blockchaincmd/add_validator.go
```diff
@@ -128,18 +129,36 @@ func addValidator(_ *cobra.Command, args []string) error {
 		return err
 	}
 
+	sc, err := app.LoadSidecar(blockchainName)
```
arturrez, line 132: this is the same code 3 times. Please move it to separate func maybe in utils or in app.
This is very useful functionality and might be even expanded later to more commads.

arturrez (commented):
this is great idea 🙌 
will approve this after comments addressed

pkg/networkoptions/network_options.go
```diff
@@ -64,6 +64,12 @@ func NetworkOptionFromString(s string) NetworkOption {
...
+			return Cluster
+		}
```
arturrez, line 72: I mean something like 
```
func NetworkOptionFromString(s string) NetworkOption {
	switch {
	case s == "Mainnet":
		return Mainnet
	case s == "Fuji Testnet":
		return Fuji
	case s == "Local Network":
		return Local
	case s == "Etna Devnet":
		return EtnaDevnet
	case s == "Devnet" || strings.Contains(s, "Devnet"):
		return Devnet
	case s == "Cluster" || strings.Contains(s, "Cluster"):
		return Cluster
	default:
		return Undefined
	}
}
```

arturrez: i just built it with this code with no issues 
image

arturrez (approved):
pls address https://github.com/ava-labs/avalanche-cli/pull/2356/files#r1854495349 
pre-approving

arturrez (approved):

PR #2369, opened 2024-11-25
Add HyperSDK Functionality to CLI

RodrigoVillar:
## Why this should be merged

With the HyperSDK maturing, we would like to make it a first-class citizen (i.e. the same level of support as Subnet-EVM). By doing this, HyperSDK developers benefit from the following:

- Fast Deployments: by being able to use the Avalanche-CLI, developers no longer have to start up their VM via integration/e2e tests
- Learning: using the Avalanche-CLI is a lower barrier-of-entry for potential HyperSDK developers than the current test framework

## How this works

This PR separates HyperVMs (VMs built with HyperSDK) from the custom VM group and makes it a group of its own. Therefore, we now have three types of VMs in the CLI:

- SubnetEVM
- HyperVMs
- Custom VMs

The biggest change is in the `avalanche blockchain create` command, where users are now able to select the option of creating a HyperVM. When selecting this option, the control flow is similar to that of Subnet-EVM, with the exception of the genesis.

In addition to being able to provide a `genesis.json` file and being able to use the default HyperSDK genesis, users can also use their VM binary to create the VM genesis on their behalf. 

## How is this documented

We plan on adding documentation for this feature in the HyperSDK docs.

cc: @felipemadero @aaronbuchwald


cmd/blockchaincmd/create.go
```diff
@@ -367,7 +372,55 @@ func createBlockchainConfig(cmd *cobra.Command, args []string) error {
...
+			defaultGen := "Use the DefaultGenesis for my VM"
+			option, err := app.Prompt.CaptureList(
+				"How would you like to provide the genesis file?",
+				[]string{providePath, binaryGen, defaultGen},
+			)
+			if err != nil {
+				return err
+			}
+			switch option {
+			case providePath:
```
arturrez, line 388: no need to add another prompt to ask for path. better UX is to just ask user to provide path to genesis right away and use it after checking if file exists. So this case branch is not needed here  and can be moved  up to `How would you like to provide the genesis file?` prompt

pkg/vm/create_hypersdk.go
```diff
@@ -0,0 +1,206 @@
...
+					})
+				}
+				return customAllocs, nil
+			}
+		}
+	}
+	return nil, nil
+}
+
+func getChainID(app *application.Avalanche) (ids.ID, error) {
```
arturrez, line 172: nit: better name `promptChainID`

pkg/vm/create_hypersdk.go
```diff
@@ -0,0 +1,206 @@
...
+	genesis.Rules.MaxBlockUnits = fees.Dimensions{1800000, math.MaxUint64, math.MaxUint64, math.MaxUint64, math.MaxUint64}
+	genesis.Rules.MinBlockGap = 100
+
+	genesis.Rules.NetworkID = uint32(1)
+	genesis.Rules.ChainID = chainID
+
+	return json.Marshal(genesis)
+}
+
+func getPreallocations(app *application.Avalanche) ([]*genesis.CustomAllocation, error) {
```
arturrez, line 87: nit: better name promptPreallocations

arturrez (commented):
commented on the code

arturrez:
looking into the code it's not clear for me what the difference between customVM and hypersdk VM option. 
I guess the only difference is ability to create a genesis data by running a command.

PR #2373, opened 2024-11-25
add etnaActivationTime

arturrez:
## Why this should be merged
provides correct etna activation time for devnet and fuji

## How this works
uses map and default 0,0 value
## How this was tested
e2e regulat etna flow 
## How is this documented
n/a

### DIFF
--- a/sdk/interchain/signature-aggregator.go
+++ b/sdk/interchain/signature-aggregator.go
@@ -8,11 +8,12 @@ import (
 	"os"
 	"time"
 
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
 	"github.com/ava-labs/avalanche-cli/pkg/models"
 	"github.com/ava-labs/avalanchego/api/info"
 	"github.com/ava-labs/avalanchego/ids"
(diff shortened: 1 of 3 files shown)


sdk/interchain/signature-aggregator.go
```diff
@@ -153,7 +153,11 @@ func NewSignatureAggregator(
 			logging.JSON.ConsoleEncoder(),
 		),
 	)
-	return initSignatureAggregator(peerNetwork, logger, registerer, subnetID, quorumPercentage)
+	etnaTime := time.Unix(0, 0)
+	if t, ok := constants.EtnaActivationTime[network.ID]; ok {
+		etnaTime = t
+	}
```
michaelkaplan13, line 159:
```suggestion
	etnaTime := constants.EtnaActivationTime[network.ID]
```

This will use a default (i.e. zero) `time.Time` instance if the key is not in the map

arturrez: totally agree. fixed

michaelkaplan13 (approved):

Issue #2334, opened 2024-11-14
[Bug]: Unable to create blockchain if previous attempt failed


RodrigoVillar:
**Describe the bug**

When using `avalanche blockchain create`, if the commands fails, then running the command again with the same blockchain name is not possible because the corresponding folder in `.avalanche-cli/subnets` already exists, but is malformed.

**To Reproduce**

1. Via the `README`, install Avalanche-CLI using the following:
```bash
curl -sSfL https://raw.githubusercontent.com/ava-labs/avalanche-cli/main/scripts/install.sh | sh -s
```
2. Call `avalanche blockchain create morpheusvm`
3.  For the genesis, use an empty JSON file
4. For the VM binary, use the MorpheusVM binary
5. After the prior two steps, you should get an error relating to a field in the genesis not being specified. This is fine.
6. Call `avalanche blockchain create morpheusvm`. This should result in an error

**Expected behavior**

I expected for the CLI to clean up the malformed folder, thus allowing me to execute the command again.

**Screenshots**

**Logs**

**Operating System**

macOS 14.5

**Additional context**
To create the binary for MorpheusVM, go to the HyperSDK repo and run the following in `examples/morpheusvm`:
```bash
go build -o ./build/morpheusvm ./cmd/morpheusvm
```

felipemadero:
for more context, the particular error message is `subnet %s has inconsistent configuration. there is no %s file present on directory %s. please backup any file and then remove the subnet`, has can happen if the subnet directory was created but by failure the sidecar not.

felipemadero:
working on a fix so automatically clean up is made by CLI

PR #2380, opened 2024-11-28
remove inconsistent blockchains found on migration

felipemadero:
## Why this should be merged
Closes https://github.com/ava-labs/avalanche-cli/issues/2334

### DIFF
--- a/sdk/utils/file.go
+++ b/sdk/utils/file.go
@@ -16,6 +16,15 @@ func FileExists(filename string) bool {
 	return !info.IsDir()
 }
 
+// DirExists checks if a directory exists.
+func DirExists(dirName string) bool {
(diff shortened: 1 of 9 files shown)


sdk/utils/file.go
```diff
@@ -16,6 +16,15 @@ func FileExists(filename string) bool {
 	return !info.IsDir()
 }
 
+// DirExists checks if a directory exists.
+func DirExists(dirName string) bool {
```
arturrez, line 20: there is existing `DirectoryExists`

arturrez (approved):

sdk/utils/file.go
felipemadero: Replacing its usage with sdk's one

Issue #2415, opened 2024-12-07
[Bug]: Deploying an L1 on a local network fails due to failure with signature aggregation


meaghanfitzgerald:
`avalanche blockchain deploy PoA1206 --local` is failing

```zsh
avalanche blockchain deploy PoA1206 --local
Deploying [PoA1206] to Local Network

Backend controller started, pid: 10007, output at: /Users/meag.fitz/.avalanche-cli/runs/server_20241206_220748/avalanche-cli-backend.log

Booting Network. Wait until healthy...

Node logs directory: /Users/meag.fitz/.avalanche-cli/runs/network_20241206_220751/node<i>/logs

Network ready to use.

Using [P-custom18jma8ppw3nhx5r4ap8clazz0dps7rv5u9xde7p] to be set as a change owner for leftover AVAX
Your blockchain control keys: [P-custom18jma8ppw3nhx5r4ap8clazz0dps7rv5u9xde7p]
Your subnet auth keys for chain creation: [P-custom18jma8ppw3nhx5r4ap8clazz0dps7rv5u9xde7p]
CreateSubnetTx fee: 0.000010278 AVAX
Subnet has been created with ID: GEieSy2doZ96bpMo5CuHPaX1LvaxpKZ9C72L22j94t6YyUb6X
Now creating blockchain...
CreateChainTx fee: 0.000095890 AVAX
+--------------------+----------------------------------------------------+
| DEPLOYMENT RESULTS |                                                    |
+--------------------+----------------------------------------------------+
| Chain Name         | PoA1206                                            |
+--------------------+----------------------------------------------------+
| Subnet ID          | GEieSy2doZ96bpMo5CuHPaX1LvaxpKZ9C72L22j94t6YyUb6X  |
+--------------------+----------------------------------------------------+
| VM ID              | cRbRuBMaFjReUaJufdz6HbyXHomh4gYxEMshKSA1H2ZwbxBFT  |
+--------------------+----------------------------------------------------+
| Blockchain ID      | 2ajYJXeqU9xVojmCEgsEXRaCZ5tqLKdC8uPLRRFgH9PSVYoyYU |
+--------------------+                                                    +
| P-Chain TXID       |                                                    |
+--------------------+----------------------------------------------------+
Now calling ConvertSubnetToL1Tx...
ConvertSubnetToL1Tx fee: 0.000053832 AVAX
[...]

PR #2422, opened 2024-12-10
Fix relayer url issue

felipemadero:
## Why this should be merged
- Closes https://github.com/ava-labs/avalanche-cli/issues/2415
- Enables downloading separate releases for relayer and signature aggregator from icm services
 
## How this works
- Ignore version format inconsistences on icm services (to be refined when version format is firm). And
ensures latest relayer is used.
- Enable release downloader to receive org, repo and tool name

## How this was tested
- CI


felipemadero:
> Should we have e2e that tests the issue described in #2415

We have certain conventions on the format for relayer versions. Really it is not a CLI issue it is expected
for it to fail if a relayer version is not correct, so as to proceed with proper fixes. Lets do a unit test
once the new relayer version format is fixed.

PR #2427, opened 2024-12-11
Enable mainnet

sukantoraymond:
Enable mainnet when it activates


felipemadero (commented):
add a ux comments telling the user on the time needed to bootstrap mainnet

felipemadero (approved):

PR #2435, opened 2024-12-12
Get balance

sukantoraymond:
## Why this should be merged
avalance validator getBalance gets balance of validator of L1

## How this works
Lets user choose whether to get balance of bootstrap validator (read from sidecar) or validator added after convertSubnetTx (using validatorID). 

Once this PR is merged https://github.com/ava-labs/avalanchego/pull/3564, we can use node id to get the node's balance, without requiring validation ID.

## How this was tested
./bin/avalanche validator getBalance 

## How is this documented
NA

### DIFF
--- a/cmd/blockchaincmd/deploy.go
+++ b/cmd/blockchaincmd/deploy.go
@@ -204,7 +204,7 @@ so you can take your locally tested Subnet and deploy it on Fuji or Mainnet.`,
 	cmd.Flags().Float64Var(
 		&deployBalanceAVAX,
 		"balance",
-		constants.BootstrapValidatorBalance/float64(units.Avax),
+		float64(constants.BootstrapValidatorBalanceNanoAVAX)/float64(units.Avax),
 		"set the AVAX balance of each bootstrap validator that will be used for continuous fee on P-Chain",
 	)
 	cmd.Flags().IntVar(&numLocalNodes, "num-local-nodes", 0, "number of nodes to be created on local machine")
--- a/cmd/blockchaincmd/add_validator.go
+++ b/cmd/blockchaincmd/add_validator.go
@@ -409,9 +409,10 @@ func CallAddValidator(
 		if err != nil {
 			return err
 		}
+	} else {
+		// convert to nanoAVAX
+		balance *= units.Avax
 	}
-	// convert to nanoAVAX
-	balance *= units.Avax
 
 	if remainingBalanceOwnerAddr == "" {
 		remainingBalanceOwnerAddr, err = getKeyForChangeOwner(network)
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -4,6 +4,8 @@ package constants
 
 import (
 	"time"
+
+	"github.com/ava-labs/avalanchego/utils/units"
 )
 
 type HTTPAccess bool
@@ -119,7 +121,9 @@ const (
 	TimeParseLayout = "2006-01-02 15:04:05"
 	MinStakeWeight  = 1
 	// Default balance when we prompt users for bootstrap validators
-	BootstrapValidatorBalance = 1000000000
+	// nAVAX
+	BootstrapValidatorBalanceNanoAVAX = uint64(BootstrapValidatorBalanceAVAX * float64(units.Avax))
+	BootstrapValidatorBalanceAVAX     = 0.1
 	// Default weight when we prompt users for bootstrap validators
 	BootstrapValidatorWeight = 100
 	// Default weight when we prompt users for non bootstrap validators
--- a/pkg/prompts/prompts.go
+++ b/pkg/prompts/prompts.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ava-labs/avalanchego/utils/units"
+
 	"github.com/ava-labs/avalanche-cli/pkg/constants"
 	"github.com/ava-labs/avalanche-cli/pkg/key"
 	"github.com/ava-labs/avalanche-cli/pkg/models"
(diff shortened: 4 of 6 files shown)


cmd/blockchaincmd/deploy.go
```diff
@@ -1174,6 +1176,22 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func setBootstrapValidatorValidationID(avaGoBootstrapValidators []*txs.ConvertSubnetToL1Validator, bootstrapValidators []models.SubnetValidator, subnetID ids.ID) error {
+	for index, avagoValidator := range avaGoBootstrapValidators {
+		for bootstrapValidatorIndex, validator := range bootstrapValidators {
```
arturre, line 1181: nit: you can use utils.MapWithError for this

cmd/blockchaincmd/deploy.go
```diff
@@ -1174,6 +1176,22 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func setBootstrapValidatorValidationID(avaGoBootstrapValidators []*txs.ConvertSubnetToL1Validator, bootstrapValidators []models.SubnetValidator, subnetID ids.ID) error {
+	for index, avagoValidator := range avaGoBootstrapValidators {
```
arturrez, line 1180: you can use utils.MapWithError for this.
or utils.Map if error can be ignored.

imho this error doesn't need to be raised up and just skipped as we try to match one thing to another so any invalid nodeID can be ignored as it would not match

cmd/blockchaincmd/deploy.go
```diff
@@ -925,7 +925,9 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
 			return err
 		}
 		ux.Logger.PrintToUser("")
-
+		if err = setBootstrapValidatorValidationID(avaGoBootstrapValidators, bootstrapValidators, subnetID); err != nil {
```
arturrez, line 928: question: do we need to do the same after we add validator? 
also it's possible if the same validator will be removed and added back.

cmd/validatorcmd/getBalance.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+		balance, err = txutils.GetValidatorPChainBalanceValidationID(network, validationID)
+		if err != nil {
+			return err
+		}
+		ux.Logger.PrintToUser("  Validator Balance: %.5f", float64(balance)/float64(units.Avax))
+		return nil
+	}
+
+	isBootstrapValidator, err := app.Prompt.CaptureYesNo("Is the validator a bootstrap validator?")
```
arturrez, line 78: you don't need to ask user this. This info is available in sidecar

cmd/validatorcmd/getBalance.go
```diff
@@ -0,0 +1,128 @@
...
+		Short: "Gets current balance of validator on P-Chain",
+		Long: `This command gets the remaining validator P-Chain balance that is available to pay
+P-Chain continuous fee`,
+		RunE: getBalance,
+		Args: cobrautils.ExactArgs(0),
+	}
+
+	networkoptions.AddNetworkFlagsToCmd(cmd, &globalNetworkFlags, true, getBalanceSupportedNetworkOptions)
+	cmd.Flags().StringVar(&l1, "l1", "", "name of L1 (to get balance of bootstrap validators only)")
+	cmd.Flags().StringVar(&subnetID, "subnet-id", "", "subnetID of L1 that the node is validating")
```
arturrez, line 45: this flag is not used in the func if I read this func correctly

cmd/validatorcmd/getBalance.go
```diff
@@ -0,0 +1,128 @@
...
+	cmd.Flags().StringVar(&validationIDStr, "validation-id", "", "validationIDStr of the validator")
+	return cmd
```
arturrez, line 47: do we want to allow user to specify NodeID ? it seems logical

cmd/validatorcmd/getBalance.go
```diff
@@ -0,0 +1,128 @@
...
+		Use:   "getBalance",
+		Short: "Gets current balance of validator on P-Chain",
+		Long: `This command gets the remaining validator P-Chain balance that is available to pay
+P-Chain continuous fee`,
+		RunE: getBalance,
+		Args: cobrautils.ExactArgs(0),
+	}
+
+	networkoptions.AddNetworkFlagsToCmd(cmd, &globalNetworkFlags, true, getBalanceSupportedNetworkOptions)
+	cmd.Flags().StringVar(&l1, "l1", "", "name of L1 (to get balance of bootstrap validators only)")
```
arturrez, line 44: this one is confusing to me.

It seems to me that we should just provide a separate command to list bootstrap validators by l1/blockchain name based on the sidecar info. And display their nodeID/validation ID pairs.

this func would just accept nodeiD or validationID to display a balance. I feel like it would give us clean and predictable UX and less prompts and checks. Like we would not fail if `l1` is not provided.

arturrez: if no validatorID/NodeID provided - fail fast

pkg/txutils/info.go
```diff
@@ -111,3 +111,13 @@ func GetOwners(network models.Network, subnetID ids.ID) (bool, []string, uint32,
 	}
 	return isPermissioned, controlKeysStrs, threshold, nil
 }
+
+func GetValidatorPChainBalanceValidationID(network models.Network, validationID ids.ID) (uint64, error) {
```
felipemadero, line 115: why is this in txutils? probably move it to utils

felipemadero (approved):
LGTM. Preapproving

cmd/blockchaincmd/deploy.go
arturrez: if you don't need err, you should be able to use utils.Map
just a nit, pls feel free to ignore it.

PR #2448, opened 2024-12-14
Hardcode icm version

felipemadero:
## Why this should be merged
blockchain create is not working because latest release of icm-contracts has unexpected version format

## How this works
It hardcodes needed icm version v1.0.0 instead of downloading latest release


michaelkaplan13 (approved):
This should be fine as a temporary fix, but we should evaluate how to define dependency versions in a more standardized and robust way going forward.

PR #2455, opened 2024-12-16
validate pos L1 duration

arturrez:
## Why this should be merged
adds L1 min staking duration for PoS as 24hrs

### DIFF
--- a/cmd/blockchaincmd/add_validator.go
+++ b/cmd/blockchaincmd/add_validator.go
@@ -376,7 +376,7 @@ func CallAddValidator(
 			}
 		}
 		if duration == 0 {
-			duration, err = PromptDuration(time.Now(), network)
+			duration, err = PromptDuration(time.Now(), network, pos)
 			if err != nil {
 				return nil
 			}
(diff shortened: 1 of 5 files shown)


pkg/prompts/validations.go
```diff
@@ -58,6 +58,21 @@ func validateMainnetStakingDuration(input string) error {
...
+	const minL1StakingDuration = 24 * time.Hour
+	d, err := time.ParseDuration(input)
+	if err != nil {
+		return err
+	}
+	if d > genesis.MainnetParams.MaxStakeDuration {
+		return fmt.Errorf("exceeds maximum staking duration of %s", ux.FormatDuration(genesis.MainnetParams.MaxStakeDuration))
+	}
+	if d < minL1StakingDuration {
+		return fmt.Errorf("below the minimum staking duration of %s", ux.FormatDuration(genesis.MainnetParams.MinStakeDuration))
```
felipemadero, line 71: also fix duration here

pkg/prompts/validations.go
```diff
@@ -58,6 +58,21 @@ func validateMainnetStakingDuration(input string) error {
 	return nil
 }
 
+func validateMainnetL1StakingDuration(input string) error {
+	const minL1StakingDuration = 24 * time.Hour
```
felipemadero, line 62: ok lets use 2 minutes if possible

pkg/prompts/prompts.go
```diff
@@ -263,6 +264,20 @@ func (*realPrompter) CaptureMainnetDuration(promptStr string) (time.Duration, er
 	return time.ParseDuration(durationStr)
 }
 
+func (*realPrompter) CaptureMainnetL1Duration(promptStr string) (time.Duration, error) {
```
felipemadero, line 267: CaptureMainnetL1StakingDuration?

cmd/blockchaincmd/add_validator.go
```diff
@@ -376,7 +376,7 @@ func CallAddValidator(
 			}
 		}
 		if duration == 0 {
-			duration, err = PromptDuration(time.Now(), network)
+			duration, err = PromptDuration(time.Now(), network, pos)
```
felipemadero, line 379: we know it is pos, maybe clear is using true? as you wish

felipemadero (approved):
preapproving with comments

pkg/prompts/prompts.go
arturrez: agree

pkg/prompts/validations.go
arturrez: good catch. thank you. fixed.
my bad

PR #2456, opened 2024-12-16
Deprecate Etna Devnet

sukantoraymond:
Disable etna devnet functionalities


cmd/blockchaincmd/add_validator.go
```diff
@@ -663,8 +662,6 @@ func PromptDuration(start time.Time, network models.Network, isPos bool) (time.D
 			d, err = app.Prompt.CaptureMainnetL1StakingDuration(txt)
 		case network.Kind == models.Mainnet && !isPos:
 			d, err = app.Prompt.CaptureMainnetDuration(txt)
-		case network.Kind == models.EtnaDevnet:
```
arturrez, line 666: you also need to remove
```
func (_m *Prompter) CaptureEtnaDuration(promptStr string) (time.Duration, error) {
```

cmd/nodecmd/create.go
```diff
@@ -220,8 +220,8 @@ func preCreateChecks(clusterName string) error {
 	}
 	if clusterConfig.Local {
 		return notImplementedForLocal("create")
-	} // bootsrap checks
-	if globalNetworkFlags.UseEtnaDevnet && (len(bootstrapIDs) != 0 || len(bootstrapIPs) != 0 || genesisPath != "" || upgradePath != "") {
```
arturrez, line 224: it was &&, not ||
so entire thing needs to be removed 
```
	if len(bootstrapIDs) != 0 || len(bootstrapIPs) != 0 || genesisPath != "" || upgradePath != "" {
		return fmt.Errorf("etna devnet uses predefined bootsrap configuration")
	}
	```

File: pkg/constants/etna.go [hunk 3]
```diff
@@ -10,39 +10,8 @@ import (
 	avagoconstants "github.com/ava-labs/avalanchego/utils/constants"
 )
```
arturrez, line 12: pls remove `	_ "embed"` also. it's not used

arturrez (commented):
commented

arturrez (approved):

Issue #2465, opened 2024-12-18
[Bug]: text file busy error during local cluster track subnet


felipemadero:
**Describe the bug**
On a docker environment inside AWS node, local cluster track can fail due to busy text error

**To Reproduce**
`blockchain deploy`

**Expected behavior**
Depend on environment, non deterministic

**Screenshots**
![CleanShot 2024-12-17 at 14 29 02](https://github.com/user-attachments/assets/7dc1efbf-f239-473c-bcff-6029f0fe1423)

**Logs**
If applicable, please include the relevant logs that indicate a problem.

**Operating System**
Which OS you used to reveal the bug.

**Additional context**

PR #2494, opened 2024-12-20
avoid text file busy error on track

felipemadero:
## Why this should be merged
Closes https://github.com/ava-labs/avalanche-cli/issues/2465


cmd/networkcmd/helpers.go
```diff
@@ -63,11 +63,13 @@ func TrackSubnet(
 	}
 
 	pluginPath := filepath.Join(app.GetPluginsDir(), vmID.String())
-	if err := utils.FileCopy(vmBin, pluginPath); err != nil {
-		return err
-	}
-	if err := os.Chmod(pluginPath, constants.DefaultPerms755); err != nil {
-		return err
+	if !utils.IsExecutable(pluginPath) {
```
michaelkaplan13, line 66: Should we be checking if the file exists before creating a new copy, and then checking if it's not already executable before doing the chmod?

pkg/node/local.go
```diff
@@ -78,12 +78,15 @@ func TrackSubnetWithLocalMachine(
...
 	}
 	rootDir := app.GetLocalDir(clusterName)
+
 	pluginPath := filepath.Join(rootDir, "node1", "plugins", vmID.String())
-	if err := utils.FileCopy(vmBin, pluginPath); err != nil {
-		return err
-	}
-	if err := os.Chmod(pluginPath, constants.DefaultPerms755); err != nil {
-		return err
+	if !utils.IsExecutable(pluginPath) {
```
michaelkaplan13, line 83: Can we make a unit for these code paths?

cmd/networkcmd/helpers.go
felipemadero: If the file exists and is not executable it is sign of a problem in the previous setup. It can be a partial copy for
example. So probably it is best to copy it again. 
Also, if it is not executable, there is not risk of encountering `text file busy` error because that happens
when the binary is being executed.

michaelkaplan13: Gotcha, could you add a comment explaining that context here? 🙏

felipemadero: ok

pkg/node/local.go
felipemadero: added

sukantoraymond (approved):

pkg/utils/file.go
```diff
@@ -38,7 +38,7 @@ func IsExecutable(filename string) bool {
 		return false
 	}
 	info, _ := os.Stat(filename)
-	return info.Mode()&0x0100 != 0
+	return info.Mode()&0o0100 != 0
```
learyce, line 41: What is this &0o0100?

pkg/utils/file.go
```diff
@@ -77,6 +77,24 @@ func FileCopy(src string, dst string) error {
 	return os.WriteFile(dst, data, constants.WriteReadReadPerms)
 }
 
+// SetupExecFile copies a file into destination and set it to have exec perms,
+// if destination either does not exists, or is not executable
+func SetupExecFile(src string, dst string) error {
+	if !IsExecutable(dst) {
```
learyce, line 83: Do we need to error/log if the file isn't executable?

pkg/utils/file_test.go
```diff
@@ -98,3 +98,67 @@ func TestReadGoVersion(t *testing.T) {
 		}
 	})
 }
+
+func TestSetupExecFile(t *testing.T) {
```
learyce, line 102: Love to see this test

pkg/utils/file.go
felipemadero: 0o0100 (better 0o100, will fix), is a bit mask that, when applying bitwise AND (&) results
in a != 0 state whenever the file can be executed by the owner.
A perm can be seen as a 9-bit sequence mapping to rwxrwxrwx read-write-execute
for owner,group,everybody.
octal 0o100 means binary 001000000.
Probably better to set a constant and also add a comment

pkg/utils/file.go
felipemadero: We don't want to error. This also happens on first install.
We are not going to error on first install, and we prefer to try to 
recover if a previous install failed.
Probably we can add log for a failed previous install, while trying to recover. 
But that log of unusual situations usually go unnoticed into the log file. 
We have logs that go into logs file, usually not seen by the user on stdout.
And we have ux output for the user on stdout (that also go to logs).

pkg/utils/file.go
felipemadero: It previously was 0x0100 == binary 100000000 that was a bug, that was all combinations where
the owner has read perms

learyce: Gotcha, would you mind adding a comment that explains this?

felipemadero: added

PR #2495, opened 2024-12-20
add l1 nodes info to local network describe

felipemadero:
## Why this should be merged
current output of local network deploy, describe, start, status is misleaving, showing only
the local primary network nodes, but not the local l1 nodes (local cluster)

### DIFF
--- a/pkg/localnet/localnet.go
+++ b/pkg/localnet/localnet.go
@@ -33,7 +33,13 @@ func GetEndpoint() (string, error) {
 }
 
 func GetClusterInfo() (*rpcpb.ClusterInfo, error) {
-	cli, err := binutils.NewGRPCClient(
+	return GetClusterInfoWithEndpoint(binutils.LocalNetworkGRPCServerEndpoint)
+}
+
+func GetClusterInfoWithEndpoint(grpcServerEndpoint string) (*rpcpb.ClusterInfo, error) {
+	cli, err := binutils.NewGRPCClientWithEndpoint(
+		grpcServerEndpoint,
+		binutils.WithAvoidRPCVersionCheck(true),
 		binutils.WithDialTimeout(constants.FastGRPCDialTimeout),
 	)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/localnet/localnet.go
```diff
@@ -33,7 +33,13 @@ func GetEndpoint() (string, error) {
 }
 
 func GetClusterInfo() (*rpcpb.ClusterInfo, error) {
-	cli, err := binutils.NewGRPCClient(
+	return GetClusterInfoWithEndpoint(binutils.LocalNetworkGRPCServerEndpoint)
+}
+
+func GetClusterInfoWithEndpoint(grpcServerEndpoint string) (*rpcpb.ClusterInfo, error) {
```
arturrez, line 39: ❤️

pkg/node/local.go
```diff
@@ -758,7 +758,7 @@ func listLocalClusters(app *application.Avalanche, clusterNamesToInclude []strin
 	return localClusters, nil
 }
 
-func DestroyCurrentIfLocalNetwork(app *application.Avalanche) error {
+func CurrentIsLocalNetwork(app *application.Avalanche) (bool, string, error) {
```
arturrez, line 761: nit: add comment what `(bool, string, error)`  means in func return as it's not clear from the name

arturrez: maybe func name is not that great, there should be a better name like `GetLocalNetworkState` which returns clusterName and state running or not

pkg/node/local.go
```diff
@@ -776,58 +776,43 @@ func DestroyCurrentIfLocalNetwork(app *application.Avalanche) error {
...
-	if currentlyRunningRootDir == "" {
-		return nil
+	if isLocal {
+		_ = DestroyLocalNode(app, clusterName)
 	}
-	localClusters, err := listLocalClusters(app, nil)
+	return nil
+}
+
+func StopCurrentIfLocalNetwork(app *application.Avalanche) error {
```
arturrez, line 809: nit: better name can be StopLocalNetwork which returns error if it's not running or not local 

pls feel free to ignore this comment

arturrez (approved):
approved with some nits

pkg/node/local.go
felipemadero: renamed to ConnectedToLocalNetwork.

PR #2498, opened 2024-12-20
improve get balance ux

felipemadero:
## Why this should be merged
- add flag for node id (used in combination with l1 name)
- prompt the user to choose among known l1s
- prompt the user to choose among validators

### DIFF
--- a/cmd/validatorcmd/getBalance.go
+++ b/cmd/validatorcmd/getBalance.go
@@ -4,14 +4,25 @@ package validatorcmd
 
 import (
 	"fmt"
+	"sort"
 
 	"github.com/ava-labs/avalanche-cli/pkg/cobrautils"
+	"github.com/ava-labs/avalanche-cli/pkg/contract"
+	"github.com/ava-labs/avalanche-cli/pkg/models"
 	"github.com/ava-labs/avalanche-cli/pkg/networkoptions"
 	"github.com/ava-labs/avalanche-cli/pkg/txutils"
+	"github.com/ava-labs/avalanche-cli/pkg/utils"
 	"github.com/ava-labs/avalanche-cli/pkg/ux"
+	"github.com/ava-labs/avalanche-cli/pkg/validatormanager"
+	validatorManagerSDK "github.com/ava-labs/avalanche-cli/sdk/validatormanager"
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/units"
+	"github.com/ava-labs/avalanchego/vms/platformvm"
+	"github.com/ava-labs/avalanchego/vms/platformvm/api"
+	"github.com/ethereum/go-ethereum/common"
+
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/maps"
 )
 
 var globalNetworkFlags networkoptions.NetworkFlags
@@ -19,6 +30,7 @@ var globalNetworkFlags networkoptions.NetworkFlags
 var (
 	l1              string
 	validationIDStr string
+	nodeIDStr       string
 )
 
 var getBalanceSupportedNetworkOptions = []networkoptions.NetworkOption{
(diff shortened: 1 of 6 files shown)


cmd/validatorcmd/getBalance.go
```diff
@@ -58,68 +71,137 @@ func getBalance(_ *cobra.Command, _ []string) error {
...
-		ux.Logger.PrintToUser("  Validator Balance: %.5f", float64(balance)/float64(units.Avax))
+	validationID, cancel, err := getNodeValidationID(network, l1, nodeIDStr, validationIDStr)
+	if err != nil {
+		return err
+	}
+	if cancel {
 		return nil
 	}
+	if validationID == ids.Empty {
+		return fmt.Errorf("the specified node is not a L1 validator")
```
arturrez, line 82: nit: add `as validationID is empty` to the error msg
pls feel free to ignore this

cmd/validatorcmd/getBalance.go
```diff
@@ -58,68 +71,137 @@ func getBalance(_ *cobra.Command, _ []string) error {
...
+
+	return nil
+}
+
+func getNodeValidationID(
+	network models.Network,
+	l1 string,
+	nodeIDStr string,
+	validationIDStr string,
+) (ids.ID, bool, error) {
```
arturrez, line 99: nit: 
it's not not clear what `bool` in return means.
pls change func name or add comment for this

cmd/validatorcmd/getBalance.go
```diff
@@ -58,68 +71,137 @@ func getBalance(_ *cobra.Command, _ []string) error {
...
+	case l1ListOption:
+		chainSpec := contract.ChainSpec{
+			BlockchainName: l1,
 		}
-		bootstrapValidatorsString := []string{}
-		bootstrapValidatorsToIndexMap := make(map[string]int)
-		for index, validator := range bootstrapValidators {
-			bootstrapValidatorsString = append(bootstrapValidatorsString, validator.NodeID)
-			bootstrapValidatorsToIndexMap[validator.NodeID] = index
+		chainSpec.SetEnabled(true, false, false, false, true)
```
arturrez, line 130: this is very cryptic line. not sure what it means `true, false, false, false, true`

arturrez: looking at 
```
func (cs *ChainSpec) SetEnabled(
	blockchainNameFlagEnabled bool,
	cChainFlagEnabled bool,
	pChainFlagEnabled bool,
	xChainFlagEnabled bool,
	blockchainIDFlagEnabled bool,
) {
	cs.blockchainNameFlagEnabled = blockchainNameFlagEnabled
	cs.cChainFlagEnabled = cChainFlagEnabled
	cs.pChainFlagEnabled = pChainFlagEnabled
	cs.xChainFlagEnabled = xChainFlagEnabled
	cs.blockchainIDFlagEnabled = blockchainIDFlagEnabled
}
```
seems to be not that useful and more confusing when used.

arturrez: pls feel free to ignore this

cmd/validatorcmd/getBalance.go
```diff
@@ -58,68 +71,137 @@ func getBalance(_ *cobra.Command, _ []string) error {
...
+			true,
+			false,
+		)
 		if err != nil {
-			return err
+			return ids.Empty, false, err
 		}
-	} else {
-		validationID, err := app.Prompt.CaptureID("What is the validator's validationID?")
+		fmt.Println(rpcURL)
```
arturrez, line 192: debug?

cmd/validatorcmd/list.go
```diff
@@ -50,6 +53,14 @@ func list(_ *cobra.Command, args []string) error {
 	}
 
 	blockchainName := args[0]
+	sc, err := app.LoadSidecar(blockchainName)
```
arturrez, line 56: we should move this code up to the top of the func. including `blockchainName := args[0]`.
these is no point of doing networkoptions.GetNetworkFromCmdLineFlags if sidecar reports blockchain as not SoV

pkg/application/app.go
```diff
@@ -620,7 +623,8 @@ func (app *Avalanche) GetBlockchainNamesOnNetwork(network models.Network) ([]str
 				}
 			}
 		}
-		if sc.Networks[networkName].BlockchainID != ids.Empty {
+		sovKindCriteria := !onlySOV || onlySOV && sc.Sovereign
```
arturrez, line 626: `!onlySOV || onlySOV` always returns `true`

arturrez: did you mean `!onlySOV || sc.Sovereign` ?

arturrez (approved):

cmd/validatorcmd/getBalance.go
felipemadero: ignoring as in the other PR

cmd/validatorcmd/getBalance.go
felipemadero: those 2 are not public

felipemadero: adding a comment to improve legibility

pkg/application/app.go
felipemadero: && has precedence over ||, this is the same as "!onlySOV || (onlySOV && sc.Sovereign)"

PR #2499, opened 2024-12-20
improve increase balance ux

felipemadero:

### DIFF
--- a/cmd/validatorcmd/increaseBalance.go
+++ b/cmd/validatorcmd/increaseBalance.go
@@ -24,7 +24,7 @@ var (
 	useLedger       bool
 	useEwoq         bool
 	ledgerAddresses []string
-	balanceFlag     float64
+	balanceFlt      float64
 )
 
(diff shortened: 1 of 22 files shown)


cmd/validatorcmd/increaseBalance.go
```diff
@@ -64,54 +65,18 @@ func increaseBalance(_ *cobra.Command, _ []string) error {
...
+
+	validationID, cancel, err := getNodeValidationID(network, l1, nodeIDStr, validationIDStr)
+	if err != nil {
+		return err
+	}
+	if cancel {
+		return nil
 	}
+	if validationID == ids.Empty {
+		return fmt.Errorf("the specified node is not a L1 validator")
```
arturrez, line 77: nit: add explanation to the error msg like `the specified node is not a L1 validator as validationID is empty` 

pls feel free to ignore this

arturrez (approved):

cmd/validatorcmd/increaseBalance.go
felipemadero: ok this will ignore

PR #2506, opened 2024-12-30
add validator: resume from any previous failed step

felipemadero:
## Why this should be merged
It enables to execute add validator again, for the same node, if some intermediate step
failed (eg aggregation).


pkg/validatormanager/registration.go
```diff
@@ -429,7 +429,7 @@ func InitValidatorRegistration(
 			if !errors.Is(err, validatorManagerSDK.ErrNodeAlreadyRegistered) {
 				return nil, ids.Empty, evm.TransactionError(tx, err, "failure initializing validator registration")
 			}
-			ux.Logger.PrintToUser("the validator registration was already initialized. Proceeding to the next step")
+			ux.Logger.PrintToUser(logging.LightBlue.Wrap("The validator registration was already initialized. Proceeding to the next step"))
```
arturrez, line 432: ❤️

pkg/utils/common.go
```diff
@@ -572,3 +573,21 @@ func IsValidSemanticVersion(version string, component string) bool {
 	}
 	return true
 }
+
+func PrintNovelErrors(
```
arturrez, line 577: not clear why novel is used in func name

arturrez (approved):

pkg/utils/common.go
felipemadero: added a comment and make a name change to PrintUnreportedErrors

PR #2507, opened 2024-12-31
Aggregator log file

felipemadero:
## Why this should be merged
Currently aggregator log level by default is Off. And it is only displayed to stdout if set to other log level. This PR
changes the default, by setting log level default to Debug, and saving logs into an
aggregator logs file.

Depends on https://github.com/ava-labs/icm-services/pull/612


cmd/blockchaincmd/add_validator.go
```diff
@@ -111,7 +112,8 @@ Testnet or Mainnet.`,
 	cmd.Flags().BoolVar(&aggregatorAllowPrivatePeers, "aggregator-allow-private-peers", true, "allow the signature aggregator to connect to peers with private IP")
 	privateKeyFlags.AddToCmd(cmd, "to pay fees for completing the validator's registration (blockchain gas token)")
 	cmd.Flags().StringVar(&rpcURL, "rpc", "", "connect to validator manager at the given rpc endpoint")
-	cmd.Flags().StringVar(&aggregatorLogLevel, "aggregator-log-level", "Off", "log level to use with signature aggregator")
+	cmd.Flags().StringVar(&aggregatorLogLevel, "aggregator-log-level", constants.DefaultAggregatorLogLevel, "log level to use with signature aggregator")
+	cmd.Flags().BoolVar(&aggregatorLogToStdout, "aggregator-log-to-stdout", false, "dump signature aggregator logs to stdout")
```
arturrez, line 116: nit: better wording "use stdout for signature aggregator logs"
pls feel free to ignore this

cmd/blockchaincmd/add_validator.go
```diff
@@ -456,6 +458,23 @@ func CallAddValidator(
 	if err != nil {
 		return err
 	}
+	logLvl, err := logging.ToLevel(aggregatorLogLevel)
+	if err != nil {
```
arturrez, line 462: nit: we should display user error message and that logLevel is set to debug 
pls add something like  `ux.Logger.PrintToUser("undefined logLevel. Setting signature aggregator log to DEBUG")` or something similar.
pls feel free to ignore this suggestion

cmd/blockchaincmd/deploy.go
```diff
@@ -195,7 +195,8 @@ so you can take your locally tested Blockchain and deploy it on Fuji or Mainnet.
 	cmd.Flags().BoolVar(&generateNodeID, "generate-node-id", false, "whether to create new node id for bootstrap validators (Node-ID and BLS values in bootstrap JSON file will be overridden if --bootstrap-filepath flag is used)")
 	cmd.Flags().StringSliceVar(&bootstrapEndpoints, "bootstrap-endpoints", nil, "take validator node info from the given endpoints")
 	cmd.Flags().BoolVar(&convertOnly, "convert-only", false, "avoid node track, restart and poa manager setup")
-	cmd.Flags().StringVar(&aggregatorLogLevel, "aggregator-log-level", "Off", "log level to use with signature aggregator")
+	cmd.Flags().StringVar(&aggregatorLogLevel, "aggregator-log-level", constants.DefaultAggregatorLogLevel, "log level to use with signature aggregator")
+	cmd.Flags().BoolVar(&aggregatorLogToStdout, "aggregator-log-to-stdout", false, "dump signature aggregator logs to stdout")
```
arturrez, line 199: same nit

cmd/blockchaincmd/remove_validator.go
```diff
@@ -67,7 +68,8 @@ these prompts by providing the values with flags.`,
 	cmd.Flags().BoolVar(&aggregatorAllowPrivatePeers, "aggregator-allow-private-peers", true, "allow the signature aggregator to connect to peers with private IP")
 	privateKeyFlags.AddToCmd(cmd, "to pay fees for completing the validator's removal (blockchain gas token)")
 	cmd.Flags().StringVar(&rpcURL, "rpc", "", "connect to validator manager at the given rpc endpoint")
-	cmd.Flags().StringVar(&aggregatorLogLevel, "aggregator-log-level", "Off", "log level to use with signature aggregator")
+	cmd.Flags().StringVar(&aggregatorLogLevel, "aggregator-log-level", constants.DefaultAggregatorLogLevel, "log level to use with signature aggregator")
+	cmd.Flags().BoolVar(&aggregatorLogToStdout, "aggregator-log-to-stdout", false, "dump signature aggregator logs to stdout")
```
arturrez, line 72: same nit

pkg/application/app.go
```diff
@@ -98,6 +98,10 @@ func (app *Avalanche) GetLocalDir(clusterName string) string {
 	return filepath.Join(app.baseDir, constants.LocalDir, clusterName)
 }
 
+func (app *Avalanche) GetLogDir() string {
```
arturrez, line 101: nit: we should also make a small change to \`func (app *Avalanche) GetAPMLog()` to use this func

cmd/contractcmd/init_validator_manager.go
```diff
@@ -154,6 +158,23 @@ func initValidatorManager(_ *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	logLvl, err := logging.ToLevel(validatorManagerFlags.aggregatorLogLevel)
+	if err != nil {
+		logLvl = logging.Debug
+	}
+	aggregatorLogDir := app.GetLogDir()
```
arturrez, line 165: nit: we use the same code in multiple places. maybe it makes sense to move it to something like `func (app *Avalanche) GetAggregatorLogger` which accepts clusterName for example and uses `aggregatorLogDir = app.GetLocalDir(clusterName)` for non-empty clusterName and `app.GetLogDir()` elsewhere

pkg/constants/constants.go
```diff
@@ -333,4 +333,7 @@ const (
 	MainnetCChainICMRegistryAddress = "0x7C43605E14F391720e1b37E49C78C4b03A488d98"
 	FujiCChainICMRegistryAddress    = "0xF86Cb19Ad8405AEFa7d09C778215D2Cb6eBfB228"
 	ValidatorUptimeDeductible       = uint64(10) // seconds to make sure all L1 validators would agree on uptime
+
+	// Aggregator
+	DefaultAggregatorLogLevel = "Debug"
```
arturrez, line 338: default value should be `Info` not `Debug`. User should explicitly set it to Debug if needed

arturrez (approved):

pkg/constants/constants.go
felipemadero: Debug is a better default IMO as Info usually does not provide hints into the error cause if any is obtained.
This is anyway going into a log file so not cluttering the user ux. Take into account that we want to 
know the cause for an aggregator failure just after seeing an error, and avoid trying to duplicate the issue
(if that is possible at all).

PR #2522, opened 2025-01-06
Refactor node setup out of node create

sukantoraymond:
Refactor node setup out of `node create`, specifically the setup() function. 

`Node create` is now node provisioning on AWS / GCP, followed by setup().

Adds a new node command, `node setup`, which installs Avalanche-CLI issued Ubuntu AMI on ubuntu based servers. This command requires the server's IP address and ssh private key. It also requires the server to have allowed access to port 22 to machine that is calling `node setup` command. The `node setup` command is essentially calling setup() function.

### DIFF
--- a/cmd/nodecmd/local.go
+++ b/cmd/nodecmd/local.go
@@ -64,6 +64,7 @@ var (
 	minimumStakeDuration         uint64
 	latestAvagoReleaseVersion    bool
 	latestAvagoPreReleaseVersion bool
+	validatorManagerAddress      string
 )
 
 // const snapshotName = "local_snapshot"


meaghanfitzgerald:
Does this conflict with https://github.com/ava-labs/avalanche-cli/pull/2337? Should 2337 be closed?

cmd/nodecmd/create.go
```diff
@@ -658,6 +657,7 @@ func createNodes(cmd *cobra.Command, args []string) error {
 	}
 	startTime := time.Now()
 	if addMonitoring {
+		spinSession := ux.NewUserSpinner()
```
arturrez, line 660: https://github.com/ava-labs/avalanche-cli/pull/2522/files#diff-af9a7cff84db7d7f438c72aad8de4aeb4f53e53bcd605dbb2a0387eac8e57d95R652 is already there. not sure why we need another local one

cmd/nodecmd/setup.go
```diff
@@ -0,0 +1,231 @@
...
+)
+
+func newSetupCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "setup",
+		Short: "Sets up a new Avalanche Node on remote server",
+		Long: `The node setup command installs Avalanche Go on specified remote servers. 
+To run the command, the remote servers' IP addresses and SSH private keys are required. 
+
+Currently, only ubuntu-based operating system is supported.`,
```
arturrez, line 40: nit: only Ubuntu OS is supported

cmd/nodecmd/setup.go
```diff
@@ -0,0 +1,231 @@
...
+	cmd.Flags().BoolVar(&partialSync, "partial-sync", true, "primary network partial sync")
+	return cmd
+}
+
+func setup(hosts []*models.Host, avalancheGoVersion string, network models.Network) error {
+	if globalNetworkFlags.UseDevnet {
+		partialSync = false
+		ux.Logger.PrintToUser("disabling partial sync default for devnet")
+	}
+	ux.Logger.PrintToUser("Starting bootstrap process on the newly created Avalanche node(s)...")
```
arturrez, line 65: nit: change wording to something like Setting up Avalachego nodes

cmd/nodecmd/setup.go
```diff
@@ -0,0 +1,231 @@
...
+		ux.Logger.PrintToUser("%s Public IP: %s | %s ", logging.Green.Wrap(">"), host.IP, logging.Green.Wrap(nodeID.String()))
+		ux.Logger.PrintToUser("staker.crt, staker.key and signer.key are stored at %s. Please keep them safe, as these files can be used to fully recreate your node.", nodePath)
+		ux.Logger.PrintLineSeparator()
+	}
+}
+
+func promptSetupNodes() error {
+	var err error
+	var numNodes int
+	ux.Logger.PrintToUser("Note that currently we only support Ubuntu operating system")
```
arturrez, line 202: nit: better working "Only Ubuntu OS is supported"

pkg/models/result.go
```diff
@@ -68,7 +68,7 @@ func (nr *NodeResults) GetErrorHostMap() map[string]error {
 	return hostErrors
 }
 
-func (nr *NodeResults) HasNodeIDWithError(nodeID string) bool {
+func (nr *NodeResults) HasIDWithError(nodeID string) bool {
```
arturrez, line 71: nit: I can see this func used for NodeIDs and NodeIPs, I feel like it should be `func  HasError(id string)` 
as id would cover NodeID and Node IP

cmd/nodecmd/setup.go
```diff
@@ -0,0 +1,231 @@
...
+
+	if wgResults.HasErrors() {
+		return fmt.Errorf("failed to deploy node(s) %s", wgResults.GetErrorHostMap())
+	} else {
+		ux.Logger.PrintToUser(logging.Green.Wrap("AvalancheGo and Avalanche-CLI installed and node(s) are bootstrapping!"))
+	}
+	return nil
+}
+
+func setupNode(_ *cobra.Command, _ []string) error {
```
arturrez, line 129: we need to verify that provided hosts are Ubuntu OS and we have enough access to do what we need

cmd/nodecmd/setup.go
```diff
@@ -0,0 +1,231 @@
...
+	cmd.Flags().StringArrayVar(&nodeIPs, "node-ips", []string{}, "IP addresses of nodes")
+	cmd.Flags().StringArrayVar(&sshKeyPaths, "ssh-key-paths", []string{}, "ssh key paths")
+	cmd.Flags().BoolVar(&useSSHAgent, "use-ssh-agent", false, "use ssh agent(ex: Yubikey) for ssh auth")
+	cmd.Flags().StringVar(&genesisPath, "genesis", "", "path to genesis file")
+	cmd.Flags().StringVar(&upgradePath, "upgrade", "", "path to upgrade file")
+	cmd.Flags().BoolVar(&partialSync, "partial-sync", true, "primary network partial sync")
+	return cmd
+}
+
+func setup(hosts []*models.Host, avalancheGoVersion string, network models.Network) error {
```
arturrez, line 60: we need to check if we are not overwriting existing staking keys on provided IP as it could be devastating for the user.

cmd/nodecmd/setup.go
arturrez: another suggestion is to verify that host meets minimum requirements( for aws provision we control this, but for user provided IP we don't know what the box is)

cmd/nodecmd/setup.go
```diff
@@ -0,0 +1,233 @@
...
+	cmd.Flags().StringVar(&useCustomAvalanchegoVersion, "custom-avalanchego-version", "", "install given avalanchego version on node/s")
+	cmd.Flags().StringVar(&useAvalanchegoVersionFromSubnet, "avalanchego-version-from-subnet", "", "install latest avalanchego version, that is compatible with the given subnet, on node/s")
+	cmd.Flags().BoolVar(&publicHTTPPortAccess, "public-http-port", false, "allow public access to avalanchego HTTP port")
+	cmd.Flags().StringArrayVar(&nodeIPs, "node-ips", []string{}, "IP addresses of nodes")
+	cmd.Flags().StringArrayVar(&sshKeyPaths, "ssh-key-paths", []string{}, "ssh key paths")
+	cmd.Flags().BoolVar(&useSSHAgent, "use-ssh-agent", false, "use ssh agent(ex: Yubikey) for ssh auth")
+	cmd.Flags().StringVar(&genesisPath, "genesis", "", "path to genesis file")
+	cmd.Flags().StringVar(&upgradePath, "upgrade", "", "path to upgrade file")
+	cmd.Flags().BoolVar(&partialSync, "partial-sync", true, "primary network partial sync")
+	cmd.Flags().BoolVar(&overrideExisting, "override-existing", true, "override existing staking files")
```
arturrez, line 58: should be `false` by default

arturrez (approved):
preapproved with one small comment

felipemadero (approved):

Issue #2428, opened 2024-12-11
[Bug]: Failure for addValidator on Fuji


felipemadero:
**Describe the bug**
A timeout happens during addValidator execution for a L1 on Fuji and the cmd ends.
 
```
Waiting for P-Chain to update validator information ... 100%
✗ failure issuing txs to activate proposer VM fork for client &ethclient.client{c:(*rpc.Client)(0xc0015446c0)}: Post "http://127.0.0.1:9650/ext/bc/wNWCENGg1E8gYNtWQMv2V6ZoLraNamaw6qgEqx1mRmQ8kLi3S/rpc": context deadline exceeded
✗ failure issuing txs to activate proposer VM fork for client &ethclient.client{c:(*rpc.Client)(0xc0015446c0)}: nonce too low: next nonce 3, tx nonce 2
✗ failure issuing txs to activate proposer VM fork for client &ethclient.client{c:(*rpc.Client)(0xc0015446c0)}: Post "http://127.0.0.1:9650/ext/bc/wNWCENGg1E8gYNtWQMv2V6ZoLraNamaw6qgEqx1mRmQ8kLi3S/rpc": context deadline exceeded
Error: failure issuing txs to activate proposer VM fork for client &ethclient.client{c:(*rpc.Client)(0xc0015446c0)}: Post "http://127.0.0.1:9650/ext/bc/wNWCENGg1E8gYNtWQMv2V6ZoLraNamaw6qgEqx1mRmQ8kLi3S/rpc": context deadline exceeded
```

**To Reproduce**
add a validator to an L1 on Fuji

**Expected behavior**
It is a non deterministic failure

**Screenshots**

**Logs**
If applicable, please include the relevant logs that indicate a problem.

**Operating System**
Which OS you used to reveal the bug.

**Additional context**

Mtsurve:
> **Describe the bug** A timeout happens during addValidator execution for a L1 on Fuji and the cmd ends.
> 
> ```
> Waiting for P-Chain to update validator information ... 100%
> ✗ failure issuing txs to activate proposer VM fork for client &ethclient.client{c:(*rpc.Client)(0xc0015446c0)}: Post "http://127.0.0.1:9650/ext/bc/wNWCENGg1E8gYNtWQMv2V6ZoLraNamaw6qgEqx1mRmQ8kLi3S/rpc": context deadline exceeded
[...]

PR #2534, opened 2025-01-08
improve evm proposer vm activation error handling

felipemadero:
## Why this should be merged
Closes https://github.com/ava-labs/avalanche-cli/issues/2428

- sets a bigger timeout
- improves err msgs (avoid repeated errors)
- adds more err info for the case of a future issue
- don't break flow in case of context timeouts errors here


pkg/evm/evm.go
```diff
@@ -660,38 +660,17 @@ func SetupProposerVM(
...
+		err = issueTxsToActivateProposerVMFork(client, chainID, privKey)
 		if err == nil {
 			break
 		}
-		err = fmt.Errorf(
-			"failure issuing txs to activate proposer VM fork for client %#v: %w",
-			client,
-			err,
-		)
+		err = fmt.Errorf("failure issuing tx to activate proposer VM: %w", err)
```
arturrez, line 669: nit: add msg that it will retry.
also you can display to user that it's retry X out of Y

pkg/evm/evm.go
```diff
@@ -760,7 +739,7 @@ func WaitForNewBlock(
 		}
 		time.Sleep(stepDuration)
 	}
-	return fmt.Errorf("new block not produced in %f seconds", totalDuration.Seconds())
+	return fmt.Errorf("not new block produced in %f seconds", totalDuration.Seconds())
```
arturrez, line 742: better wording `no new block produced ...`

pkg/utils/common.go
```diff
@@ -78,17 +78,22 @@ func SplitKeyValueStringToMap(str string, delimiter string) (map[string]string,
...
 }
 
 // Context for API requests with large timeout
 func GetAPILargeContext() (context.Context, context.CancelFunc) {
-	return context.WithTimeout(context.Background(), constants.APIRequestLargeTimeout)
+	return GetTimedContext(constants.APIRequestLargeTimeout)
+}
+
+// Timed Context
+func GetTimedContext(timeout time.Duration) (context.Context, context.CancelFunc) {
```
arturrez, line 95: ❤️

arturrez (approved):
approved with couple small nits

pkg/evm/evm.go
felipemadero: this is not an error that will be displayed rt, but afterwards all repetitions failed (so retry msg not needed).
the idea is to avoid cluttering the user with the same error on the many repetitions. if adding X/Y it will also be considered a different message and will clutter output. keeping as it is for the moment.

Issue #2532, opened 2025-01-08
[Bug]: blockchain changeWeight is doing validators checks for a permissioned subnet


felipemadero:
**Describe the bug**
trying to use change weight on a L1 gets the msg that the node is not a subnet validator. the check
should be based on GetValidatorsAt API call, not on GetCurrentValidators API call.

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**

**Screenshots**

**Logs**
If applicable, please include the relevant logs that indicate a problem.

**Operating System**
Which OS you used to reveal the bug.

**Additional context**

PR #2545, opened 2025-01-09
Improve change weight

felipemadero:
## Why this should be merged
- Closes https://github.com/ava-labs/avalanche-cli/issues/2532
- Implements missing parts of changeWeight command.
- Add checks for weight churn rate to addValidator
- Move some validator cmd group related functions to SDK


sdk/utils/utils.go
```diff
@@ -22,6 +22,15 @@ func Unique[T comparable](arr []T) []T {
 	return unique
 }
 
+func Belongs[T comparable](input []T, elem T) bool {
```
arturrez, line 25: ❤️

cmd/blockchaincmd/change_weight.go
```diff
@@ -141,59 +134,133 @@ func setWeight(_ *cobra.Command, args []string) error {
...
+	validationID, err := validatorManagerSDK.GetValidationID(rpcURL, nodeID)
 	if err != nil {
 		return err
 	}
 
-	// TODO: we need to wait for the balance from the removed validator to arrive in changeAddr
-	// set arbitrary time.sleep here?
-
-	weight, err = app.Prompt.CaptureWeight("What weight would you like to assign to the validator?")
+	vdrInfo, err := validatorManagerSDK.GetValidatorInfo(network.SDKNetwork(), validationID)
```
arturrez, line 159: sorry not happy with variable name `validatorInfo` in much better name imho
pls feel free ignore this comment :)

cmd/blockchaincmd/change_weight.go
```diff
@@ -141,59 +134,133 @@ func setWeight(_ *cobra.Command, args []string) error {
...
-}
+	deployer := subnet.NewPublicDeployer(app, kc, network)
 
-// getValidatorBalanceFromPChain gets remaining balance of validator from p chain
-func getValidatorBalanceFromPChain() (uint64, error) {
-	return 0, nil
-}
+	var remainingBalanceOwnerAddr, disableOwnerAddr string
+	hrp := key.GetHRP(network.ID)
+	if len(vdrInfo.RemainingBalanceOwner.Addrs) > 0 {
```
arturrez, line 213: should we check if `vdrInfo.RemainingBalanceOwner` is not `nil` to avoid possible crash

cmd/blockchaincmd/change_weight.go
```diff
@@ -141,59 +134,133 @@ func setWeight(_ *cobra.Command, args []string) error {
...
+		remainingBalanceOwnerAddr, err = address.Format("P", hrp, vdrInfo.RemainingBalanceOwner.Addrs[0][:])
+		if err != nil {
+			return err
+		}
+	}
+	if len(vdrInfo.DeactivationOwner.Addrs) > 0 {
```
arturrez, line 219: same: what if `vdrInfo.DeactivationOwner` is nil ?

cmd/blockchaincmd/change_weight.go
```diff
@@ -141,59 +134,133 @@ func setWeight(_ *cobra.Command, args []string) error {
...
+		return err
+	}
 
-// getChangeAddrFromPChain gets validator change addr from info api
-func getChangeAddrFromPChain() (string, error) {
-	return "", nil
+	balance := vdrInfo.Balance
+	if len(vdrInfo.RemainingBalanceOwner.Addrs) > 0 {
+		availableBalance, err := utils.GetNetworkBalance([]ids.ShortID{vdrInfo.RemainingBalanceOwner.Addrs[0]}, network.Endpoint)
+		if err != nil {
```
arturrez, line 243: I am concerned that if  we change weight by removing and adding validator back we should not stop and return err in the middle, we should try to recover and add validator back at least with the same weight.

maybe it makes sense to have error passed to `defer func` that can try adding a validator back if `err` is not nil

pkg/models/network.go
```diff
@@ -226,6 +226,20 @@ func (n *Network) BootstrappingContext() (context.Context, context.CancelFunc) {
 	return context.WithTimeout(context.Background(), timeout)
 }
 
+func (n Network) SDKNetwork() sdkNetwork.Network {
```
arturrez, line 229: i feel like we should deprecate CLI network in favour of SDK network so this bridge is not needed.
should be a separate PR for this

sdk/validatormanager/helpers.go
```diff
@@ -0,0 +1,88 @@
...
+	out, err := contract.CallToMethod(
+		rpcURL,
+		managerAddress,
+		"registeredValidators(bytes)->(bytes32)",
+		nodeID[:],
+	)
+	if err != nil {
+		return ids.Empty, err
+	}
+	validatorID, b := out[0].([32]byte)
```
arturrez, line 83: should we check if len(out)>0 before accessing out[0]?

arturrez, line 83: should `b` be replaced with `err` ?

pkg/models/network.go
felipemadero: yes, a separate PR is needed

sdk/validatormanager/helpers.go
felipemadero: CallToMethod already checks that. But adding anyway because it is a reasonable safeguard.

felipemadero: rename to typeIsOk

arturrez (approved):

arturrez:
ref: https://github.com/ava-labs/avalanche-cli/issues/2556 - create separate issue for Network model deprecation

PR #2559, opened 2025-01-16
build and unittests on ubuntu 20.04 to avoid glibc issues on 24.04

arturrez:
## Why this should be merged
fixes glibc issues for CLI on Ubuntu 24.04
## How this works
changes CI to use ubuntu 20.04 and binary is linked there 
## How this was tested
CI


michaelkaplan13:
JFYI, Ubuntu 20.04 is reaching end of support at the start of April, so we should have a plan for moving everything away from it by then.

https://ubuntu.com/about/release-cycle

PR #2582, opened 2025-01-30
Rename network type from kind to type

sukantoraymond:
This PR: 
- Renames NetworkKind to NetworkType to better reflect what it does
- Assigns globalNetworkFlag value based on network value in node create command

### DIFF
--- a/tests/e2e/utils/versions.go
+++ b/tests/e2e/utils/versions.go
@@ -227,6 +227,7 @@ func GetVersionMapping(mapper VersionMapper) (map[string]string, error) {
 			break
 		}
 	}
+
 	// when running Avago only, always use latest
 	binaryToVersion[OnlyAvagoKey] = OnlyAvagoValue
 


tests/e2e/testcases/subnet/non-sov/local/suite.go
```diff
@@ -277,30 +277,6 @@ var _ = ginkgo.Describe("[Local Subnet non SOV]", ginkgo.Ordered, func() {
...
-
-		commands.CreateSubnetEvmConfigNonSOV(subnetName, utils.SubnetEvmGenesisPath)
-		deployOutput := commands.DeploySubnetLocallyNonSOV(subnetName)
-		_, err := utils.ParseRPCsFromOutput(deployOutput)
-		if err != nil {
-			fmt.Println(deployOutput)
-		}
-		gomega.Expect(err).Should(gomega.BeNil())
-
-		output, err := commands.ListValidators(subnetName, "local")
```
felipemadero, line 295: I believe this test does not need to be removed, but ListValidators implementation fixed (really `blockchain validators` fixed)

pkg/vm/evm_prompts.go
```diff
@@ -200,7 +200,7 @@ func PromptSubnetEVMGenesisParams(
 		}
 	}
 
-	// Gas Kind
+	// Gas Type
```
felipemadero, line 203: you probably want to rename other Gas Kind relayer stuff, eg promptGasTokenKind, defaultsKind, etc. I mean, this comment relates to Gas Kind so all other stuff also applies (or keep comment unmodified for the moment on)

cmd/nodecmd/create.go
```diff
@@ -291,11 +304,12 @@ func createNodes(cmd *cobra.Command, args []string) error {
 		networkoptions.NonLocalSupportedNetworkOptions,
 		"",
 	)
+	setGlobalNetworkFlags(network)
```
felipemadero, line 307: why do we continue using globalNetworkFlags after the network is decided? most probably
we need to avoid using flags from here onwards

felipemadero: I don't think we need this functon, the idea is to generate network from flags, not the other way

cmd/keycmd/list.go
```diff
@@ -612,7 +612,7 @@ func getEvmBasedChainAddrInfo(
 }
 
 func printAddrInfos(addrInfos []addressInfo) {
-	header := []string{"Kind", "Name", "Subnet", "Address", "Token", "Balance", "Network"}
+	header := []string{"Type", "Name", "Subnet", "Address", "Token", "Balance", "Network"}
```
felipemadero, line 615: again, if we replace here, probably we want to replace all addIngo.kind stuff, or avoid replacing here

felipemadero (commented):
you also continue having e2e issues. please use latest ANR to fix those.

PR #2589, opened 2025-01-31
Convert Subnet To L1

sukantoraymond:
Convert cmd `avalanche blockchain convert` converts a subnet into a sovereign L1

Before proceeding: 

- Ensure that the nodes intended to be bootstrap validators of sovereign L1 are tracking the blockchain. 
   
To check, ssh into each of the nodes and go to `./avalanchego/configs/node.json`. Ensure that the field `track-subnets` has value of the Blockchain ID of your blockchain. 

- That there is a publicly accessible RPC URL

This can be done through: 

- Setting up an API node that is tracking the blockchain. The RPC URL will look like `http://API_NODE_IP_ADDRESS:9650/ext/bc/BLOCKCHAIN_ID/rpc`

- Or Make one of the bootstrap validators publicly accessible (http-host: 0.0.0.0), but this is not recommended for security reasons. The RPC URL will look like `http://NODE_IP_ADDRESS:9650/ext/bc/BLOCKCHAIN_ID/rpc`

Usage (for mult-sig): 

`./bin/avalanche blockchain convert <subnetName>`

This command will create a file at specified output destination that contains the partially signed tx. Send this file to be signed by the remaining signers. 

The remaining signers can sign this file by calling `avalanche transaction sign --input-tx-filepath <filePath>`. 

Once everyone has signed, commit the tx by calling  `avalanche transaction commit --input-tx-filepath <filePath>`. The command will return the ConvertToL1 Tx ID.

Once this part is complete, call `./bin/avalanche contract initValidatorManager <blockchainName>` to initialize validator manager (this part is pending)

### DIFF
--- a/tests/e2e/testcases/network/suite.go
+++ b/tests/e2e/testcases/network/suite.go
@@ -83,7 +83,7 @@ var _ = ginkgo.Describe("[Network]", ginkgo.Ordered, func() {
 	ginkgo.It("can stop and restart a deployed subnet SOV", func() {
 		commands.CreateSubnetEvmConfigSOV(subnetName, utils.SubnetEvmGenesisPoaPath)
 		// TODO: use commands.DeploySubnetLocallySOV once having etna release
-		deployOutput := commands.DeploySubnetLocallyWithVersionSOV(subnetName, utils.EtnaAvalancheGoVersion)
+		deployOutput := commands.DeploySubnetLocallyWithVersionSOV(subnetName, "")
 		rpcs, err := utils.ParseRPCsFromOutput(deployOutput)
 		if err != nil {
 			fmt.Println(deployOutput)


cmd/blockchaincmd/convert.go
```diff
@@ -0,0 +1,677 @@
...
+	validatorManagerSDK "github.com/ava-labs/avalanche-cli/sdk/validatormanager"
+	"github.com/ava-labs/avalanchego/api/info"
+	"github.com/ava-labs/avalanchego/config"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/ava-labs/avalanchego/utils/units"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/spf13/cobra"
+)
+
+// avalanche blockchain deploy
```
meaghanfitzgerald, line 41:
```suggestion
// avalanche blockchain convertToL1
```

pkg/ssh/ssh.go
```diff
@@ -589,10 +590,14 @@ func RunSSHRenderAvalancheNodeConfig(
 		avagoConf.BootstrapIDs = bootstrapIDs
 		avagoConf.BootstrapIPs = bootstrapIPs
 		partialSyncI, ok := remoteAvagoConf[config.PartialSyncPrimaryNetworkKey]
-		if ok {
-			partialSync, ok := partialSyncI.(bool)
-			if ok {
-				avagoConf.PartialSync = partialSync
+		if !ok {
```
felipemadero, line 593: why is this value not bool? probably we want to fix it on the input json

cmd/nodecmd/create.go
```diff
@@ -291,6 +304,7 @@ func createNodes(cmd *cobra.Command, args []string) error {
 		networkoptions.NonLocalSupportedNetworkOptions,
 		"",
 	)
+	setGlobalNetworkFlags(network)
```
felipemadero, line 307: I don't believe this should be needed. Remaining code should use the network object, not the flags.
Can you fix that

felipemadero (commented):
probably move the new functions, used both by deploy and convert, to helpers

felipemadero (approved):

PR #2612, opened 2025-02-18
deployed bytecode diff

sukantoraymond:


felipemadero:
provided bytecodes have issues on size and link information

PR #2618, opened 2025-02-20
Add signature aggregator wait for healthy

felipemadero:
## Why this should be merged
Issues are being observed regarding collecting signatures that are just related to network timeouts
and not to misconfigurations or invalid warp message esp/state
This PR is first step to try to aleviate issues, by waiting for full connection to tracked subnet and
primary network before attempting to recollect signatures


sukantoraymond (approved):
Nice One!

PR #2622, opened 2025-02-21
to remove validators decide sov/nonsov based on nodeID

felipemadero:
## Why this should be merged
Users that make conversions from previous non SOV need to be able to remove non SOV validators
after convert.

### DIFF
--- a/sdk/validator/validator.go
+++ b/sdk/validator/validator.go
@@ -15,6 +15,15 @@ import (
 	"golang.org/x/exp/maps"
 )
 
+type ValidatorKind int64
+
+const (
+	UndefinedValidatorKind ValidatorKind = iota
+	NonValidator
+	SovereignValidator
+	NonSovereignValidator
+)
+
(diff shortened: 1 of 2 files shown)


sdk/validator/validator.go
```diff
@@ -89,3 +98,26 @@ func GetRegisteredValidator(
...
+	for _, v := range vs {
+		if v.NodeID == nodeID {
+			if v.TxID == ids.Empty {
+				return SovereignValidator, nil
+			}
+			return NonSovereignValidator, nil
+		}
+	}
+	return NonValidator, nil
+}
```
sukantoraymond, line 123: we can actually get if sov or not from getSubnet API. If it is sov: permissioned = false

example: https://github.com/ava-labs/avalanche-cli/blob/18af22b0dceefc823bd05432e5e4d05d9b3d7839/cmd/blockchaincmd/add_validator.go#L150

sukantoraymond (approved):

PR #2633, opened 2025-02-25
add function to get blockchain id from warp precompile

felipemadero:
## Why this should be merged
To avoid the user to input information we cat retrieve from the validators

### DIFF
--- a/pkg/precompiles/warp.go
+++ b/pkg/precompiles/warp.go
@@ -21,8 +21,11 @@ func WarpPrecompileGetBlockchainID(
 	if err != nil {
 		return ids.Empty, err
 	}
-	received, b := out[0].([32]byte)
-	if !b {
+	if len(out) == 0 {
+		return ids.Empty, fmt.Errorf("error at getBlockchainID call: no return value")
+	}
+	received, ok := out[0].([32]byte)
+	if !ok {
 		return ids.Empty, fmt.Errorf("error at getBlockchainID call, expected ids.ID, got %T", out[0])
 	}
 	return received, nil
--- a/pkg/precompiles/warp.go
+++ b/pkg/precompiles/warp.go
@@ -1,0 +1,29 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package precompiles
+
+import (
+	_ "embed"
+	"fmt"
+
+	"github.com/ava-labs/avalanche-cli/pkg/contract"
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+func WarpPrecompileGetBlockchainID(
+	rpcURL string,
+) (ids.ID, error) {
+	out, err := contract.CallToMethod(
+		rpcURL,
+		WarpPrecompile,
+		"getBlockchainID()->(bytes32)",
+	)
+	if err != nil {
+		return ids.Empty, err
+	}
+	received, b := out[0].([32]byte)
+	if !b {
+		return ids.Empty, fmt.Errorf("error at getBlockchainID call, expected ids.ID, got %T", out[0])
+	}
+	return received, nil
+}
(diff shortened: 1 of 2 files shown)


pkg/precompiles/warp.go
```diff
@@ -0,0 +1,29 @@
...
+) (ids.ID, error) {
+	out, err := contract.CallToMethod(
+		rpcURL,
+		WarpPrecompile,
+		"getBlockchainID()->(bytes32)",
+	)
+	if err != nil {
+		return ids.Empty, err
+	}
+	received, b := out[0].([32]byte)
```
learyce, line 24: Do we need to validate the out array here before we call this?

pkg/precompiles/warp.go
```diff
@@ -0,0 +1,29 @@
...
+	if !b {
```
learyce, line 25: What is the variable b?

pkg/precompiles/warp.go
learyce: For the second question, I could be rusty on Go. More so, I am looking to confirm that this conversion fails, if it'll throw an uncaught error

pkg/precompiles/warp.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	_ "embed"
+	"fmt"
+
+	"github.com/ava-labs/avalanche-cli/pkg/contract"
+	"github.com/ava-labs/avalanchego/ids"
+)
+
+func WarpPrecompileGetBlockchainID(
```
learyce, line 13: Do we have any sort of framework/etc in place for testing functions like these (ie one's that would require a lot of mocking)?

pkg/precompiles/warp.go
sukantoraymond: if the value exists, should prob rename it to ok @felipemadero

pkg/precompiles/warp.go
sukantoraymond: we can test this on local network. Do u want to create separate PR for this? @felipemadero

pkg/precompiles/warp.go
felipemadero: ok

pkg/precompiles/warp.go
felipemadero: we don't have framework in place for testing this as unit test. we usually do E2E testing with a local network
deployment, not with a mock. for sure we can consider mocking if we want to really add testing for 
all code paths.

felipemadero: really, how to unit test this stuff, and how to properly tests the commands ux, for example
(prompts, flags paths), needs good design

pkg/precompiles/warp.go
felipemadero: First question. We are counting that the underlying implementation, if does not error, returns out
with  the appropriate size. This is the current semantic.
That say, better to check for len!

felipemadero: Second question. The type conversion is safe if you also get the b boolean. I mean, there is no uncaught error.

If you do
```
received := out[0].([32]byte)
```
you can have uncaught error.

If you do
```
received, b := out[0].([32]byte)
```
you will have b==false in conversion error

pkg/precompiles/warp.go
felipemadero: open to freely talk on implementation options

pkg/precompiles/warp.go
felipemadero: b == type conversion ok

pkg/precompiles/warp.go
felipemadero: regarding testing this specific feature on E2E, we don't have a command to return blockchain ID,
but it is going to be part of a larger feature in the short term. 
https://github.com/ava-labs/avalanche-cli/issues/2637

felipemadero: do we want to add annotations to the functions to point to the specific E2E?

sukantoraymond (approved):

PR #2641, opened 2025-02-26
Validate convert txs

felipemadero:
## Why this should be merged
- It adds validation of blockchain ID, subnet ID, manager address before generating
a convert tx.
- It gives details on the tx about to be signed or committed before deciding to do so, on
multisig management commands

### DIFF
--- a/cmd/blockchaincmd/convert.go
+++ b/cmd/blockchaincmd/convert.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/ava-labs/avalanche-cli/pkg/blockchain"
+	"github.com/ava-labs/avalanche-cli/pkg/clierrors"
 	"github.com/ava-labs/avalanche-cli/pkg/cobrautils"
 	"github.com/ava-labs/avalanche-cli/pkg/constants"
 	"github.com/ava-labs/avalanche-cli/pkg/contract"
@@ -419,6 +420,15 @@ func convertSubnetToL1(
 	subnetAuthKeysList []string,
 	validatorManagerAddressStr string,
 ) ([]*txs.ConvertSubnetToL1Validator, bool, error) {
+	if subnetID == ids.Empty {
+		return nil, false, clierrors.ErrNoSubnetID
+	}
+	if blockchainID == ids.Empty {
+		return nil, false, clierrors.ErrNoBlockchainID
+	}
+	if !common.IsHexAddress(validatorManagerAddressStr) {
+		return nil, false, clierrors.ErrInvalidValidatorManagerAddress
+	}
 	avaGoBootstrapValidators, err := ConvertToAvalancheGoSubnetValidator(bootstrapValidators)
 	if err != nil {
 		return avaGoBootstrapValidators, false, err
@@ -525,6 +535,13 @@ func convertBlockchain(_ *cobra.Command, args []string) error {
 
 	subnetID := sidecar.Networks[network.Name()].SubnetID
 	blockchainID := sidecar.Networks[network.Name()].BlockchainID
+
+	if subnetID == ids.Empty {
+		return clierrors.ErrNoSubnetID
+	}
+	if blockchainID == ids.Empty {
+		return clierrors.ErrNoBlockchainID
+	}
 
 	if !convertOnly {
 		if validatorManagerAddress == "" {
--- a/cmd/transactioncmd/helpers.go
+++ b/cmd/transactioncmd/helpers.go
@@ -1,0 +1,36 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package transactioncmd
+
+import (
+	"fmt"
+
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+func validateConvertOperation(tx *txs.Tx, action string) (bool, error) {
+	convertToL1Tx, ok := tx.Unsigned.(*txs.ConvertSubnetToL1Tx)
+	if !ok {
+		return false, fmt.Errorf("expected tx to be of type txs.ConvertSubnetToL1Tx, found %T", tx.Unsigned)
+	}
+	ux.Logger.PrintToUser("You are about to %s a txs.ConvertSubnetToL1Tx with the following content:", action)
+	ux.Logger.PrintToUser("  Subnet ID: %s", convertToL1Tx.Subnet)
+	ux.Logger.PrintToUser("  Blockchain ID: %s", convertToL1Tx.BlockchainID)
+	ux.Logger.PrintToUser("  Manager Address: %s", common.BytesToAddress(convertToL1Tx.Address).Hex())
+	ux.Logger.PrintToUser("  Validators:")
+	for _, val := range convertToL1Tx.Validators {
+		nodeID, err := ids.ToNodeID(val.NodeID)
+		if err != nil {
+			return false, fmt.Errorf("unexpected node ID on tx")
+		}
+		ux.Logger.PrintToUser("    %s", nodeID)
+	}
+	ux.Logger.PrintToUser("")
+	ux.Logger.PrintToUser("Please review details and decide if it is safe to continue")
+	ux.Logger.PrintToUser("")
+	return app.Prompt.CaptureYesNo(fmt.Sprintf("Do you want to %s the transaction?", action))
+}
--- a/cmd/transactioncmd/transaction_sign.go
+++ b/cmd/transactioncmd/transaction_sign.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	"github.com/ava-labs/avalanche-cli/cmd/blockchaincmd"
+	"github.com/ava-labs/avalanche-cli/pkg/clierrors"
 	"github.com/ava-labs/avalanche-cli/pkg/cobrautils"
 	"github.com/ava-labs/avalanche-cli/pkg/keychain"
 	"github.com/ava-labs/avalanche-cli/pkg/models"
@@ -15,6 +16,7 @@ import (
 	"github.com/ava-labs/avalanche-cli/pkg/txutils"
 	"github.com/ava-labs/avalanche-cli/pkg/ux"
 	"github.com/ava-labs/avalanchego/ids"
+
 	"github.com/spf13/cobra"
 )
 
@@ -25,8 +27,6 @@ var (
 	keyName         string
 	useLedger       bool
 	ledgerAddresses []string
-
-	errNoSubnetID = errors.New("failed to find the subnet ID for this subnet, has it been deployed/created on this network?")
 )
 
 // avalanche transaction sign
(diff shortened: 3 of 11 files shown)


cmd/transactioncmd/transaction_sign.go
```diff
@@ -112,7 +123,7 @@ func signTx(_ *cobra.Command, args []string) error {
 			}
 			subnetID = sc.Networks[network.Name()].SubnetID
 			if subnetID == ids.Empty {
```
sukantoraymond, line 125: are we currently displaying what tx id is going to be signed before user is signing it?

sukantoraymond: something like u will be signing tx.... and then prompt if user wants to continue

cmd/transactioncmd/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+	ux.Logger.PrintToUser("  Validators:")
+	for _, val := range convertToL1Tx.Validators {
+		nodeID, err := ids.ToNodeID(val.NodeID)
+		if err != nil {
+			return false, fmt.Errorf("unexpected node ID on tx")
+		}
+		ux.Logger.PrintToUser("    %s", nodeID)
+	}
+	ux.Logger.PrintToUser("")
+	ux.Logger.PrintToUser("Please review details and decide if it is safe to continue")
```
sukantoraymond, line 33: Please review the details of the ConvertSubnetToL1 Transaction

cmd/transactioncmd/transaction_sign.go
sukantoraymond: seems like ur already doing it for convert, i guess wondering for createchain / createsubnet but seems that this will be less common now, feel free to ignore

sdk/validatormanager/root.go
```diff
@@ -43,10 +43,10 @@ type NativeTokenValidatorManagerSettings struct {
 
 const (
 	ValidatorMessagesContractAddress = "0x9C00629cE712B0255b17A4a657171Acd15720B8C"
-	ValidatorContractAddress         = "0xC0DEBA5E0000000000000000000000000000000"
-	ProxyContractAddress             = "0xFEEDC0DE0000000000000000000000000000000"
+	ValidatorContractAddress         = "0x0C0DEBA5E0000000000000000000000000000000"
+	ProxyContractAddress             = "0x0FEEDC0DE0000000000000000000000000000000"
```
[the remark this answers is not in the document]
felipemadero: it is the same value but passes more validation checks

cmd/transactioncmd/transaction_sign.go
felipemadero: tx id depends on signature

felipemadero: we can add for other txs types later on, this is for convert only

sukantoraymond (approved):

cmd/blockchaincmd/convert.go
```diff
@@ -526,6 +561,36 @@ func convertBlockchain(_ *cobra.Command, args []string) error {
...
 	blockchainID := sidecar.Networks[network.Name()].BlockchainID
 
+	if doStrongInputChecks && subnetID != ids.Empty {
+		ux.Logger.PrintToUser("Subnet ID to be used is %s", subnetID)
+		if acceptValue, err := app.Prompt.CaptureYesNo("Is this value correct?"); err != nil {
+			return err
+		} else if !acceptValue {
+			subnetID = ids.Empty
+		}
+	}
```
sukantoraymond, line 571: i dont think we need to do this check because users will already check details of subnet id and blockchain id in line 444. A user not using skip check flag will have to verify three times

cmd/blockchaincmd/convert.go
```diff
@@ -526,6 +561,36 @@ func convertBlockchain(_ *cobra.Command, args []string) error {
...
+
+	if doStrongInputChecks && blockchainID != ids.Empty {
+		ux.Logger.PrintToUser("Blockchain ID to be used is %s", blockchainID)
+		if acceptValue, err := app.Prompt.CaptureYesNo("Is this value correct?"); err != nil {
+			return err
+		} else if !acceptValue {
+			blockchainID = ids.Empty
+		}
+	}
```
sukantoraymond, line 586: same comment as above

cmd/blockchaincmd/convert.go
```diff
@@ -526,6 +561,36 @@ func convertBlockchain(_ *cobra.Command, args []string) error {
...
+	if blockchainID == ids.Empty {
+		blockchainID, err = app.Prompt.CaptureID("Which is the blockchain ID?")
```
sukantoraymond, line 588: What is the blockchain ID?

cmd/blockchaincmd/convert.go
```diff
@@ -418,13 +422,44 @@ func convertSubnetToL1(
...
 	managerAddress := common.HexToAddress(validatorManagerAddressStr)
+
+	if doStrongInputsCheck {
+		ux.Logger.PrintToUser("You are about to create a ConvertSubnetToL1Tx on %s with the following content:", network.Name())
+		ux.Logger.PrintToUser("  Subnet ID: %s", subnetID)
+		ux.Logger.PrintToUser("  Blockchain ID: %s", blockchainID)
+		ux.Logger.PrintToUser("  Manager Address: %s", managerAddress.Hex())
+		ux.Logger.PrintToUser("  Validators:")
+		for _, val := range bootstrapValidators {
+			ux.Logger.PrintToUser("    %s", val.NodeID)
```
sukantoraymond, line 450: can we print Node ID: val.NodeID
Balance: float64(val.Balance)/float64(units.Avax)

cmd/blockchaincmd/convert.go
```diff
@@ -418,13 +422,44 @@ func convertSubnetToL1(
...
+			ux.Logger.PrintToUser("    %.5f", float64(val.Balance)/float64(units.Avax))
+		}
+		ux.Logger.PrintToUser("")
+		ux.Logger.PrintToUser("Please review the details of the ConvertSubnetToL1 Transaction")
+		ux.Logger.PrintToUser("")
+		if doContinue, err := app.Prompt.CaptureYesNo("Do you want to create the transaction?"); err != nil {
+			return avaGoBootstrapValidators, false, false, err
+		} else if !doContinue {
+			return avaGoBootstrapValidators, true, false, err
+		}
```
sukantoraymond, line 460: doContinue, err := app.Prompt.CaptureYesNo("Do you want to create the transaction?")
return avaGoBootstrapValidators, doContinue, false, err

pkg/clierrors/clierrors.go
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package clierrors
+
+import "errors"
+
+var (
+	ErrNoBlockchainID                 = errors.New("failed to find the blockchain ID for this subnet, has it been deployed/created on this network?\nyou can use 'avalanche blockchain import' if having partial information on a deployed subnet")
```
sukantoraymond, line 8: no blockchain ID found. Use 'avalanche blockchain import' to provide the required Subnet details

pkg/clierrors/clierrors.go
```diff
@@ -0,0 +1,11 @@
...
+	ErrNoSubnetID                     = errors.New("failed to find the subnet ID for this subnet, has it been deployed/created on this network?\nyou can use 'avalanche blockchain import' if having partial information on a deployed subnet")
```
sukantoraymond, line 9: no Subnet ID found. Use 'avalanche blockchain import' to provide the required Subnet details

pkg/clierrors/clierrors.go
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package clierrors
```
sukantoraymond, line 3: is it better to place this under pkg/constants/error.go?

sukantoraymond: im fine either way

cmd/blockchaincmd/convert.go
felipemadero: A user will have to verify two times, not three times. And the second verification includes
much more information. 
This specific check and input for blockchain id was requested and acorded (i added the subnet id
which is in similar state), so probably the 444 check can be 
removed. I will let devrel to vote on this cc @martineckardt  after trying out the command.
cc @learyce

cmd/blockchaincmd/convert.go
felipemadero: I believe it is cleaner this way

pkg/clierrors/clierrors.go
felipemadero: blckchain ID can also be absent for the blockchain not being deployed. np changing if you also include reference to this option

pkg/clierrors/clierrors.go
felipemadero: lets keep it here but happy to hear @learyce opinion

cmd/blockchaincmd/convert.go
```diff
@@ -418,13 +422,44 @@ func convertSubnetToL1(
...
+
+	if doStrongInputsCheck {
+		ux.Logger.PrintToUser("You are about to create a ConvertSubnetToL1Tx on %s with the following content:", network.Name())
+		ux.Logger.PrintToUser("  Subnet ID: %s", subnetID)
+		ux.Logger.PrintToUser("  Blockchain ID: %s", blockchainID)
+		ux.Logger.PrintToUser("  Manager Address: %s", managerAddress.Hex())
+		ux.Logger.PrintToUser("  Validators:")
+		for _, val := range bootstrapValidators {
+			ux.Logger.PrintToUser("    Node ID: %s", val.NodeID)
+			ux.Logger.PrintToUser("    Balance: %.5f", float64(val.Balance)/float64(units.Avax))
```
martineckardt, line 451: Also show weight here

cmd/blockchaincmd/convert.go
martineckardt: This is a very critical operation. The more confirmation the better

cmd/blockchaincmd/convert.go
martineckardt: Same here. Better to double check. If this value is messed up, the L1 cannot be recovered

cmd/transactioncmd/helpers.go
```diff
@@ -0,0 +1,43 @@
...
+	ux.Logger.PrintToUser("  Blockchain ID: %s", convertToL1Tx.BlockchainID)
+	ux.Logger.PrintToUser("  Manager Address: %s", common.BytesToAddress(convertToL1Tx.Address).Hex())
+	ux.Logger.PrintToUser("  Validators:")
+	for _, val := range convertToL1Tx.Validators {
+		nodeID, err := ids.ToNodeID(val.NodeID)
+		if err != nil {
+			return false, fmt.Errorf("unexpected node ID on tx")
+		}
+		ux.Logger.PrintToUser("    Node ID: %s", nodeID)
+		ux.Logger.PrintToUser("    Balance: %.5f", float64(val.Balance)/float64(units.Avax))
```
martineckardt, line 37: Weight of validator

martineckardt (approved):

pkg/clierrors/clierrors.go
learyce: Let's get Javier's opinion on this text (you'll have to ping him on slack)

pkg/clierrors/clierrors.go
learyce: I'd say constants

felipemadero: moving into constants

pkg/clierrors/clierrors.go
javiertc: No `blockchainID` found. To resolve this:
- Use `avalanche blockchain deploy` to deploy the blockchain and generate a `blockchainID`.
- Or use `avalanche blockchain import` to import an existing configuration.

felipemadero: will use latest one

PR #2644, opened 2025-02-27
enable metrics by default, and set random user id

felipemadero:
## Why this should be merged
Improve CLI metrics collection

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -3,19 +3,16 @@
 package metrics
 
 import (
-	"crypto/sha256"
-	"encoding/base64"
-	"fmt"
 	"os"
-	"os/user"
 	"path/filepath"
 	"runtime"
 	"strings"
 
 	"github.com/ava-labs/avalanche-cli/pkg/application"
 	"github.com/ava-labs/avalanche-cli/pkg/constants"
-
 	"github.com/ava-labs/avalanche-cli/pkg/utils"
+	"github.com/ava-labs/avalanche-cli/pkg/ux"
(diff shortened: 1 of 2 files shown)


pkg/metrics/metrics.go
```diff
@@ -40,14 +37,32 @@ func GetCLIVersion() string {
 	return string(content)
 }
 
-func userIsOptedIn(app *application.Avalanche) bool {
-	if app.Conf.ConfigFileExists() && app.Conf.GetConfigBoolValue(constants.ConfigMetricsEnabledKey) {
-		return true
+func getMetricsUserID(app *application.Avalanche) string {
+	if !app.Conf.ConfigFileExists() || !app.Conf.ConfigValueIsSet(constants.ConfigMetricsUserIDKey) {
+		userID := utils.RandomString(20)
```
sukantoraymond, line 42: why are we replacing user id?

Current implementation
`hash := sha256.Sum256([]byte(codespaceName))
userID = base64.StdEncoding.EncodeToString(hash[:])`
		
SHA 256 is irreversible

sukantoraymond: and it eliminates the need to check for app config files

felipemadero: legal considers this approach, totally independant of user name for id generation, to be the most
beneficial for the purpose cc @javiertc @learyce

felipemadero: checking app config file is not too much of an problem IMO

sukantoraymond (approved):

PR #2650, opened 2025-03-01
Support PoA ACP99 + Externals EVM signatures for validator setup txs

felipemadero:
## Why this should be merged
ACP99 is needed soon. This adds support for ACP99 PoA while keeping support for v1.0.0
There are users which have external signature processes for evm txs. Eg mutlisig.
The PR enable dumping the txs so can be separately signed, for adding , removing, and changing 
weight.


friskyfoxdk:
Excited about this! Going to be testing it today

friskyfoxdk:
Was able to successfully get a changeWeight end to end adjustment completed today using this feature. Much appreciated!

I'll try adding a validator (and removing one too probably) on Monday.

The main tweak that would be helpful for us is if the complete step could be automatically called using a stored-key, rather than outputting the call data and access list. since those methods are not onlyOwner, anyone (including stored-keys) should be able to call them, saving us a manual step :)

cmd/blockchaincmd/convert.go
```diff
@@ -403,6 +410,7 @@ func InitializeValidatorManager(
 			aggregatorAllowPrivatePeers,
 			aggregatorLogger,
 			validatorManagerAddrStr,
+			useACP99,
```
sukantoraymond, line 413: With ACP 99, we no longer have to differentiate initializing validator manager for poa and pos. We can just initialize validator manager and if user wants to do pos, we can do additional steps after.

Example implementation:
https://github.com/ava-labs/avalanche-cli/blob/f0a6d96b7c4a29e1e5a9f81482450f92ba37d9bb/cmd/blockchaincmd/convert.go#L374

pkg/validatormanager/validatormanager.go
```diff
@@ -186,6 +216,7 @@ func SetupPoA(
 	aggregatorAllowPrivatePeers bool,
 	aggregatorLogger logging.Logger,
 	validatorManagerAddressStr string,
+	useACP99 bool,
 ) error {
 	return subnet.InitializeProofOfAuthority(
```
sukantoraymond, line 221: doesn't have to be InitializeProofOfAuthority anymore. Can just be initialize alidator manager

sukantoraymond:
Don't think we need to support both ACP 99 and validator manager 1.0.0 contracts. Supporting both adds additional complexity esp to our sdk. Plus once ACP 99 is released, don't think there's a need for users to use validator manager 1.0.0 contracts. Deferring to @learyce

cmd/blockchaincmd/add_validator.go
```diff
@@ -122,8 +125,9 @@ Testnet or Mainnet.`,
 	cmd.Flags().BoolVar(&waitForTxAcceptance, "wait-for-tx-acceptance", true, "(for Subnets, not L1s) just issue the add validator tx, without waiting for its acceptance")
 	cmd.Flags().Uint16Var(&delegationFee, "delegation-fee", 100, "(PoS only) delegation fee (in bips)")
 	cmd.Flags().StringVar(&subnetIDstr, "subnet-id", "", "subnet ID (only if blockchain name is not provided)")
-	cmd.Flags().StringVar(&validatorManagerOwnerAddress, "validator-manager-owner", "", "validator manager owner address (only if blockchain name is not provided)")
 	cmd.Flags().Uint64Var(&weight, validatorWeightFlag, uint64(constants.DefaultStakeWeight), "set the weight of the validator")
+	cmd.Flags().StringVar(&validatorManagerOwner, "validator-manager-owner", "", "force using this address to issue transactions to the validator manager")
```
sukantoraymond, line 129: we technically dont even need to get validator manager using flag anymore. Currently we only get validator manager from on chain if user doesn't provide argument to add validator. But I think its better if we just do this for all cases -> meaning even if user provides argument to add validator. This means that we don't even need `validator-manager-owner ` flag anymore in add_validator, same goes for remove_validator

cmd/blockchaincmd/remove_validator.go
```diff
@@ -66,6 +66,8 @@ these prompts by providing the values with flags.`,
 	cmd.Flags().BoolVar(&aggregatorLogToStdout, "aggregator-log-to-stdout", false, "use stdout for signature aggregator logs")
 	cmd.Flags().Uint64Var(&uptimeSec, "uptime", 0, "validator's uptime in seconds. If not provided, it will be automatically calculated")
 	cmd.Flags().BoolVar(&force, "force", false, "force validator removal even if it's not getting rewarded")
+	cmd.Flags().BoolVar(&externalValidatorManagerOwner, "external-validator-manager-owner", false, "validator manager owner is external, make hex dump of ech evm transactions, so they can be signed in a separate flow")
+	cmd.Flags().StringVar(&validatorManagerOwner, "validator-manager-owner", "", "force using this address to issue transactions to the validator manager")
```
sukantoraymond, line 70: same comment here, dont think we need validator manager flag anymore, like in add_validator

cmd/blockchaincmd/change_weight.go
```diff
@@ -47,6 +59,14 @@ The L1 has to be a Proof of Authority L1.`,
 	cmd.Flags().StringVar(&nodeEndpoint, "node-endpoint", "", "gather node id/bls from publicly available avalanchego apis on the given endpoint")
 	cmd.Flags().BoolVarP(&useLedger, "ledger", "g", false, "use ledger instead of key (always true on mainnet, defaults to false on fuji/devnet)")
 	cmd.Flags().StringSliceVar(&ledgerAddresses, "ledger-addrs", []string{}, "use the given ledger addresses")
+	cmd.Flags().BoolVar(&externalValidatorManagerOwner, "external-validator-manager-owner", false, "validator manager owner is external, make hex dump of ech evm transactions, so they can be signed in a separate flow")
+	cmd.Flags().StringVar(&validatorManagerOwner, "validator-manager-owner", "", "force using this address to issue transactions to the validator manager")
```
sukantoraymond, line 63: same comment on no need for `validator-manager-owner ` flag like in add_validator

cmd/blockchaincmd/add_validator.go
sukantoraymond: same goes for change_weight

pkg/contract/contract.go
```diff
@@ -319,15 +327,25 @@ func TxToMethod(
...
+		txOpts = &bind.TransactOpts{
+			From:   from,
+			Signer: idempotentSigner,
+			NoSend: true,
+		}
+	} else {
+		txOpts, err = evm.GetTxOptsWithSigner(client, privateKey)
+		if err != nil {
+			return nil, nil, err
+		}
```
sukantoraymond, line 341: nice one!

sdk/validatormanager/validator_manager_poa.go
```diff
@@ -23,25 +23,46 @@ func PoAValidatorManagerInitialize(
 	privateKey string,
 	subnetID ids.ID,
 	ownerAddress common.Address,
+	useACP99 bool,
```
sukantoraymond, line 26: We don't have to differentiate PoAValidatorManagerInitialize and PoSValidatorManagerInitialize anymore. We can just have one function InitializeValidatorManager and have it in root.go / some common file

pkg/vm/create_evm.go
```diff
@@ -127,8 +130,13 @@ func CreateEVMGenesis(
 	}
 	if params.UsePoAValidatorManager {
 		validatormanager.AddTransparentProxyContractToAllocations(params.initialTokenAllocation, proxyOwner)
-		validatormanager.AddValidatorMessagesContractToAllocations(params.initialTokenAllocation)
-		validatormanager.AddPoAValidatorManagerContractToAllocations(params.initialTokenAllocation)
+		if useACP99 {
+			validatormanager.AddValidatorMessagesACP99ContractToAllocations(params.initialTokenAllocation)
```
sukantoraymond, line 134: deployedValidatorMessagesACP99Bytecode has the same value as validator manager 1.0.0 btw

sukantoraymond: so can just reuse AddValidatorMessagesContractToAllocations

pkg/contract/contract.go
```diff
@@ -319,15 +327,25 @@ func TxToMethod(
 	}
 	defer client.Close()
 	contract := bind.NewBoundContract(contractAddress, *abi, client, client, client)
-	txOpts, err := evm.GetTxOptsWithSigner(client, privateKey)
-	if err != nil {
-		return nil, nil, err
+	var txOpts *bind.TransactOpts
+	if generateRawTxOnly {
+		txOpts = &bind.TransactOpts{
+			From:   from,
```
sukantoraymond, line 333: why can't we just use  crypto.PubkeyToAddress(pk.PublicKey), this way no need to use from in argument

pkg/contract/contract.go
```diff
@@ -319,15 +327,25 @@ func TxToMethod(
...
+		if err != nil {
+			return nil, nil, err
+		}
 	}
 	txOpts.Value = payment
 	tx, err := contract.Transact(txOpts, methodName, params...)
 	if err != nil {
 		trace, traceCallErr := DebugTraceCall(
 			rpcURL,
+			from,
```
sukantoraymond, line 348: same here

pkg/evm/evm.go
```diff
@@ -264,21 +264,31 @@ func FundAddress(
...
+		if err != nil {
+			return nil, err
+		}
 	}
-	address := crypto.PubkeyToAddress(privateKey.PublicKey)
-	gasFeeCap, gasTipCap, nonce, err := CalculateTxParams(client, address.Hex())
+	if from == (common.Address{}) {
+		from = crypto.PubkeyToAddress(privateKey.PublicKey)
+	}
+	gasFeeCap, gasTipCap, nonce, err := CalculateTxParams(client, from.Hex())
```
sukantoraymond, line 291: I don't foresee `from` being an empty address, especially if we are getting "validator manager owner" address on chain anyways. This means that we don't need `from` in argument and we can just use crypto.PubkeyToAddress(privateKey.PublicKey)

pkg/contract/contract.go
```diff
@@ -289,11 +288,20 @@ func ParseSpec(
 	return name, string(abiBytes), nil
 }
 
+func idempotentSigner(
+	_ common.Address,
+	tx *types.Transaction,
+) (*types.Transaction, error) {
+	return tx, nil
+}
+
```
sukantoraymond, line 297: why dont we name this emptySigner

pkg/evm/evm.go
```diff
@@ -264,21 +264,31 @@ func FundAddress(
...
+		privateKey, err = crypto.HexToECDSA(privateKeyStr)
+		if err != nil {
+			return nil, err
+		}
 	}
-	address := crypto.PubkeyToAddress(privateKey.PublicKey)
-	gasFeeCap, gasTipCap, nonce, err := CalculateTxParams(client, address.Hex())
+	if from == (common.Address{}) {
+		from = crypto.PubkeyToAddress(privateKey.PublicKey)
+	}
```
sukantoraymond, line 290: Seems that either privateKeyStr == ""  or from == "", maybe we can just do if else here to prevent unintended overriding of `from` values

pkg/evm/evm.go
```diff
@@ -264,21 +264,31 @@ func FundAddress(
...
 	client ethclient.Client,
+	generateRawTxOnly bool,
+	from common.Address,
 	privateKeyStr string,
 	warpMessage *avalancheWarp.Message,
 	contract common.Address,
 	callData []byte,
 	value *big.Int,
 ) (*types.Transaction, error) {
 	const defaultGasLimit = 2_000_000
```
sukantoraymond, line 277: should we add a check here where either privateKeyStr != "" or from != "" and both can't be empty and both can't be not empty also?

sukantoraymond: same for txtomethod

cmd/blockchaincmd/add_validator.go
```diff
@@ -122,8 +125,9 @@ Testnet or Mainnet.`,
...
+	cmd.Flags().BoolVar(&externalValidatorManagerOwner, "external-validator-manager-owner", false, "validator manager owner is external, make hex dump of ech evm transactions, so they can be signed in a separate flow")
```
sukantoraymond, line 130: unclear what `external-validator-manager-owner ` means, I think we can just rename this to `generate-raw-tx-only`, with desc as "set this value to true when signing validator manager tx outside of cli" or sth similar

sukantoraymond: and say in description for multisig or ledger

cmd/blockchaincmd/convert.go
felipemadero: the PT still supports v1.0.0

pkg/validatormanager/validatormanager.go
felipemadero: same. another PR be needed to check that PoS is supported on ACP99. For the moment this is
v1.0.0/ACP99 for PoA, v1.0.0 for PoS

felipemadero:
> Don't think we need to support both ACP 99 and validator manager 1.0.0 contracts. Supporting both adds additional complexity esp to our sdk. Plus once ACP 99 is released, don't think there's a need for users to use validator manager 1.0.0 contracts. Deferring to @learyce

I believe we currently have users on both sides of the contracts. We may deprecate v1.0.0 in a future release and take
this as a transitioning release.

cmd/blockchaincmd/add_validator.go
felipemadero: The flag is to be used on special situations:
- the user has a multisig setup for the init tx. in this case the multisig (owner) contract does not even have to have balance. the init tx is sent and paid from another account so the address change needs to be considered by CLI
- the user uses a different account for init than for completion tx. particullary, anyone can send the completion tx. same idea, to enable the user to specify which account to use

pkg/vm/create_evm.go
felipemadero: did not tested. can we make another PR for this?

pkg/evm/evm.go
felipemadero: this is for a user that just nows the address but not the private key. eg a multisig.

felipemadero: to be used in general in cases where a user does not now a private key but want to create anyway an unsigned
tx. this evm library is more general than validator manager owners

cmd/blockchaincmd/add_validator.go
felipemadero: changing external-evm-signature, to differentiate from p-chain signatures, but open to other names

pkg/contract/contract.go
felipemadero: idempotent is kind of more expressive IMO. empty can refer to something that generated empty output?

sukantoraymond (approved):

PR #2651, opened 2025-03-04
Upgrading ubuntu 20.04 in github actions

Tonix517:
https://ava-labs.atlassian.net/browse/DPTN-1340

ref: https://github.com/ava-labs/glacier-api/pull/2034

### DIFF
--- a/.github/workflows/assignee.yml
+++ b/.github/workflows/assignee.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   assign:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


jasonatran:
I think we should stick with the latest LTS here.

sukantoraymond:
Good to know that latest passes CI

Tonix517:
> I think we should stick with explicitly naming the latest LTS here.

updated all `ubuntu-latest` to `ubuntu-24.04` as suggested.

sukantoraymond (approved):

PR #2670, opened 2025-03-11
Skip initialize validator manager for contract initValidatorManager

sukantoraymond:
enables `contract initValidatorManager` to skip initialize validator manager using `--skip-init-validator-manager` flag


felipemadero (approved):
LGTM

Issue #2579, opened 2025-01-27
Replace ANR usage with TMPNET for Local Clusters


felipemadero:
## Goal
This Issue is dependent on https://github.com/ava-labs/avalanche-cli/issues/2511

Replace the ANR with Tmpnet as the backend service utilized when spinning up local clusters.

Issue #2528, opened 2025-01-07
[Bug]: Can't spin up multiple local L1s


owenwahlgren:
**Describe the bug**
When trying to deploy multiple L1s locally, only the most recent L1 deployment's RPC is functional.

**To Reproduce**
1. Create chain configs
```bash
avalanche blockchain create chain1 --evm --proof-of-authority --test-defaults --evm-token="TEST" --evm-chain-id=888 
```
```bash
avalanche blockchain create chain2 --evm --proof-of-authority --test-defaults --evm-token="TEST" --evm-chain-id=999 
```

2. Deploy Chain1
```bash
avalanche blockchain deploy chain1 --local
```

3. Note RPC URL then confirm fetching balance for chain1 is functional
```bash
cast balance 0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC --rpc-url=<chain1 rpc>
```
```bash
999999976431274999435251
```
4. Deploy chain2
```bash
avalanche blockchain deploy chain2 --local
```

5. Note RPC URL then confirm fetching balance for chain2 is functional
```bash
cast balance 0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC --rpc-url=<chain2 rpc>
```
```bash
999999976431274999435251
```

6. Try fetching balance on chain1 rpc again
```bash
cast balance 0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC --rpc-url=<chain1 rpc>
```

```bash
Error: error sending request for url (http://127.0.0.1:43791/ext/bc/S3rjJ3oAo2vo1z4qwxDma8ii3JxAEfci17EFSaxsbtTGPKuZf/rpc)

Context:
- Error #0: error sending request for url (http://127.0.0.1:43791/ext/bc/S3rjJ3oAo2vo1z4qwxDma8ii3JxAEfci17EFSaxsbtTGPKuZf/rpc)
- Error #1: client error (Connect)
- Error #2: tcp connect error: Connection refused (os error 111)
- Error #3: Connection refused (os error 111)
```

**Expected behavior**
Both RPCs should be functional

**Operating System**
MacOSX / Ubuntu

felipemadero:
Closed by https://github.com/ava-labs/avalanche-cli/pull/2680

Issue #2091, opened 2024-08-26
Migrating ANR to TMPNET [with subtasks]


felipemadero:
- [ ] Create TMPNET package/service within CLI https://github.com/ava-labs/avalanche-cli/issues/2509
- [ ] Prototype basic local network implementation using tmpnet https://github.com/ava-labs/avalanche-cli/issues/2510
- [ ] Replace ANR usage with tmpnet for local networks https://github.com/ava-labs/avalanche-cli/issues/2511
- [ ] Replace ANR usage with tmpnet for local clusters https://github.com/ava-labs/avalanche-cli/issues/2579

meaghanfitzgerald:
@marun Tooling will schedule time with you to discuss if tmpnet supports all use cases currently fulfilled by ANR

felipemadero:
@meaghanfitzgerald some points/questions:

- I don't believe we need to have both implementations working at the same time (--tmpnet flag): which is the criteria for that? with direct replacement we can directly use the local network and local cluster e2e CI stuff which has pretty good coverage. also, having two options living together is going to complicate the flow implementations. we can just have the branch for that and do the merge into main when it is good enough.
- local cluster implementation is missing in the migration steps:  also uses ANR
- IMO the two main sub issues should be: (1) replace local networks (also having a tmpnet CLI package as a deliberable), (2) replace local clusters (using that deliverable)

felipemadero:
Closed by https://github.com/ava-labs/avalanche-cli/pull/2680

PR #2680, opened 2025-03-12
Localclusters tmpnet

felipemadero:
## Why this should be merged
Closes https://github.com/ava-labs/avalanche-cli/issues/2579 https://github.com/ava-labs/avalanche-cli/issues/2091 https://github.com/ava-labs/avalanche-cli/issues/2528 https://github.com/ava-labs/avalanche-cli/issues/2688

There is also a couple of minor extra features added on it:
- support specifying http and staking port for local cluster l1 validators
- support blockchain aliases on local clusters
- minor improvements in `network status` output


sukantoraymond:
It solves https://github.com/ava-labs/avalanche-cli/issues/2688, does this mean we can have two separate active clusters of local machine validator now at the same time?

sukantoraymond:
I think it does since it also solves https://github.com/ava-labs/avalanche-cli/issues/2528

cmd/blockchaincmd/add_validator.go
```diff
@@ -264,54 +267,35 @@ func addValidator(cmd *cobra.Command, args []string) error {
...
 			return err
 		}
-		nodeIDStr, publicKey, pop, err = node.GetNodeData(nodeInfo.Uri)
+		nodeIDStr, publicKey, pop, err = utils.GetNodeID(node.URI)
 		if err != nil {
 			return err
 		}
-		// update sidecar with new node
-		if err := node.AddNodeInfoToSidecar(&sc, nodeInfo, network); err != nil {
+		if err := app.AddDefaultBlockchainRPCsToSidecar(blockchainName, network, []string{node.URI}); err != nil {
```
sukantoraymond, line 292: why not also add &sc as argument in AddDefaultBlockchainRPCsToSidecar so that there is no need to reload

cmd/blockchaincmd/add_validator.go
```diff
@@ -360,7 +344,15 @@ func addValidator(cmd *cobra.Command, args []string) error {
...
 		sc,
-	)
+	); err != nil {
+		return err
+	}
+	if createLocalValidator {
+		if err := localnet.RefreshLocalClusterAliases(app, localValidatorClusterName); err != nil {
+			return err
+		}
+	}
```
sukantoraymond, line 354: can you write a brief description what this does on top?

cmd/blockchaincmd/convert.go
```diff
@@ -224,22 +219,35 @@ func StartLocalMachine(
 		if network.Kind == models.Mainnet {
 			globalNetworkFlags.UseMainnet = true
 		}
+		nodeSettingsLen := max(len(httpPorts), len(stakingPorts))
```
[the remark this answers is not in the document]
sukantoraymond: should we just err out if they are not equal in len?

cmd/networkcmd/stop.go
```diff
@@ -76,14 +75,20 @@ func stopAndSaveNetwork(flags StopFlags) error {
...
-	if err := node.StopLocalNetworkConnectedCluster(app); err != nil {
+	clusterNames, err := localnet.GetRunningLocalClustersConnectedToLocalNetwork(app)
+	if err != nil {
 		return err
 	}
+	for _, clusterName := range clusterNames {
+		if err := localnet.LocalClusterStop(app, clusterName); err != nil {
+			return err
+		}
+	}
```
sukantoraymond, line 91: should we continue closing the remaining clusters even if there is an error. Create an []error before calling LocalClusterStop and add to []error if any and just print out the error for the cluster name

cmd/nodecmd/local.go
```diff
@@ -240,43 +286,60 @@ func localStartNode(_ *cobra.Command, args []string) error {
...
+		nodeSettings,
 		avaGoVersionSetting,
 		network,
 	)
 }
 
-func localStopNode(_ *cobra.Command, _ []string) error {
-	return node.StopLocalNode(app)
+func localStopNode(_ *cobra.Command, args []string) error {
+	if len(args) == 1 {
```
sukantoraymond, line 297: local stop node cmd has maxargs 1. Why not make it exact 1, so we always have 1 arg

cmd/nodecmd/local.go
```diff
@@ -240,43 +286,60 @@ func localStartNode(_ *cobra.Command, args []string) error {
...
+		return nil
+	}
+	clusterNames, err := localnet.GetRunningLocalClusters(app)
+	if err != nil {
+		return err
+	}
+	if len(clusterNames) == 0 {
+		ux.Logger.PrintToUser("no clusters to stop")
+		return nil
+	}
```
sukantoraymond, line 319: then this can be removed

cmd/nodecmd/local.go
```diff
@@ -129,7 +130,7 @@ func newLocalStopCmd() *cobra.Command {
 		Use:   "stop",
 		Short: "Stop local node",
 		Long:  `Stop local node.`,
-		Args:  cobra.ExactArgs(0),
+		Args:  cobra.MaximumNArgs(1),
```
sukantoraymond, line 133: why not make it exact 1, since it fails when args = 0 fails

felipemadero:
> I think it does since it also solves #2528

Both issues are pretty similar. Answer is yes we can have two separate local clusters at the same time.

cmd/blockchaincmd/add_validator.go
felipemadero: that is going to assume the sidecar is always going to be already loaded, which may not be the case in general.
 I can make this function to return the new sidecar

cmd/blockchaincmd/convert.go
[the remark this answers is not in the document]
felipemadero: users are more interested in keeping the rpc port unchanged, so are more prone to just set http port.
but for completeness we also enable them to specify staking if desired so

pkg/utils/common.go
```diff
@@ -455,6 +456,25 @@ func GetChainIDs(endpoint string, chainName string) (string, string, error) {
 	return "", "", fmt.Errorf("%s not found on primary network blockchains", chainName)
 }
 
+func GetNodeID(endpoint string) (
+	string, // nodeID
+	string, // public key
+	string, // PoP
```
sukantoraymond, line 462: should we just return a nodeID object here, instead of three strings?

pkg/ux/output.go
```diff
@@ -90,3 +90,6 @@ func ConvertToStringWithThousandSeparator(input uint64) string {
 	s := p.Sprintf("%d", input)
 	return strings.ReplaceAll(s, ",", "_")
 }
+
+func NoPrint(string, ...interface{}) {
+}
```
sukantoraymond, line 95: whats this for?

pkg/localnet/avalanchegoDbDownload.go
```diff
@@ -0,0 +1,94 @@
...
+	log logging.Logger,
+	printFunc func(msg string, args ...interface{}),
+) error {
+	// only for fuji
+	if clusterNetwork.Kind != models.Fuji {
+		return nil
+	}
+	network := network.FujiNetwork()
+	log.Info("downloading public archive for network", zap.String("clustername", clusterNetwork.Name()))
+	printFunc("Downloading public archive for network %s", clusterNetwork.Name())
```
sukantoraymond, line 34: Do we need two print statements here?

pkg/localnet/localnet.go
```diff
@@ -185,22 +184,41 @@ func LocalNetworkHealth(
...
-		if !blockchainBootstrapped {
-			return pChainBootstrapped, false, nil
+			blockchainBootstrappedOnSomeCluster := false
+			for _, clusterName := range clusters {
+				if hasValidators, err := LocalClusterHasValidatorsForSubnet(app, clusterName, blockchain.SubnetID); err != nil {
+					return pChainBootstrapped, false, err
+				} else if !hasValidators {
+					continue
+				}
+				blockchainBootstrapped, err := IsLocalClusterBlockchainBootstrapped(app, clusterName, blockchain.ID.String(), blockchain.SubnetID)
```
sukantoraymond, line 204: why not just assign blockchainBootstrappedOnSomeCluster here

pkg/localnet/localcluster.go
```diff
@@ -0,0 +1,601 @@
...
+}
+
+// Gets a list of clusters filtered by running status, network model, and
+// validated blockchains
+func GetFilteredLocalClusters(
+	app *application.Avalanche,
+	running bool,
+	network models.Network,
+	blockchainName string,
+) ([]string, error) {
```
sukantoraymond, line 312: Seems that this function is used whenever running = true, should we remove this from argument?

sukantoraymond:
```
AVL-7H2W7V:avalanche-cli raymondsukanto$ ./bin/avalanche node local destroy newSubnet-local-node-fuji
AVL-7H2W7V:avalanche-cli raymondsukanto$ ./bin/avalanche node local destroy newSubnet2-local-node-fuji
```

Seems that node local destroy doesn't output anything?

sukantoraymond:
We should prob output that its destroyed

sukantoraymond:
```
AVL-7H2W7V:avalanche-cli raymondsukanto$ ./bin/avalanche blockchain deploy localnetowrk
Found running cluster fujiSubnet-local-node-fuji. Will restart after migration.
Unexpected format on cluster at /Users/raymondsukanto/.avalanche-cli/local/local-local-node-local-network. Please manually recover
Migrating fujiSubnet-local-node-fuji
Migrating newSubnet-local-node-local-network
Restarting cluster fujiSubnet-local-node-fuji.
```

After i upgrade to a new cli release using this PR, If i want to deploy a new local network, seems that I have to wait for migration if i have any local nodes left (which people will do if they dont call node local destroy), can we give them an option if they want to migrate? If they dont' want to migrate, we will just destroy these local nodes instead of migrating them

pkg/ux/output.go
felipemadero: ups... not used anymore

felipemadero: nice catch

pkg/localnet/avalanchegoDbDownload.go
felipemadero: removing the log one

pkg/localnet/localnet.go
felipemadero: we iterate over all the clusters , the last one can be false and overwrite some previous true value. I believe it is cleaner this way that to assign to blockchainBootstrappedOnSomeCluster, check the value and continue if true

pkg/localnet/localcluster.go
felipemadero: I would rather prefer to keep the option available for the future. It is useful for debugging and dev tasks.

pkg/localnet/localnet.go
felipemadero: ok will implement the version with break

pkg/utils/common.go
felipemadero: I prefer this reformatting to be done on another PR

felipemadero: https://github.com/ava-labs/avalanche-cli/issues/2702

sukantoraymond (approved):

PR #2681, opened 2025-03-12
fix: install script location

ThatOneCalculator:
## Why this should be merged

- Sets default location to `~/.local/bin` (standard location on Linux, macOS, etc)

- Removes trailing slash from install path (prevents cases like `install.sh -b /usr/local/bin/` producing `ava-labs/avalanche-cli info installed /usr/local/bin//avalanche`)

## How this works

- Sets `BINDIR` to itself but with trailing slash removed if it exists
- Then sets `BINDIR` to default (`~/.local/bin`) if it's an empty string

## How this was tested

- Ran install script

## How is this documented

- Code comments

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -21,10 +21,9 @@ EOF
 RUN_COMPLETIONS=true
 
 parse_args() {
-  #BINDIR is ./bin unless set be ENV
-  # over-ridden by flag below
-
-  BINDIR=${BINDIR:-~/bin}
+  # BINDIR is ~/.local/bin unless set be overridden by the -b flag (trailing slash is removed) 
+  BINDIR=${BINDIR%/}
+  BINDIR=${BINDIR:-~/.local/bin}
   while getopts "b:ndh?x" arg; do
     case "$arg" in
       b) BINDIR="$OPTARG" ;;


scripts/install.sh
```diff
@@ -21,10 +21,9 @@ EOF
 RUN_COMPLETIONS=true
 
 parse_args() {
-  #BINDIR is ./bin unless set be ENV
-  # over-ridden by flag below
-
-  BINDIR=${BINDIR:-~/bin}
```
felipemadero, line 27: Hi! Can you keep the current default location? It is already used on many machines and on its time it was decided to be set there after considering alternatives. ~/.local/bin is meant as a standard location on linux/mac, but there are many tools that just don't use it. Also, being hidden, it is not so easy to remember or locate.

ThatOneCalculator: Not too sure about it being "easy or remember to locate", since `~/.local/bin` is the canonical location for `XDG_BIN_HOME`, meaning most OSes already have it in `$PATH` -- with `~/bin` (non-standard location seemingly made just for `avalanche-cli`, I have not seen a single tool that uses this), users have to *manually* add `~/bin` to their `$PATH`. If you still feel that `~/bin` is the best choice, I'll change it back, but I'd appreciate it if you would consider `~/.local/bin` one last time as to not break standard XDG convention.

felipemadero: Yes please change it back. Sorry for the delay here.

felipemadero: And thanks so much for the contribution and the feedback

ThatOneCalculator: @felipemadero reverted it back to `~/bin`

PR #2690, opened 2025-03-17
[Test Coverage Improvement] Validator Manager Flags

sukantoraymond:
This PR is the first PR in Test Coverage Improvement Epic. It finds common flags across commands and consolidates them into a common file to prevent inconsistencies in flag name and values. 

Consolidate flags below across all commands into 1 file.

```
AggregatorLogLevel
AggregatorLogToStdout
rpcURL
```

Removes the flags below, as they are no longer needed. They were initially implemented for the earliest implementation of ACP-77.

```
AggregatorExtraEndpoints
AggregatorAllowPrivatePeers
```

This PR is tested by manually creating an L1 on Fuji using local machine as bootstrap validator as well as using remote AWS nodes as bootstrap validators.


cmd/blockchaincmd/change_weight.go
```diff
@@ -51,7 +53,7 @@ The L1 has to be a Proof of Authority L1.`,
 		Args: cobrautils.ExactArgs(1),
 	}
 	networkoptions.AddNetworkFlagsToCmd(cmd, &globalNetworkFlags, true, networkoptions.DefaultSupportedNetworkOptions)
-
+	flags.AddValidatorManagerFlagsToCmd(cmd, &validatorManagerFlags, true)
```
felipemadero, line 56: lets stop using the same variable in different commands

cmd/blockchaincmd/add_validator.go
```diff
@@ -53,21 +55,17 @@ var (
...
 	errMutuallyExclusiveStartOptions    = errors.New("--use-default-start-time/--use-default-validator-params and --start-time are mutually exclusive")
 	errMutuallyExclusiveWeightOptions   = errors.New("--use-default-validator-params and --weight are mutually exclusive")
 	ErrNotPermissionedSubnet            = errors.New("subnet is not permissioned")
-	aggregatorExtraEndpoints            []string
-	aggregatorAllowPrivatePeers         bool
 	clusterNameFlagValue                string
 	createLocalValidator                bool
 	externalValidatorManagerOwner       bool
 	validatorManagerOwner               string
+	validatorManagerFlags               flags.ValidatorManagerFlags
```
felipemadero, line 68: lets stop using the same variable in different commands. we can take the opportunity here
to define a struct per command, with all his flags, and defined a struct instance for the command,
and use it

felipemadero (commented):
I would have preferred to move the flags removal to another PR, it seems an easy step to be done, and we can focus here in the flags redesign phase

sdk/interchain/signature-aggregator.go
```diff
@@ -62,7 +61,7 @@ func createAppRequestNetwork(
 		InfoAPI: &apiConfig.APIConfig{
 			BaseURL: network.Endpoint,
 		},
-		AllowPrivateIPs:  allowPrivatePeers,
+		AllowPrivateIPs:  true,
```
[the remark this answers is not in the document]
sukantoraymond: use a constant for a boolean value?

cmd/contractcmd/init_validator_manager.go
```diff
@@ -28,16 +30,6 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type ValidatorManagerFlags struct {
```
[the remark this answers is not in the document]
sukantoraymond: I dont think we need a struct just for network, private key and (if we are leaving rpc). Esp for network and rpc, they are everywhere and not specific to validator manager

cmd/flags/validator_manager.go
```diff
@@ -0,0 +1,32 @@
...
+type ValidatorManagerFlags struct {
+	RPC         string
+	SigAggFlags SignatureAggregatorFlags
+}
+
+type SignatureAggregatorFlags struct {
+	AggregatorLogLevel    string
+	AggregatorLogToStdout bool
+}
+
```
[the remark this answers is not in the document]
sukantoraymond: good point for aggregator log level to check for specific values only

cmd/nodecmd/local.go
```diff
@@ -66,6 +65,7 @@ var (
 	latestAvagoPreReleaseVersion bool
 	validatorManagerAddress      string
 	useACP99                     bool
+	validatorManagerFlags        flags.ValidatorManagerFlags
```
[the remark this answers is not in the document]
sukantoraymond: If we have all flags of a commands as a struct that is just used locally within the command file, this does not allow for unified testing of what is allowed and not allowed values for all these variables across all commands. To do this, we will have to replicate this test for every single command file. 

I agree that global variables currently cause errors in the codebase, which is why I think we should start having flags that is used across different files as function arguments, which is also part of this PR.

For example in : https://github.com/ava-labs/avalanche-cli/pull/2690/files#diff-96f5e78f4211dda4f8b541558fbdbc314bc85c8394a1a83a018b4012301a4825R776

cmd/flags/validator_manager.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+	"github.com/spf13/cobra"
+)
+
+const (
+	rpcURLFLag                = "rpc"
+	aggregatorLogLevelFlag    = "aggregator-log-level"
+	aggregatorLogToStdoutFlag = "aggregator-log-to-stdout"
+)
+
```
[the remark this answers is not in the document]
sukantoraymond: same point above

cmd/blockchaincmd/add_validator.go
sukantoraymond: combined into flags instead, no more using local variable in different cmd files. This will allow for better testing of values

cmd/flags/validator_manager.go
```diff
@@ -0,0 +1,32 @@
...
+	RPC         string
+	SigAggFlags SignatureAggregatorFlags
+}
+
+type SignatureAggregatorFlags struct {
+	AggregatorLogLevel    string
+	AggregatorLogToStdout bool
+}
+
+func AddValidatorManagerFlagsToCmd(cmd *cobra.Command, flags *ValidatorManagerFlags, addRPCFlag bool) {
```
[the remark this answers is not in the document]
sukantoraymond: addressed

cmd/flags/signature_aggregator.go
```diff
@@ -0,0 +1,55 @@
...
+	cmd.Flags().StringVar(&SigAggFlags.AggregatorLogLevel, aggregatorLogLevelFlag, constants.DefaultAggregatorLogLevel, "log level to use with signature aggregator")
+	cmd.PreRunE = func(cmd *cobra.Command, args []string) error {
+		if err := validateSignatureAggregatorFlags(); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	cmd.Flags().BoolVar(&SigAggFlags.AggregatorLogToStdout, aggregatorLogToStdoutFlag, false, "use stdout for signature aggregator logs")
+}
```
sukantoraymond, line 55: We can do this for all flags that require input validation (e.g. above a certain value, etc)

cmd/flags/blockchain.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"github.com/spf13/cobra"
+)
+
+const (
+	rpcURLFLag = "rpc"
+)
+
+var RPC string
```
Tonix517, line 13: nit: maybe `RPCUrl` or `RPCEndpoint` for better readability

cmd/blockchaincmd/add_validator.go
```diff
@@ -183,13 +176,13 @@ func addValidator(cmd *cobra.Command, args []string) error {
 	}
 
 	if len(args) == 0 {
-		if rpcURL == "" {
-			rpcURL, err = app.Prompt.CaptureURL("What is the RPC endpoint?", false)
+		if flags.RPC == "" {
```
learyce, line 179: Is there anyway to refactor the flags to have an input argument to be required and for the prompt to be baked into the flag creation/adding to the cmd

cmd/blockchaincmd/change_weight.go
```diff
@@ -51,7 +53,8 @@ The L1 has to be a Proof of Authority L1.`,
 		Args: cobrautils.ExactArgs(1),
 	}
 	networkoptions.AddNetworkFlagsToCmd(cmd, &globalNetworkFlags, true, networkoptions.DefaultSupportedNetworkOptions)
-
+	flags.AddRPCFlagToCmd(cmd)
```
learyce, line 56: This is much cleaner than the below 👍

cmd/contractcmd/init_validator_manager.go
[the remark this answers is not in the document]
learyce: Hey @felipemadero, for the approach for keeping a struct with all the flags, would we be able to still keep all flags defined in a single place?

cmd/contractcmd/init_validator_manager.go
```diff
@@ -63,13 +56,10 @@ func newInitValidatorManagerCmd() *cobra.Command {
...
 	}
-	networkoptions.AddNetworkFlagsToCmd(cmd, &validatorManagerFlags.Network, true, networkoptions.DefaultSupportedNetworkOptions)
-	validatorManagerFlags.PrivateKeyFlags.AddToCmd(cmd, "as contract deployer")
-	cmd.Flags().StringVar(&validatorManagerFlags.rpcEndpoint, "rpc", "", "deploy the contract into the given rpc endpoint")
-	cmd.Flags().StringSliceVar(&validatorManagerFlags.aggregatorExtraEndpoints, "aggregator-extra-endpoints", nil, "endpoints for extra nodes that are needed in signature aggregation")
-	cmd.Flags().BoolVar(&validatorManagerFlags.aggregatorAllowPrivatePeers, "aggregator-allow-private-peers", true, "allow the signature aggregator to connect to peers with private IP")
-	cmd.Flags().StringVar(&validatorManagerFlags.aggregatorLogLevel, "aggregator-log-level", constants.DefaultAggregatorLogLevel, "log level to use with signature aggregator")
-	cmd.Flags().BoolVar(&validatorManagerFlags.aggregatorLogToStdout, "aggregator-log-to-stdout", false, "dump signature aggregator logs to stdout")
+	networkoptions.AddNetworkFlagsToCmd(cmd, &network, true, networkoptions.DefaultSupportedNetworkOptions)
+	privateKeyFlags.AddToCmd(cmd, "as contract deployer")
```
learyce, line 60: What is the "as contract deployer"?

cmd/flags/signature_aggregator.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+var SigAggFlags SignatureAggregatorFlags
+
+type SignatureAggregatorFlags struct {
+	AggregatorLogLevel    string
+	AggregatorLogToStdout bool
+}
+
+func validateSignatureAggregatorFlags() error {
```
learyce, line 25: Let's add a few test cases around how we validate these inputs

cmd/flags/signature_aggregator.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+	"github.com/ava-labs/avalanchego/utils/logging"
+	"github.com/spf13/cobra"
+)
+
+const (
+	aggregatorLogLevelFlag    = "aggregator-log-level"
```
learyce, line 14: 👍

cmd/flags/blockchain.go
```diff
@@ -0,0 +1,17 @@
...
+
+func AddRPCFlagToCmd(cmd *cobra.Command) {
```
learyce, line 15: Let's add a validation function + testcases to this case to confirm only valid RPCs URLs are being passed in.

-We should validate the URL structure
-What are your thoughts on trying to connect to the API to make sure that it's actually able to serve the necessary use case

cmd/flags/signature_aggregator.go
```diff
@@ -0,0 +1,53 @@
...
+			logging.Debug.LowerString(),
+			logging.Trace.LowerString(),
+			logging.Verbo.LowerString(),
+		)
+	}
+	return nil
+}
+
+func AddSignatureAggregatorFlagsToCmd(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&SigAggFlags.AggregatorLogLevel, aggregatorLogLevelFlag, constants.DefaultAggregatorLogLevel, "log level to use with signature aggregator")
```
learyce, line 44: A little bit rusty, but how does &SigAggFlags.AggregatorLogLevel get the actual value?

learyce, line 44: Where are we setting the DefaultAggregatorLogLevel  values for the flags currently?

cmd/nodecmd/local.go
```diff
@@ -477,16 +474,15 @@ func localValidate(_ *cobra.Command, args []string) error {
...
-	extraAggregatorPeers, err := blockchain.GetAggregatorExtraPeers(app, clusterName, []string{})
+	extraAggregatorPeers, err := blockchain.GetAggregatorExtraPeers(app, clusterName)
 	if err != nil {
 		return err
 	}
 	aggregatorLogger, err := utils.NewLogger(
 		constants.SignatureAggregatorLogName,
-		aggregatorLogLevel,
+		flags.SigAggFlags,
 		constants.DefaultAggregatorLogLevel,
```
learyce, line 484: why do we pass in the default log level here if we're already including the flag?

pkg/utils/common.go
```diff
@@ -594,16 +596,15 @@ func PrintUnreportedErrors(
...
 	logDir string,
-	logToStdout bool,
 	print func(string, ...interface{}),
 ) (logging.Logger, error) {
-	logLevel, err := logging.ToLevel(logLevelStr)
+	logLevel, err := logging.ToLevel(sigAggFlags.AggregatorLogLevel)
 	if err != nil {
-		if logLevelStr != "" {
-			print("undefined logLevel %s. Setting %s log to %s", logLevelStr, logName, defaultLogLevelStr)
+		if sigAggFlags.AggregatorLogLevel != "" {
```
learyce, line 606: Do we need this check if we're already validating in the flag creation?

pkg/utils/common.go
```diff
@@ -616,7 +617,7 @@ func NewLogger(
 		},
 		LogLevel: logLevel,
 	}
-	if logToStdout {
+	if sigAggFlags.AggregatorLogToStdout {
```
learyce, line 620: Isn't the log level set either way?

sdk/interchain/signature-aggregator.go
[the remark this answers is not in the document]
learyce: Whats' the change here? Ie what does it mean for us to always allow private ips?

cmd/flags/signature_aggregator.go
sukantoraymond: AggregatorLogLevel is a string in SigAggFlags  struct and we assign it default value of constants.DefaultAggregatorLogLevel unless user passes aggregatorLogLevelFlag in their command

sukantoraymond: constants.DefaultAggregatorLogLevel is defined in https://github.com/ava-labs/avalanche-cli/blob/main/pkg/constants/constants.go#L346

pkg/utils/common.go
sukantoraymond: Good catch. technically not since we are already setting default and already have the validate function

cmd/nodecmd/local.go
sukantoraymond: yup, addressed

cmd/flags/signature_aggregator.go
```diff
@@ -0,0 +1,53 @@
...
+
+var SigAggFlags SignatureAggregatorFlags
+
+type SignatureAggregatorFlags struct {
+	AggregatorLogLevel    string
+	AggregatorLogToStdout bool
+}
+
+func validateSignatureAggregatorFlags() error {
+	if _, err := logging.ToLevel(SigAggFlags.AggregatorLogLevel); err != nil {
```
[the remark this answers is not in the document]
sukantoraymond: AggregatorLogToStdout means that we are printing out signature aggregator logs in terminal instead of saving it in a file. Both cases require log level

cmd/contractcmd/init_validator_manager.go
sukantoraymond: It is to customize the flag description 
```


func (pkf *PrivateKeyFlags) AddToCmd(
	cmd *cobra.Command,
	goal string,
) {
	pkf.fillDefaultFlagNames()
	cmd.Flags().StringVar(
		&pkf.PrivateKey,
		pkf.privateKeyFlagName,
		"",
		fmt.Sprintf("private key to use %s", goal),
	)
	cmd.Flags().StringVar(
		&pkf.KeyName,
		pkf.keyFlagName,
		"",
		fmt.Sprintf("CLI stored key to use %s", goal),
	)

```

sdk/interchain/signature-aggregator.go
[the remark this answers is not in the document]
sukantoraymond: AllowPrivateIPs is a config setting in avalanchego config. It is to enable signature aggregator to connect to peers with private IP

cmd/contractcmd/init_validator_manager.go
learyce: Would it be worth moving this into a constant or no?

cmd/flags/signature_aggregator.go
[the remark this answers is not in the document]
learyce: Okay, makes sense. Thank you

sdk/interchain/signature-aggregator.go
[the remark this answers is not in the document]
felipemadero: I would have preferred a constant defaultAllowPrivatePeersValue on top of this pkg or in sdk constants.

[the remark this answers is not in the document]
felipemadero: @learyce we were having a true value as default but opted the user to set it to false using a flag.
there are situations where the aggregator can be only connectin to public peers and having this 
setting set to false is somewhat safer on that case (I dont know on technical details I believe stephen preferred a false value unless needed). At CLI we default to true because the local validators have private IP.
That's the reason the PR proposed to have it always being true.

cmd/nodecmd/local.go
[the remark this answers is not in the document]
felipemadero: I am not sure on passing flags into inner functions. Most of the functions are to be moved into a SDK and
a cmdline flag related struct seems not much cleaner.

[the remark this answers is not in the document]
felipemadero: I by no means try to avoid unified testing, just want an encompassing object that contains as childs all flag-related objects that are verified with unified testing

cmd/flags/validator_manager.go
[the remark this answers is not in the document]
felipemadero: this point seems more advance, for a future iteration, will keep it as it

cmd/contractcmd/init_validator_manager.go
sukantoraymond: since its a one time occurrence, don't think we need to make it a constant.

cmd/blockchaincmd/add_validator.go
sukantoraymond: added this for rpc url, but only for addValidator. Will make another PR to extend it to other commands

cmd/nodecmd/local.go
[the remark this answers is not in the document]
felipemadero: anyway, probably we can move this req to another step on this redesign

cmd/contractcmd/init_validator_manager.go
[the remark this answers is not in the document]
felipemadero: @learyce the implementation changed since my last comment. Initially, here there was a struct with all the flags:

```
type ValidatorManagerFlags struct {
        Network                     networkoptions.NetworkFlags
        PrivateKeyFlags             contract.PrivateKeyFlags
        rpcEndpoint                 string
        aggregatorLogLevel          string
        aggregatorLogToStdout       bool
        aggregatorExtraEndpoints    []string
        aggregatorAllowPrivatePeers bool
}
```

And a variable

```
var validatorManagerFlags   ValidatorManagerFlags
```

That centralized all the needed flags for the command as an object. That enabled clean discrimination of which are the flags for this command (flag variables usually were dispersed among many files), also disabled secondary effects between command conflicting flag initializations, and finally enabled clean call of the command from another command, by setting the desired flags, avoiding (again) secondary effects.

Previous PR code removed the struct and keep the variables, grouping aggregator variables into a struct defined at flags pkg. Something like this:

```
var (
        Network                       networkoptions.NetworkFlags
        PrivateKeyFlags           contract.PrivateKeyFlags
        rpcEndpoint                 flags.RPC
        aggregatorSettings flags.AggregatorSettings
)
```

That still enabled defining the struct with something like:

```
type ValidatorManagerFlags struct {
        Network                       networkoptions.NetworkFlags
        PrivateKeyFlags           contract.PrivateKeyFlags
        rpcEndpoint                 flags.RPC
        aggregatorSettings flags.AggregatorSettings
}
```

And a variable

```
var validatorManagerFlags   ValidatorManagerFlags
```

Which was my suggestion.

Current PR code has the variables, not only the structs, defined at flags pkg. So there is no way to separate the variable instances between commands. That has the benefit of grouping the flag definitions into a package, but enables secondary effects between conflicting flag initializations (which can be solved by forcing same default initialization value, which is anyway going to change some command semantics), and also enables secondary effects when calling a command from another command (say the called command uses a flag from the calling command, but you really don't want to use that flag on the second call - much cleaner to define the flags that you need, that to check all flags to try to avoid unexpected effects).

That said, I let you decide if the current proposed scheme is good enough for our needs.

cmd/blockchaincmd/remove_validator.go
```diff
@@ -145,30 +144,31 @@ func removeValidator(_ *cobra.Command, args []string) error {
...
+	}
+
 	validatorKind, err := validatorsdk.IsSovereignValidator(network.SDKNetwork(), subnetID, nodeID)
 	if err != nil {
 		return err
 	}
 	if validatorKind == validatorsdk.NonValidator {
 		// it may be unregistered from P-Chain, but registered on validator manager
 		// due to a previous partial removal operation
-		if rpcURL == "" {
```
learyce, line 155: This is great that we can remove these checks!

cmd/blockchaincmd/add_validator.go
felipemadero: there may be some dependence between flag values, are flags that should only be used if some other flag value
is set. this will complicate automatic prompting unless we make something generic that takes this into account.

cmd/contractcmd/init_validator_manager.go
felipemadero: probably we can also move networkoptions and privatekey flags into flags pkg

cmd/flags/blockchain.go
felipemadero: there is a validation function for urls into pkg/prompts/ we may consider to move this into pkg/validation and share.
the validation function checks or not URL API liveness under request

cmd/flags/signature_aggregator.go
felipemadero: probably we need to move default flag values into flags pkg?

pkg/utils/common.go
felipemadero: display level is specific to stdout, if not set, no logs to stdout

cmd/blockchaincmd/add_validator.go
```diff
@@ -186,13 +179,7 @@ func addValidator(cmd *cobra.Command, args []string) error {
 	}
 
 	if len(args) == 0 {
-		if rpcURL == "" {
-			rpcURL, err = app.Prompt.CaptureURL("What is the RPC endpoint?", false)
```
felipemadero, line 190: automatic prompting associated with flags is a very very nice idea, but we need to consider that depending on 
context, flags will or wont be required. and the context is maybe another flags or another prompts.

felipemadero: I totally agree with automatic default initialization, validation, and prompting if missing, but we need input values dependences and relationships into the solution

cmd/blockchaincmd/change_weight.go
```diff
@@ -150,8 +148,8 @@ func setWeight(_ *cobra.Command, args []string) error {
 		BlockchainName: blockchainName,
 	}
 
-	if rpcURL == "" {
-		rpcURL, _, err = contract.GetBlockchainEndpoints(
+	if flags.RPC == "" {
+		flags.RPC, _, err = contract.GetBlockchainEndpoints(
```
felipemadero, line 152: I'm not sure if changing input flags during code execution is the thing we want to do in the future, after the redesign

cmd/blockchaincmd/convert.go
```diff
@@ -268,6 +267,7 @@ func InitializeValidatorManager(
 	validatorManagerAddrStr string,
 	proxyContractOwner string,
 	useACP99 bool,
+	signatureAggregatorFlags flags.SignatureAggregatorFlags,
```
felipemadero, line 270: we need to restrict passing in this flags objects, the limit is the CLI/SDK frontier, but probably we should also
avoid passing in flags objects outside of cmd helpers.

felipemadero (commented):
Making general observations and questions on this design. For @sukantoraymond  @learyce to take into account. But having not strong opinion on either way, the subject is complicated and it is needed some tradeoff between being practical and having a design that takes into account all cases. Will happily agree with team opinion that surely will
keep the balance.

cmd/blockchaincmd/add_validator.go
```diff
@@ -372,6 +364,7 @@ func CallAddValidator(
 	remainingBalanceOwnerAddr string,
 	disableOwnerAddr string,
 	sc models.Sidecar,
+	rpcURL string,
```
artemanosov, line 367: What's the point of passing this and other global variables (flags in this case) to this function as arguments as opposed to just using them directly?

cmd/flags/blockchain.go
```diff
@@ -0,0 +1,49 @@
...
+		return rpcPreRun(cmd, args)
+	}
+}
+
+func ValidateRPC(app *application.Avalanche, rpc *string, cmd *cobra.Command, args []string) error {
+	var err error
+	// TODO: modify check below to extend prompting for rpc to commands other than addValidator
+	if *rpc == "" {
+		if cmd.Name() == "addValidator" && len(args) == 0 {
+			*rpc, err = app.Prompt.CaptureURL("What is the RPC endpoint?", false)
```
artemanosov, line 41: Do we need to call `prompts.ValidateURLFormat(*rpc)` after we get the rpc url from the prompt?

cmd/flags/blockchain.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+		return rpcPreRun(cmd, args)
+	}
+}
+
+func ValidateRPC(app *application.Avalanche, rpc *string, cmd *cobra.Command, args []string) error {
+	var err error
+	// TODO: modify check below to extend prompting for rpc to commands other than addValidator
+	if *rpc == "" {
+		if cmd.Name() == "addValidator" && len(args) == 0 {
```
artemanosov, line 40: Why don't we validate the arg for this command in case it's provided?

cmd/flags/blockchain.go
```diff
@@ -0,0 +1,49 @@
...
+			*rpc, err = app.Prompt.CaptureURL("What is the RPC endpoint?", false)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+	return prompts.ValidateURLFormat(*rpc)
```
artemanosov, line 48: Also, I noticed `CaptureURL` is doing validation (which is not explicitly mentioned, even though I think it should be) by calling `ValidateURL` which calls `ValidateURLFormat` and also calls the url expecting 2xx. I was wondering why don't `ValidateURL` here instead of `ValidateUrlFormat`?

cmd/blockchaincmd/remove_validator.go
```diff
@@ -145,30 +150,31 @@ func removeValidator(_ *cobra.Command, args []string) error {
 		}
 	}
 
+	if sc.Sovereign && removeValidatorFlags.RPC == "" {
```
artemanosov, line 153: Why did we add `sc.Sovereign` check?

cmd/nodecmd/local.go
```diff
@@ -435,13 +437,13 @@ func localValidate(_ *cobra.Command, args []string) error {
 		}
 	}
 
-	if rpcURL == "" {
-		rpcURL, err = app.Prompt.CaptureURL("What is the RPC endpoint?", false)
+	if localValidateFlags.RPC == "" {
+		localValidateFlags.RPC, err = app.Prompt.CaptureURL("What is the RPC endpoint?", false)
```
artemanosov, line 441: Since in RPC flag validation we use command specific logic, I think this validation should be done there (like we do for `add_validator`

cmd/blockchaincmd/add_validator.go
```diff
@@ -52,25 +54,26 @@ var (
...
 	clusterNameFlagValue                string
 	createLocalValidator                bool
 	externalValidatorManagerOwner       bool
 	validatorManagerOwner               string
 	httpPort                            uint32
 	stakingPort                         uint32
+	addValidatorFlags                   BlockchainAddValidatorFlags
 )
 
+type BlockchainAddValidatorFlags struct {
```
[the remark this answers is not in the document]
felipemadero: - to isolate variables among commands with the same flag, because cobra initializes all commands in sequence and
the variable retains only the value of the command that was last initialized (a conflict for command that have different initialization values)
- to be able in a subsequent PR, to make clean calls into the command logic from another command, by passing the flags object to it (see eg cmd/interchaincmd/tokentransaferrercmd/deploy.go for CallDeploy)

felipemadero (commented):
Overall idea LGTM. Will approve once all comments for other team members are addressed.

cmd/blockchaincmd/add_validator.go
felipemadero: the goal is to be able to call this command from another command, and pass to it only the needed values. this concept is best represented in the tokentransferrer deploy cmd, here the implementation is partial. 
secondary goal is to avoid using global variables at all: refactoring of common code, and different dev styles, showed that the risk to use/set global variables among several functions is possible and happens, so, to avoid secondary effects, and difficult to follow up code flow, it is best to just not use them.

artemanosov:
Is it possible to hold off merging this PR please @sukantoraymond @felipemadero . I just have some things I want to test, that might lead to proposing a bit different design. This was agreed with @learyce

felipemadero:
> I strongly believe that we should not mix in special to commands logic inside `preRunE` ("middleware") functions. Main reason is that if we start putting special to commands logic inside “middleware” function, it can grow ugly with many if/else blocks and will be hard to read. Example: 
> 
> ```
> if cmd.Name() == “AddValidator” || cmd.Name() == “RemoveValidator” || cmd.Name() == “CreateBlockchain” {   
>   // do some common validation   
>   if cmd.Name() == “AddValidation” {     
>     // special to AddValidation logic    
>   }    
>   
>   if cmd.Name() == “RemoveValidator” || cmd.Name() == “CreateBlockchain” {    
>     // common logic here   
>   }     
>   
>   if cmd.Name() == “CreateBlockchain” {   
>     // special logic here
>   }   
>     // the rest of the validation
> }
>   
> ```
> 
> While if we decide to use switch for easier read, there will be redundancy if two commands have most of the special validation logic the same but with one small difference. This is a bit exaggerated example, but I think we could run into some similar case. Another drawback of such design is that it hides the details of special validation logic, thus potentially misleading someone who’s gonna read the code, since some validation will be abstracted away in the flag validation "middleware" and not shown in the linear flow inside the command execution function. Finally, it will be easier to make mistakes since validation is done in a separate place from the actual command execution logic, thus some validation or nuances may be missed.
> 
> I propose two alternative approaches for validating input and reducing redundancy that we can go with:
> 
> 1. Hybrid
> 2. Separation of concerns
> 
> `Hybrid`: First, make validation of a flag optional and regulated using an argument. Second, validation of flag inside `preRunE` ("middleware") function should be simple and not require any special logic for a command. Third, create validation functions for each flag in a shared package (ex. ../flags/validation) and use those whenever we need a special logic for validation with turned off validation in “middleware”.
> 
> `Separation of concerns`: Alternative design would be to just separate flags values collection and validation completely. Then validation functions will be defined in shared package and called depending on the need from the command execution functions.
> 
> Example of simple validation function:
> 
> ```
> func ValidateRPCURL(app *application.Avalanche, rpcURL string) (err error) {
> 	if rpcURL == "" {
> 		return errors.New("RPC URL cannot be empty")
> 	}
> 
> 	return prompts.ValidateURL(rpcURL)
> }
> ```
> 
> @sukantoraymond @felipemadero Please let me know what you guys think

Totally agree with the problem. And this comment was very readable +1! 

I would propose something along this line (really the idea here is to continue the discussion):

- Have a  separated shared package for validation functions in general, not only used for flags but also for prompts. Currently we have good amount of validation functions on pkg/prompts/, would rather move it to something like pkg/validations/.

- Have cmd/flags package (maybe move it to pkg/flags?) to provide:
    - a way of keep track of the set of flags of a command, and set up the command with that set
    - standard flag names, datatypes, descriptions, default values
    - provide a way to set a different default value if needed by the command
    - group of flags definition. a group of flags may all relate to the underlying same value, so it may be queried for it, finding if it was indeed defined on flags, and which was the value.    
    - default validation for the full set of flags (eg we want to avoid conflicting flags to be defined at the same time)
   
-   Have pkg/inputs package, that manage both flags, prompts and validations. To be initialized with the cmdline flags. Containing default validation functions for all flags, but that can be set up (specialization) by the command. Containing default prompt function for all flags, but that can set up (specialization) by the command. With a initial step that checks all flags provided in command line to be valid and either ends execution with error or prompts the user for the invalid values (configurable). Afterwards, it works under cmd request, because the cmd knows when and why to ask for an input (that most generally depends on the previous inputs). So the cmd ask inputs pkg for such an input:
 1) if the user did not provided the value on flags.
   1a) if the default flag value is marked as acceptable (configured at inputs pkg level), tries to validate, if validation do not pass, prompts
   1b) if the default flag value is not marked as acceptable, prompt for it
 2) if the user provided the value on flags, try to validate it, if validation do not pass, prompts
  
- Have a way to pass validation functions of interest, located either on pkg/validation, or on some specific-named (if really too specialized, eg validations.go) file belonging to the command code, to the inputs pkg. If not passed, default validation is used.
- Same for prompts.

felipemadero:
> > I strongly believe that we should not mix in special to commands logic inside `preRunE` ("middleware") functions. Main reason is that if we start putting special to commands logic inside “middleware” function, it can grow ugly with many if/else blocks and will be hard to read. Example: > 
> > ```
> > if cmd.Name() == “AddValidator” || cmd.Name() == “RemoveValidator” || cmd.Name() == “CreateBlockchain” { >   
> >   // do some common validation >   
> >   if cmd.Name() == “AddValidation” { >     
> >     // special to AddValidation logic >    
> >   } >  >   
> >   
> >   if cmd.Name() == “RemoveValidator” || cmd.Name() == “CreateBlockchain” { >    
> >     // common logic here >   
> >   }  >  >   
> >   
> >   if cmd.Name() == “CreateBlockchain” { >   
> >     // special logic here
> >   } >   
> >     // the rest of the validation
> > }
> >   
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > While if we decide to use switch for easier read, there will be redundancy if two commands have most of the special validation logic the same but with one small difference. This is a bit exaggerated example, but I think we could run into some similar case. Another drawback of such design is that it hides the details of special validation logic, thus potentially misleading someone who’s gonna read the code, since some validation will be abstracted away in the flag validation "middleware" and not shown in the linear flow inside the command execution function. Finally, it will be easier to make mistakes since validation is done in a separate place from the actual command execution logic, thus some validation or nuances may be missed.
> > I propose two alternative approaches for validating input and reducing redundancy that we can go with:
> > 
> > 1. Hybrid
> > 2. Separation of concerns
> > 
> > `Hybrid`: First, make validation of a flag optional and regulated using an argument. Second, validation of flag inside `preRunE` ("middleware") function should be simple and not require any special logic for a command. Third, create validation functions for each flag in a shared package (ex. ../flags/validation) and use those whenever we need a special logic for validation with turned off validation in “middleware”.
> > `Separation of concerns`: Alternative design would be to just separate flags values collection and validation completely. Then validation functions will be defined in shared package and called depending on the need from the command execution functions.
> > Example of simple validation function:
> > ```
> > func ValidateRPCURL(app *application.Avalanche, rpcURL string) (err error) {
> > 	if rpcURL == "" {
> > 		return errors.New("RPC URL cannot be empty")
> > 	}
> > 
> > 	return prompts.ValidateURL(rpcURL)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @sukantoraymond @felipemadero Please let me know what you guys think
> 
> Totally agree with the problem. And this comment was very readable +1!
> 
> I would propose something along this line (really the idea here is to continue the discussion):
> 
> * Have a  separated shared package for validation functions in general, not only used for flags but also for prompts. Currently we have good amount of validation functions on pkg/prompts/, would rather move it to something like pkg/validations/.
> * Have cmd/flags package (maybe move it to pkg/flags?) to provide:
>   
>   * a way of keep track of the set of flags of a command, and set up the command with that set
>   * standard flag names, datatypes, descriptions, default values
>   * provide a way to set a different default value if needed by the command
>   * group of flags definition. a group of flags may all relate to the underlying same value, so it may be queried for it, finding if it was indeed defined on flags, and which was the value.
>   * default validation for the full set of flags (eg we want to avoid conflicting flags to be defined at the same time)
> * Have pkg/inputs package, that manage both flags, prompts and validations. To be initialized with the cmdline flags. Containing default validation functions for all flags, but that can be set up (specialization) by the command. Containing default prompt function for all flags, but that can set up (specialization) by the command. With a initial step that checks all flags provided in command line to be valid and either ends execution with error or prompts the user for the invalid values (configurable). Afterwards, it works under cmd request, because the cmd knows when and why to ask for an input (that most generally depends on the previous inputs). So the cmd ask inputs pkg for such an input:
> 
> 1. if the user did not provided the value on flags.
>    1a) if the default flag value is marked as acceptable (configured at inputs pkg level), tries to validate, if validation do not pass, prompts
>    1b) if the default flag value is not marked as acceptable, prompt for it
> 2. if the user provided the value on flags, try to validate it, if validation do not pass, prompts
> 
> * Have a way to pass validation functions of interest, located either on pkg/validation, or on some specific-named (if really too specialized, eg validations.go) file belonging to the command code, to the inputs pkg. If not passed, default validation is used.
> * Same for prompts.

Would like clarity something. Cobra requires default values for all cmdline flags. But on CLI code, we use flags in one of this two different ways: 
- 1) either the default value is used, or a cmdline flag overwrites it, without prompting (DEFAULT OR FLAG)
- 2) the default value is not used, so either a cmdline flag was received, or we prompt the user (FLAG OR PROMPT)

That is reason why I made comments such as "the default flag value is not, marked as acceptable" (because cobra always require a default value).

--Note: in 2) an explicit user action is always required only if the value is found to be needed by the cmd flow--

So I believe we can have an input abstraction, of different "types" as given by:
- flag (or group of flags) that can be queried if the value was set, and obtain the value
- default value (or not)
- default validation function 
- default prompt function

Where the last 3 can be modified by the user when instantiating it (eg remove default value or change it)

So the command spec should start by defining a list of input instances for the command. Then use a function to
automatically add the flags associated to those inputs to the cmd.
Then at start of command execution do automatic validation of flags (mutual exclusions and value validations)
Then let the cmd ask for the different inputs on due time.

With some rules:
- if a default value is defined, use validation function on it, if validation fails, quit with error (input package)
- when starting, validate all given flag regarding prohibited flag combinations (flag package)
- when starting, validate all given flag values, if someone is invalid, quit with error (input package)
- when asked for the value: (input package)
  - if default value is defined, use it, unless a flag value was given
  - if no default value is defined, use a flag value if defined, if not, prompt (and validate prompt)

Tell me what do you think on this last piece (ofc validation functions should be on pkg/validations, prompt functions
on pkg/prompts, flag definitions on pkg/flags and input definitions on pkg/inputs)

artemanosov:
First comment from @felipemadero:
I agree, prompts should not have built in validation. I would argue that this structure is better: `internal/input/flags`, `internal/input/prompts` and `internal/input/validation` (this way we indicate that this kind of input processing is specific to this avalanche-cli repo and cannot be imported. Also, everything related to input is grouped under the same path)

The `/flags` should contain all possible flags, while `/validation` would contain validation for them. The reason for having all flags under this package is that if we put there only the flags that are reused, it may create inconsistencies, since some non-reused flags may have validation and may also need a prompt. Also every new flag will have to be checked first to make sure it's does not exist, which in this case I think may cause mistakes and redundancy.

Agreed, all basic flag related things should be defined in that package: `standard flag names, datatypes, descriptions, default values`, as well as `provide a way to set a different default value if needed by the command` and grouping the flags. I am not sure about having default validation in flags package. As I mentioned, we should decide if we want to completely separate the validation from data collection.

for `pkg/inputs`, I think this stuff will be defined in the command files. Since they are command specific, I think the logic should be shown in those files for logical integrity and explicitness. Additionally, based on looking at the code for some time, the validation is often quite complex and depends on some processing done in command execution functions. That is why I think `internal/input` packages should provide just the tools and be simple, while command execution function should use these tools, but maintain the linear explicit logic. That said, I also wonder if it's possible to not call prompts from command execution functions (maybe by using a some input argument to a flag to prompt or not)? I believe we should collect all inputs at first and then do the validation as opposed to getting all the flags, then do some processing, then check if we still need to prompt etc. (`if the user did not provided the value on flags.
1a) if the default flag value is marked as acceptable (configured at inputs pkg level), tries to validate, if validation do not pass, prompts
1b) if the default flag value is not marked as acceptable, prompt for it` I think this can replace calling prompts from commands files)

`if the user provided the value on flags, try to validate it, if validation do not pass, prompts`, I don't think we should prompt in that case, instead just return an error so user retries the command with "better" input. 

Not a fan of higher order functions especially in golang (`Have a way to pass validation functions of interest, located either on pkg/validation, or on some specific-named (if really too specialized, eg validations.go) file belonging to the command code, to the inputs pkg. If not passed, default validation is used.`) I think it's better to just call all the "tools" for processing/validating input inside command execution function (maybe extract it into `validateInput` function) rather than passing a function with special logic to a flag. But that's my subjective opinion, we can debate this.

artemanosov:
Reply to second comment:
If we decide to combine simple validation with input collection than I aggre that we should have flag creating functions accept parameters: defaultValue, prompt, validation. Prompt would be a pointer to string, so if it's nil, means not use prompt. Validation is simple validation and regulated with a boolean flag. End result looks like this:
```
func AddSomeFlag(defaultValue string, promptString *string, useValidation boolean) {
 // add a flag with default value here 
    if promptString != nil { 	
       set prompt function in middleware 
    }

   if useValidation { 	
      set validation function in middleware 
   } 
}
```

cmd/blockchaincmd/add_validator.go
sukantoraymond: this is outdated, alrdy addressed

cmd/flags/blockchain.go
sukantoraymond: I think moving validation things around in the codebase is out of scope for this PR

pkg/utils/common.go
```diff
@@ -614,29 +614,22 @@ func PrintUnreportedErrors(
 
 func NewLogger(
 	logName string,
-	logLevelStr string,
-	defaultLogLevelStr string,
+	aggregatorLogLevel string,
```
felipemadero, line 617: I believe we may want to keep this function a bit more generic, not so aggregator oriented.
It may be possible to use it on another packages. Probably it can be used to set up main app 
log with a bit of refactor. For the moment lets keep the previos more generic arg names

cmd/flags/blockchain.go
```diff
@@ -0,0 +1,49 @@
...
+				return err
+			}
+		}
+		return rpcPreRun(cmd, args)
+	}
+}
+
+func ValidateRPC(app *application.Avalanche, rpc *string, cmd *cobra.Command, args []string) error {
+	var err error
+	// TODO: modify check below to extend prompting for rpc to commands other than addValidator
```
felipemadero, line 38: lets avoid command dependent logic here, and defer how to do automatic prompting to following up work. https://github.com/ava-labs/avalanche-cli/issues/2736

felipemadero:
> Reply to the first comment from @felipemadero: I agree, prompts should not have built in validation. I would argue that this structure is better: `internal/input/flags`, `internal/input/prompts` and `internal/input/validation` (this way we indicate that this kind of input processing is specific to this avalanche-cli repo and cannot be imported. Also, everything related to input is grouped under the same path)
> 
> The `/flags` should contain all possible flags, while `/validation` would contain validation for them. The reason for having all flags under this package is that if we put there only the flags that are reused, it may create inconsistencies, since some non-reused flags may have validation and may also need a prompt. Also every new flag will have to be checked first to make sure it's does not exist, which in this case I think may cause mistakes and redundancy.
> 
> Agreed, all basic flag related things should be defined in that package: `standard flag names, datatypes, descriptions, default values`, as well as `provide a way to set a different default value if needed by the command` and grouping the flags. I am not sure about having default validation in flags package. As I mentioned, we should decide if we want to completely separate the validation from data collection.
> 
> for `pkg/inputs`, I think this stuff will be defined in the command files. Since they are command specific, I think the logic should be shown in those files for logical integrity and explicitness. Additionally, based on looking at the code for some time, the validation is often quite complex and depends on some processing done in command execution functions. That is why I think `internal/input` packages should provide just the tools and be simple, while command execution function should use these tools, but maintain the linear explicit logic. That said, I also wonder if it's possible to not call prompts from command execution functions (maybe by using a some input argument to a flag to prompt or not)? I believe we should collect all inputs at first and then do the validation as opposed to getting all the flags, then do some processing, then check if we still need to prompt etc. (`if the user did not provided the value on flags. 1a) if the default flag value is marked as acceptable (configured at inputs pkg level), tries to validate, if validation do not pass, prompts 1b) if the default flag value is not marked as acceptable, prompt for it` I think this can replace calling prompts from commands files)
> 
> `if the user provided the value on flags, try to validate it, if validation do not pass, prompts`, I don't think we should prompt in that case, instead just return an error so user retries the command with "better" input.
> 
> Not a fan of higher order functions in golang (`Have a way to pass validation functions of interest, located either on pkg/validation, or on some specific-named (if really too specialized, eg validations.go) file belonging to the command code, to the inputs pkg. If not passed, default validation is used.`) I think it's better to just call all the "tools" for processing/validating input inside command execution function (maybe extract it into `validateInput` function) rather than passing a function with special logic to a flag. But that's my subjective opinion, we can debate this.

I believe there is a base for a new design here, at least for default cases of validation and prompting. But the purpose of this PR is not to discuss this. https://github.com/ava-labs/avalanche-cli/issues/2736

Regarding dir location, also to be discussed and addressed somewhere. https://github.com/ava-labs/avalanche-cli/issues/2737

I personally like having all together under a main inputs package, but should put it on pkg, because we already use pkg for a lot of internal stuff, and really it is expected for
sdk/ to cover external usage.

felipemadero (approved):
Preaproving, but lets move command dependent logic back to the command.

cmd/flags/blockchain.go
sukantoraymond: we are moving all input collection to cmd/flags. This will be done in subsequent PRs

pkg/utils/common.go
sukantoraymond: abstracted into NewSignatureAggregatorLoggerNewLogger

PR #2695, opened 2025-03-21
reduce panic chance by checking CallToMethod return len on all cases

felipemadero:
## Why this should be merged

- Introduces contract.GetMethodReturn to be used after calling contract.CallToMethod, to always check returned slice is not empty, and also properly check for return type
- Also removes duplicated function GetRegisteredValidator in favor of GetValidationID


sdk/validator/validator.go
```diff
@@ -70,12 +69,7 @@ func GetValidatorInfo(net network.Network, validationID ids.ID) (platformvm.L1Va
 	return vdrInfo, nil
 }
 
-func GetValidationID(rpcURL string, nodeID ids.NodeID, validatorManagerAddressStr string) (ids.ID, error) {
-	managerAddress := common.HexToAddress(validatorManagerAddressStr)
-	return GetRegisteredValidator(rpcURL, managerAddress, nodeID)
-}
-
-func GetRegisteredValidator(
+func GetValidationID(
```
sukantoraymond, line 72: Can we add description to this function?

// Verifies
 /**
     * @notice Returns a validation ID registered to the given nodeID
     * @param nodeID ID of the node associated with the validation ID
     */

pkg/contract/contract.go
```diff
@@ -604,6 +604,21 @@ func CallToMethod(
 	return out, nil
 }
 
+func GetMethodReturn[T any](methodName string, out []interface{}) (T, error) {
```
sukantoraymond, line 607: GetSmartContractCallResult?

sukantoraymond (approved):

PR #2696, opened 2025-03-21
improve validator manager kind retrieval

felipemadero:
## Why this should be merged
Currently the function used to decide on validator manager kind takes consider any connection 
or execution failure as a proof for one kind of validator management, which is prone to errors.
Replacing with a function that returns the kind (PoA/PoS/Unknown) plus error, instead of just
a boolean por PoA/PoS. Also making explicit call to the function used to decide if PoS, instead of
including the function inside the check function itself.


sdk/validatormanager/root.go
```diff
@@ -303,3 +306,25 @@ func InitializeValidatorsSet(
 		uint32(0),
 	)
 }
+
+// ValidatorManagerKind returns models.ProofOfAuthority if validator manager is verified to be Proof of Authority
+// If validator manager is verified to be Proof of Stake, returns models.ProofOfStake
+// In other cases, returns models.UndefinedValidatorManagement and the associated error
+func ValidatorManagerKind(
```
sukantoraymond, line 313: Lets rename this to GetValidatorManagerKind

sukantoraymond (approved):

PR #2711, opened 2025-03-28
make SDK to not depend on pkg/models

felipemadero:
## Why this should be merged
Step on works towards SDK 2.0


sukantoraymond:
Why not entirely remove models.network

felipemadero:
> Why not entirely remove models.network

already set into another issue. let's make this easier step first. which is easier to make, test, understand.
main goal is to free sdk from pkg deps

sukantoraymond (approved):

PR #2712, opened 2025-03-28
Sdk to not use pkg ux

felipemadero:
## Why this should be merged
Step on works towards SDK 2.0


Tonix517 (approved):

sdk/blockchain/blockchain.go
```diff
@@ -391,7 +389,7 @@ func (c *Subnet) InitializeProofOfAuthority(
 		if !errors.Is(err, validatormanager.ErrAlreadyInitialized) {
 			return evm.TransactionError(tx, err, "failure initializing poa validator manager")
 		}
-		ux.Logger.PrintToUser("Warning: the PoA contract is already initialized.")
+		log.Warn("the PoA contract is already initialized")
```
sukantoraymond, line 392: i think this can be log.info. And we can make it:  the PoA contract is already initialized, skipping initializing Proof of Authority contract

sdk/blockchain/blockchain.go
```diff
@@ -440,7 +439,7 @@ func (c *Subnet) InitializeProofOfStake(
 		c.RPC,
 		privateKey,
 	); err != nil {
-		ux.Logger.RedXToUser("failure setting proposer VM on L1: %s", err)
+		log.Error("failure setting proposer VM on L1", zap.Error(err))
```
sukantoraymond, line 442: same here should we return err instead?

sukantoraymond (approved):

sdk/blockchain/blockchain.go
```diff
@@ -376,7 +374,7 @@ func (c *Subnet) InitializeProofOfAuthority(
...
```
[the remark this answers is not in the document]
felipemadero: no. this is done just in case proposer vm was not previously setup. if it fails, it is anyway a good chance
proposer vm was already set up, so just continues.

sdk/blockchain/blockchain.go
felipemadero: same response

PR #2717, opened 2025-04-03
add signal interrupt to local network contexts

felipemadero:
## Why this should be merged
This PR addresses a couple of suggestions by @ARR4N and @martineckardt , together with fixing a cleaning up issue found
while looking at the problem:
- add signal interrupts to local network waiting contexts, so it can be interrupted with ctrl+C
- add messages to tell the user where to find relevant logs in case of context timeout while trying to track a blockchain
- stop the local network in case the local network creation or local network start fails

## How this works

Example of how it is now telling to the user on the logs

Screenshot 2025-04-03 at 6 28 24 AM


sdk/utils/utils.go
```diff
@@ -48,5 +50,6 @@ func GetAPILargeContext() (context.Context, context.CancelFunc) {
 
 // Timed Context
 func GetTimedContext(timeout time.Duration) (context.Context, context.CancelFunc) {
-	return context.WithTimeout(context.Background(), timeout)
+	parent, _ := signal.NotifyContext(context.Background(), os.Interrupt)
+	return context.WithTimeout(parent, timeout)
```
ARR4N, line 54: You can avoid dropping the cancellation by wrapping the two together:

```suggestion
	parent, sigCancel := signal.NotifyContext(context.Background(), os.Interrupt)
	ctx, timeCancel := context.WithTimeout(parent, timeout)
	return ctx, func() {
		sigCancel()
		timeCancel()
	}()
```

Separately, but beyond the scope of this PR, typically `context.Background()` is reserved for `main()` and `init()` while all other functions should receive it. [More info](https://google.github.io/styleguide/go/decisions.html#:~:text=It%20is%20very%20rare%20for%20code%20in%20the%20middle%20of%20a%20callchain%20to%20require%20creating%20a%20base%20context%20of%20its%20own%20using%20context.Background().%20Always%20prefer%20taking%20a%20context%20from%20your%20caller%2C%20unless%20it%E2%80%99s%20the%20wrong%20context.).

ARR4N (commented):
Thank you for adding these features!

felipemadero:
> Thank you for adding these features!

Thanks to you for the good feedbacks

pkg/localnet/helpers.go
```diff
@@ -78,6 +80,19 @@ func TrackSubnet(
 		perNodeBlockchainConfig,
 		wallet,
 	); err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			printFunc("")
+			printFunc("A context timeout has occurred while trying to bootstrap the blockchain.")
+			printFunc("")
+			logPaths, _ := GetTmpNetAvailableLogs(networkDir, blockchainID, false)
+			if len(logPaths) != 0 {
+				printFunc("Please check this log files for more information on the error cause:")
```
sukantoraymond, line 89: these log files

pkg/localnet/helpers.go
```diff
@@ -78,6 +80,19 @@ func TrackSubnet(
...
+				for _, logPath := range logPaths {
+					printFunc("  " + logPath)
```
sukantoraymond, line 91: Can we add unit test for helpers_test.go where we intentionally cause error and verify that we get the expected logs dir

sukantoraymond (approved):

pkg/localnet/helpers.go
felipemadero: added test for GetTmpNetAvailableLogs

PR #2719, opened 2025-04-03
Add error to metrics

felipemadero:
## Why this should be merged
Current CLI metrics don't collect information on failed commands. This PR add err message collection
in a new field, error.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"fmt"
 
 	"github.com/ava-labs/avalanche-cli/pkg/application"
 	"github.com/ava-labs/avalanche-cli/pkg/constants"
@@ -110,9 +111,10 @@ func trackMetrics(app *application.Avalanche, commandPath string, flags map[stri
 	for propertyKey, propertyValue := range flags {
 		telemetryProperties[propertyKey] = propertyValue
 	}
-	_ = client.Enqueue(posthog.Capture{
+	err := client.Enqueue(posthog.Capture{
 		DistinctId: userID,
 		Event:      "cli-command",
 		Properties: telemetryProperties,
 	})
+	fmt.Println("METRICS DUMP", telemetryProperties, err)
 }
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -7,7 +7,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
-	"fmt"
 
 	"github.com/ava-labs/avalanche-cli/pkg/application"
 	"github.com/ava-labs/avalanche-cli/pkg/constants"
@@ -16,7 +15,6 @@ import (
 	"github.com/ava-labs/avalanchego/utils/logging"
 
 	"github.com/posthog/posthog-go"
-	"github.com/spf13/cobra"
 )
 
 // telemetryToken value is set at build and install scripts using ldflags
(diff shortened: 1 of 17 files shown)


VERSION
```diff
@@ -1 +1 @@
-1.8.8
\ No newline at end of file
+VERSION
```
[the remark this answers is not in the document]
felipemadero: not at all. thanks

pkg/metrics/metrics.go
```diff
@@ -66,29 +80,33 @@ func HandleTracking(cmd *cobra.Command, commandPath string, app *application.Ava
...
 }
 
-func trackMetrics(app *application.Avalanche, commandPath string, flags map[string]string) {
+func trackMetrics(app *application.Avalanche, flags map[string]string, cmdErr error) {
 	if telemetryToken == "" {
 		telemetryToken = os.Getenv(constants.MetricsAPITokenEnvVarName)
 	}
+	if telemetryToken == "" && !utils.IsE2E() {
+		app.Log.Warn("no token is configured for sending metrics")
+	}
```
sukantoraymond, line 102: do we need to users to see this?

pkg/metrics/metrics.go
```diff
@@ -66,29 +80,33 @@ func HandleTracking(cmd *cobra.Command, commandPath string, app *application.Ava
...
 	if telemetryToken == "" || utils.IsE2E() {
 		return
 	}
-	client, _ := posthog.NewWithConfig(telemetryToken, posthog.Config{Endpoint: telemetryInstance})
-
-	defer client.Close()
+	client, err := posthog.NewWithConfig(telemetryToken, posthog.Config{Endpoint: telemetryInstance})
+	if err != nil {
+		app.Log.Warn(fmt.Sprintf("failure creating metrics client: %s", err))
+	}
```
sukantoraymond, line 109: same here, why not just print nothing to logs?

pkg/metrics/metrics.go
```diff
@@ -56,7 +57,20 @@ func userIsOptedIn(app *application.Avalanche) bool {
 	return app.Conf.ConfigFileExists() && app.Conf.GetConfigBoolValue(constants.ConfigMetricsEnabledKey)
 }
 
-func HandleTracking(cmd *cobra.Command, commandPath string, app *application.Avalanche, flags map[string]string) {
+func HandleTracking(
+	app *application.Avalanche,
+	flags map[string]string,
+	err error,
+) {
+	if sent {
```
sukantoraymond, line 65: if user calls command 1, sent will be set to true after tracking info is sent. What about when user calls command 2, since sent would have been set to true here, will the tracking info be sent?

sukantoraymond: I guess best way to test this is see if we can see the second command on the dashboard after first command is called

pkg/metrics/metrics.go
felipemadero: this is only going into avalanche.log, not stdout. it is useful for debugging. not expected to be printed on releases.

pkg/metrics/metrics.go
felipemadero: this is only going into avalanche.log, not stdout. we are seeing that not all events seem to be sent, so we need to have some logs on this, so we can see if something is happening and maybe try again based on the error

pkg/metrics/metrics.go
```diff
@@ -110,9 +132,15 @@ func trackMetrics(app *application.Avalanche, commandPath string, flags map[stri
...
 		DistinctId: userID,
 		Event:      "cli-command",
 		Properties: telemetryProperties,
-	})
+	}
+	if err := client.Enqueue(event); err != nil {
+		app.Log.Warn(fmt.Sprintf("failure sending metrics %#v: %s", telemetryProperties, err))
+	}
+	if err := client.Close(); err != nil {
+		app.Log.Warn(fmt.Sprintf("failure closing metrics client %#v: %s", telemetryProperties, err))
```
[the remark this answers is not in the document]
felipemadero: this is only going into avalanche.log, not stdout. we are seeing that not all events seem to be sent, so we need to have some logs on this, so we can see if something is happening and maybe try again based on the error

sukantoraymond (approved):

PR #2727, opened 2025-04-07
add error boolean to metrics, together with some renamings

felipemadero:
## Why this should be merged
enables easier grouping and analysis of errors on the dashboards. also follows up some renamings
asked by product manager.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -117,14 +117,16 @@ func trackMetrics(app *application.Avalanche, flags map[string]string, cmdErr er
 
 	telemetryProperties := make(map[string]interface{})
 	telemetryProperties["command"] = app.Cmd.CommandPath()
-	telemetryProperties["version"] = version
+	telemetryProperties["cli_version"] = version
 	telemetryProperties["os"] = runtime.GOOS
-	telemetryProperties["error"] = ""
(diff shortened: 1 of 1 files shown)


pkg/metrics/metrics.go
```diff
@@ -117,14 +117,16 @@ func trackMetrics(app *application.Avalanche, flags map[string]string, cmdErr er
...
-	telemetryProperties["version"] = version
+	telemetryProperties["cli_version"] = version
 	telemetryProperties["os"] = runtime.GOOS
-	telemetryProperties["error"] = ""
+	telemetryProperties["was_successful"] = true
+	telemetryProperties["error_msg"] = ""
 	if cmdErr != nil {
-		telemetryProperties["error"] = cmdErr.Error()
+		telemetryProperties["was_successful"] = false
+		telemetryProperties["error_msg"] = cmdErr.Error()
```
sukantoraymond, line 126: why can't we just check if error_msg is empty to check if there is error

sukantoraymond: meaning there is no need for was_successful

felipemadero: I see. It is for convenience in using the metrics analytics tool, that works better with simple boolean on basic dashboards

PR #2729, opened 2025-04-08
Fix mainnet key error msg

sukantoraymond:
Clarifies error message to use ledger on mainnet


Tonix517 (approved):

pkg/keychain/keychain.go
```diff
@@ -30,7 +30,8 @@ const (
 
 var (
 	ErrMutuallyExlusiveKeySource = errors.New("key source flags --key, --ewoq, --ledger/--ledger-addrs are mutually exclusive")
-	ErrStoredKeyOrEwoqOnMainnet  = errors.New("key sources --key, --ewoq are not available for mainnet operations")
+	ErrStoredKeyOnMainnet        = errors.New("--key flag is not supported for mainnet operations, please use ledger instead")
+	ErrNonEwoqKeyOnMainnet       = errors.New("key source --ewoq is not available for mainnet operations")
```
felipemadero, line 34: we can also advice to use ledger here

felipemadero (approved):

PR #2734, opened 2025-04-11
Check that custom dependency version is at least equal to min version defined in latest.json

sukantoraymond:

### DIFF
--- a/cmd/blockchaincmd/deploy.go
+++ b/cmd/blockchaincmd/deploy.go
@@ -477,9 +477,11 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
 
 	if network.Kind == models.Local {
 		app.Log.Debug("Deploy local")
-
+		// TODO: place this check in how we handle input and flags for dependency versioning
+		if err = dependencies.CheckVersionIsOverMin(app, constants.AvalancheGoRepoName, network, userProvidedAvagoVersion); err != nil {
+			return err
+		}
 		avagoVersion := userProvidedAvagoVersion
-
 		if avagoVersion == constants.DefaultAvalancheGoVersion && avagoBinaryPath == "" {
 			avagoVersion, err = dependencies.GetLatestCLISupportedDependencyVersion(app, constants.AvalancheGoRepoName, network, &sidecar.RPCVersion)
 			if err != nil {
--- a/pkg/dependencies/min_version.go
+++ b/pkg/dependencies/min_version.go
@@ -1,0 +1,39 @@
+// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package dependencies
+
+import (
+	"encoding/json"
+	"fmt"
+	"golang.org/x/mod/semver"
+
+	"github.com/ava-labs/avalanche-cli/pkg/models"
+
+	"github.com/ava-labs/avalanche-cli/pkg/application"
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+)
+
+func CheckVersionIsOverMin(app *application.Avalanche, dependencyName string, network models.Network, version string) error {
+	dependencyBytes, err := app.Downloader.Download(constants.CLILatestDependencyURL)
+	if err != nil {
+		return err
+	}
+
+	var parsedDependency models.CLIDependencyMap
+	if err = json.Unmarshal(dependencyBytes, &parsedDependency); err != nil {
+		return err
+	}
+
+	switch dependencyName {
+	case constants.AvalancheGoRepoName:
+		// version has to be at least higher than minimum version specified for the dependency
+		minVersion := parsedDependency.AvalancheGo[network.Name()].MinimumVersion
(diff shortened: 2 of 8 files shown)


pkg/dependencies/min_version.go
```diff
@@ -0,0 +1,42 @@
...
+	if err = json.Unmarshal(dependencyBytes, &parsedDependency); err != nil {
+		return err
+	}
+
+	switch dependencyName {
+	case constants.AvalancheGoRepoName:
+		// version has to be at least higher than minimum version specified for the dependency
+		minVersion := parsedDependency.AvalancheGo[network.Name()].MinimumVersion
+		versionComparison := semver.Compare(version, minVersion)
+		fmt.Printf("versionComparison %s, %s, result %s \n", version, minVersion, strconv.Itoa(versionComparison))
```
felipemadero, line 34: debug print?

pkg/dependencies/min_version.go
```diff
@@ -0,0 +1,42 @@
...
+	"golang.org/x/mod/semver"
+
+	"github.com/ava-labs/avalanche-cli/pkg/models"
+
+	"github.com/ava-labs/avalanche-cli/pkg/application"
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+)
+
+func CheckVersionIsOverMin(app *application.Avalanche, dependencyName string, network models.Network, version string) error {
+	dependencyBytes, err := app.Downloader.Download(constants.CLILatestDependencyURL)
```
felipemadero, line 19: should you just return nil if version is empty? eg local networks

pkg/dependencies/min_version.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	switch dependencyName {
+	case constants.AvalancheGoRepoName:
+		// version has to be at least higher than minimum version specified for the dependency
+		minVersion := parsedDependency.AvalancheGo[network.Name()].MinimumVersion
+		versionComparison := semver.Compare(version, minVersion)
+		fmt.Printf("versionComparison %s, %s, result %s \n", version, minVersion, strconv.Itoa(versionComparison))
+		if versionComparison == -1 {
+			return fmt.Errorf("minimum version of %s that is supported by CLI is %s", dependencyName, minVersion)
```
felipemadero, line 36: can you also add the given version in the error msg?

pkg/dependencies/min_version.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+		return nil
+	default:
+		return fmt.Errorf("can't check minimum version for dependency: %s", dependencyName)
```
felipemadero, line 40: can you say that check for minimum version is unsupported for the dependency

cmd/blockchaincmd/deploy.go
```diff
@@ -483,9 +483,11 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
```
felipemadero, line 483: you can also add checks to network start and node create, and use the empty string in the json. in some
case we may want to be able to also limit those

pkg/dependencies/min_version.go
sukantoraymond: removed

pkg/dependencies/min_version.go
sukantoraymond: Its not returning error right now if its local networks. The test in min_version_test for local network shows that.

cmd/blockchaincmd/deploy.go
sukantoraymond: network start is for local network and we don't have checks for min version for avalanchego

cmd/blockchaincmd/deploy.go
```diff
@@ -483,9 +483,11 @@ func deployBlockchain(cmd *cobra.Command, args []string) error {
 
 	if network.Kind == models.Local {
 		app.Log.Debug("Deploy local")
-
+		// TODO: place this check in how we handle input and flags for dependency versioning
+		if err = dependencies.CheckVersionIsOverMin(app, constants.AvalancheGoRepoName, network, userProvidedAvagoVersion); err != nil {
```
sukantoraymond, line 487: now that i think about it this line is not required, its for local network

cmd/blockchaincmd/deploy.go
sukantoraymond: addressed for node create

felipemadero: can you ask check for min version also in local network? it seems to be also desirable. also good for e2e

sukantoraymond: Do we need it for local network? The main reason it is added for Fuji & Mainnet is because nodes are required to have at least a specific version of avalanchego to interact with these networks. 

In the case of local network, there is no such limitation, so users are free to use whatever version of AvalancheGo that they want (including older versions).

We should have E2E but I think its better to place it for Fuji E2E

felipemadero (approved):

PR #2739, opened 2025-04-14
enable using stored keys on mainnet

felipemadero:
## Why this should be merged
Recently a couple user cases showed the need to be flexible on this. For example, regarding executing
flows on a cloud node.

This also fixes a lint issue on main


pkg/keychain/keychain.go
```diff
@@ -157,15 +155,6 @@ func GetKeychainFromCmdLineFlags(
 				return nil, err
 			}
 		}
-	case network.Kind == models.Mainnet:
-		// mainnet requires ledger usage
-		if keyName != "" {
```
Tonix517, line 162: do we need to keep this check?

Tonix517 (approved):

pkg/keychain/keychain.go
felipemadero: we don't need. this PR enables using stored keys on mainnet

sukantoraymond:
Screenshot 2025-04-16 at 11 47 25 AM

felipemadero:
> ```
> AVL-7H2W7V:avalanche-cli raymondsukanto$ ./bin/avalanche key transfer
> ✔ Mainnet
> ✔ P-Chain
> ✔ C-Chain
> ✔ Use stored key
> ✔ Amount to send (AVAX units): 0.1█
> ✔ Get address from an existing stored key (created from avalanche key create or avalanche key import)
> ✔ newTestKey
> Error: received status code: 429
> ```
> 
> Screenshot 2025-04-16 at 12 13 57 PM

this is most properly a key transfer issue on mainnet, separated from this PR. Can you create a new issue?

felipemadero:
> Screenshot 2025-04-16 at 11 47 25 AM

this is most properly a key list issue on mainnet, separate from this PR. can you create an issue for it?

sukantoraymond:
i dont think its separate. If we want to enable stored keys on mainnet then users should be able to list and transfer using stored keys on mainnet.

sukantoraymond:
Otherwise, when we make a release, and users are unable to create and transfer funds to their keys, they can't deploy blockchain on mainnet anyways

felipemadero:
> Otherwise, when we make a release, and users are unable to create and transfer funds to their keys, they can't deploy blockchain on mainnet anyways

It is separate in the sense this was happening before the PR. If a user wants to fund some key, can also use

PR #2740, opened 2025-04-14
Subnet evm min version check

sukantoraymond:


sukantoraymond:
As I am working through this PR, I realized that we don't even need subnet-evm version defined in sidecar.json. 
Since sidecar.VMVersion is only used for local machine / remote tracking a blockchain, we should find out the avalanchego version of the blockchain that we want to sync to and just download the latest subnet-evm binary that is supported by avalanchego version. This is compared to our current approach of looking up sidecar.VMVersion to get subnet evm version to download.

The benefit to this approach is that we avoid an issue where a blockchain's Subnet EVM has been upgraded, but not reflected in local sidecar, and when we try to sync to that blockchain, we will install the incorrect version of Subnet-EVM

pkg/node/sync.go
```diff
@@ -29,6 +32,16 @@ func SyncSubnet(app *application.Avalanche, clusterName, blockchainName string,
 	if _, err := subnet.ValidateSubnetNameAndGetChains(app, []string{blockchainName}); err != nil {
 		return err
 	}
+	sc, err := app.LoadSidecar(blockchainName)
+	if err != nil {
+		return err
+	}
+	if sc.VM == models.SubnetEvm {
+		if err = dependencies.CheckVersionIsOverMin(app, constants.SubnetEVMRepoName, clusterConfig.Network, sc.VMVersion); err != nil {
+			ux.Logger.PrintToUser(dependencies.UpdateSubnetEVMInstruction)
```
felipemadero, line 41: nice

pkg/localnet/helpers.go
```diff
@@ -43,6 +46,12 @@ func TrackSubnet(
 		blockchainConfig []byte
 		subnetConfig     []byte
 	)
+	if sc.VM == models.SubnetEvm {
+		if err = dependencies.CheckVersionIsOverMin(app, constants.SubnetEVMRepoName, networkModel, sc.VMVersion); err != nil {
+			ux.Logger.PrintToUser(dependencies.UpdateSubnetEVMInstruction)
```
felipemadero, line 51: use printFunc

pkg/dependencies/min_version.go
```diff
@@ -15,6 +14,8 @@ import (
 	"github.com/ava-labs/avalanche-cli/pkg/constants"
 )
 
+var UpdateSubnetEVMInstruction = "To update the blockchain's Subnet-EVM version, call avalanche blockchain upgrade vm <blockchainName> --config --version <version>"
```
felipemadero, line 17: please don't use blockchain upgrade vm, but blockchain configure. it is planned to move everything, upgrades and vm binary/source changes, into blockchain configure

felipemadero (commented):
please add the min check to create also

felipemadero (approved):
preapproving. please address comment

Issue #2421, opened 2024-12-10
[Bug]: Non deterministic failure to create blockchain tx after creating subnet tx


felipemadero:
**Describe the bug**
Error msg obtained by user after executing `blockchain deploy ..` of a non sovereign blockchain on mainnet

```
we created the configurations and choose 3 Control keys and 2 as threshold.
However, after just signing with one ledger wallet the subnet create transaction was issued and create blockchain transaction didn’t follow up.
Subnet has been created with ID: u6Rc43d1GwYP8t14vKsLMVg9VDHPDxisbz4pm92tnUdA22NyT Error: subnet tx u6Rc43d1GwYP8t14vKsLMVg9VDHPDxisbz4pm92tnUdA22NyT query error: failed to decode client response: not found
u6Rc43d1GwYP8t14vKsLMVg9VDHPDxisbz4pm92tnUdA22NyT - subnet was created and 1 avax was deducted as fees.
```

**To Reproduce**
avalanche blockchain create name --soverign=false
avalanche blockchain deploy name --mainnet

**Expected behavior**
It is not deterministic

**Screenshots**

**Logs**
If applicable, please include the relevant logs that indicate a problem.

**Operating System**
Which OS you used to reveal the bug.

**Additional context**

meaghanfitzgerald:
Was this resolved?

PR #2743, opened 2025-04-15
add wait after cleaning wallet cache

felipemadero:
## Why this should be merged
Closes https://github.com/ava-labs/avalanche-cli/issues/2421

Issue is that wallet retrieves utxos that are about to be marked as spent for a previous operation. This PR
add a wait after cleaning the wallet, so when it is refreshed, it does not retrieve such utxos.

Also fixes a lint issue on main


pkg/subnet/public.go
```diff
@@ -626,6 +626,8 @@ func (d *PublicDeployer) loadWallet(subnetIDs ...ids.ID) (*primary.Wallet, error
 
 func (d *PublicDeployer) CleanCacheWallet() {
 	d.wallet = nil
```
Tonix517, line 628: question: i'm not seeing this cached `d.wallet` is being used in this file. it is all a local `wallet` var is being used in other functions. for my understanding: what does `d.wallet` do in this file?

felipemadero: It is returned by loadCacheWallet, which is called by many functions here.

Tonix517: ok got it. missed the part that `d.wallet` is returned.

felipemadero: this code needs cleanup, and using d.wallet instead of loadCacheWallet is one of those. probably that can wait until refactoring this into SDK, but feel free to make a PR for this.

PR #2746, opened 2025-04-16
make public import to support SOV L1s

felipemadero:
## Why this should be merged
Closes https://github.com/ava-labs/avalanche-cli/issues/2704

Currently, `avalanche blockchain import public` only supports importing non SOV blockchains.
As the trend is to only support L1, a change is needed here.

This also fixes `avalanche blockchain validators` so as to correctly tell is the validator is SOV or NOSOV 

This also removes local network version of helper to get subnet validators, changing also `avalanche blockchain validators` part related to this.


cmd/blockchaincmd/import_public.go
```diff
@@ -70,6 +71,7 @@ plag.`,
 		"the blockchain ID",
 	)
 	cmd.Flags().StringVar(&rpcURL, "rpc", "", "rpc endpoint for the blockchain")
+	cmd.Flags().BoolVar(&noRPCAvailable, "no-rpc-available", false, "use this when an RPC if offline and can't be accesed")
```
Tonix517, line 74: do we want to check if user sets invalid option combination here, i.e. `no-rpc-available` as true, but non empty `rpc`

cmd/blockchaincmd/import_public.go
```diff
@@ -48,12 +49,12 @@ func newImportPublicCmd() *cobra.Command {
 
 By default, an imported Blockchain
 doesn't overwrite an existing Blockchain with the same name. To allow overwrites, provide the --force
-flag.`,
+plag.`,
```
Tonix517, line 52: `flag`?

cmd/blockchaincmd/import_public.go
```diff
@@ -221,62 +180,103 @@ func importPublic(*cobra.Command, []string) error {
...
+				return models.Sidecar{}, nil, fmt.Errorf("could not obtain validator manager type: %w", err)
+			}
+			printFunc("  Validation Kind: %s", sc.ValidatorManagement)
+			if sc.ValidatorManagement == validatormanagertypes.ProofOfAuthority {
+				owner, err := contract.GetContractOwner(rpcURL, common.HexToAddress(validatorManagerAddress))
+				if err != nil {
+					return models.Sidecar{}, nil, err
+				}
+				sc.ValidatorManagerOwner = owner.String()
+				printFunc("  Validator Manager Owner: %s", sc.ValidatorManagerOwner)
```
Tonix517, line 276: nit: move this new block to its own member func

cmd/blockchaincmd/import_public.go
```diff
@@ -69,6 +70,8 @@ flag.`,
 		"",
 		"the blockchain ID",
 	)
+	cmd.Flags().StringVar(&rpcURL, "rpc", "", "rpc endpoint for the blockchain")
+	cmd.Flags().BoolVar(&noRPCAvailable, "no-rpc-available", false, "use this when an RPC if offline and can't be accesed")
```
sukantoraymond, line 74: Seems that noRPCAvailable flag needs to be set to true so that importBlockchain can:
````
if rpcIsAvailable {
			sc.ValidatorManagement, err = validatorManagerSDK.GetValidatorManagerType(rpcURL, common.HexToAddress(validatorManagerAddress))
			if err != nil {
				return models.Sidecar{}, nil, fmt.Errorf("could not obtain validator manager type: %w", err)
			}
			printFunc("  Validation Kind: %s", sc.ValidatorManagement)
			if sc.ValidatorManagement == validatormanagertypes.ProofOfAuthority {
				owner, err := contract.GetContractOwner(rpcURL, common.HexToAddress(validatorManagerAddress))
				if err != nil {
					return models.Sidecar{}, nil, err
				}
				sc.ValidatorManagerOwner = owner.String()
				printFunc("  Validator Manager Owner: %s", sc.ValidatorManagerOwner)
			}
		}
````
and
````
if rpcIsAvailable {
			blockchainID, _ = precompiles.WarpPrecompileGetBlockchainID(rpcURL)
		}
		if blockchainID == ids.Empty {
			blockchainID, err = app.Prompt.CaptureID("What is the Blockchain ID?")
			if err != nil {
				return models.Sidecar{}, nil, err
			}
		}

`````

Comment (sukantoraymond) [reply cmd/blockchaincmd/import_public.go:74]: why don't we just remove this flag and: 
return error if rpc is unavailable when: 
```
sc.ValidatorManagement, err = validatorManagerSDK.GetValidatorManagerType(rpcURL, common.HexToAddress(validatorManagerAddress)
```
since we need to know if poa/ pos anyways

and if
```
blockchainID, _ = precompiles.WarpPrecompileGetBlockchainID(rpcURL)
		}
```
returns error due to url being unavailable we can just prompt for the blockchainID

Comment (sukantoraymond) [reply cmd/blockchaincmd/import_public.go:74]: TLDR: I think we should remove this flag

File: cmd/blockchaincmd/import_public.go [hunk 5]
```diff
@@ -69,6 +70,8 @@ flag.`,
 		"",
 		"the blockchain ID",
 	)
+	cmd.Flags().StringVar(&rpcURL, "rpc", "", "rpc endpoint for the blockchain")
```
sukantoraymond, line 73: whats the difference between nodeEndpoint and rpcURL? I assume for most users, rpcURL = nodeEndpoint.
You can get the same information from nodeEndpoint as rpcURL (blockchainID, vmType, etc). Maybe except for VM Version? 

IMO we should just keep 1 URL flag whether its nodeEndpoint or RPCURL to prevent user confusion

cmd/blockchaincmd/import_public.go
felipemadero: can you clarify what this new function abstracts

cmd/blockchaincmd/import_public.go
felipemadero: There are cases where a rpc in unavailable because all rpc nodes are down but you anyway want to be
able to partially import the L1, most notably to increase validator balances, or to try out some local tests, or for any
misc debugging reason like be able to easily have access to certain L1 info. 
I indeed used this functionality twice in the past month

felipemadero: eg if all your validator drawn the balance, you need to use increase balance, and a reasonable flow would be to
call first import on the rpc, and then increase the balances

cmd/blockchaincmd/import_public.go
felipemadero: Node endpoint is basically the ip of a public tracking node, rpc endpoint is the url provided by an API node.
The first one is used to directly access Pchain on the L1 validator so as to get the vm version.
The second one is how to access the L1.
Do you propose to remove all vm version stuff altogether? This is outside of this PR but can be done anyway.
which should be your proposal to have a vm version set in sidecar, as a replacement of current code?

cmd/blockchaincmd/import_public.go
sukantoraymond: instead of using flag why not we just make a call to this RPC, and if this RPC is not returning any response then we know that it is not available (don't return the error) and respond accordingly as if we have the flag set as true

cmd/blockchaincmd/import_public.go
sukantoraymond: I think we can just merge rpcURL and node Endpoint flags together. I think we can just keep rpc url and remove node endpoint. Basically we automatically detect if the url provided is RPC or end point. If it is node endpoint then we get the vm info from that url

sukantoraymond:
TLDR: I think we want to avoid adding more flags into our commands. If possible we should replace / reuse our existing flags.

cmd/blockchaincmd/import_public.go
felipemadero: we may need both usually, so most probably we remove node endpoint

felipemadero: if you agree we can just remove the node endpoint stuff @sukantoraymond

cmd/blockchaincmd/import_public.go
felipemadero: it can be the case that there is a temporal network failure and the import from a failed rpc will be assumed to be successful, being not the case, also, the idea of this flag is to even avoid asking for the url instead of requiring the user to provide
a fake url.
not opposing to this but the flag, being kind of a bit more ugly, seems a better option to me. 
as you wish

felipemadero: or we can just accept any URL string at prompt, and tell the user which to use for empty rpc, eg just enter

cmd/blockchaincmd/import_public.go
sukantoraymond: ya we can remove node endpoint

sukantoraymond (approved):

PR #2748, opened 2025-04-16
[DPTN-1569] Fixing validate earliest ts no upcoming UT

Tonix517:
## Why this should be merged
Time waiting was called at wrong step of the UT, and it may fail the condition at the line below when block timestamp is at the same second of `time.Now()`:
https://github.com/ava-labs/avalanche-cli/blob/main/cmd/blockchaincmd/upgradecmd/apply.go#L492

## How this was tested
- [X] `./scripts/unit_test.sh`


artemanosov (approved):

felipemadero (approved):

felipemadero:
Nice! Thanks!

PR #2754, opened 2025-04-17
[DPTN-1307] Adding e2e and ut combined coverage checking scripts

Tonix517:
## Why this should be merged
Now e2e and unit test coverage data are merged and the combined coverage percentage is being checked in a script.
**NEXT** to add in GitHub action to invoke it.

## How this works
borrowing part from https://github.com/ava-labs/avalanche-cli/blob/main/scripts/check_coverage_unit_test.sh

## How this was tested
- [X] local run
```
COVERAGE_MODE=true ./scripts/build.sh 
COVERAGE_MODE=true ./scripts/run.e2e.sh
COVERAGE_MODE=true ./scripts/run.e2e.sh --filter "can stop and restart a deployed subnet non SOV"
./scripts/unit_test.sh
./scripts/check_total_coverage.sh
```

Screenshot 2025-04-17 at 3 25 57 PM
Screenshot 2025-04-17 at 3 25 40 PM


.github/workflows/unit-test.yml
```diff
@@ -32,7 +32,3 @@ jobs:
           scripts/unit_test.sh
         env:
           CGO_CFLAGS: "-O -D__BLST_PORTABLE__" # Set the CGO flags to use the portable version of BLST
-
-      - name: Check Code Coverage
-        run: |
-          scripts/check_coverage_unit_test.sh
```
Tonix517, line 38: in next PR i will add it back with the new script

.github/workflows/unit-test.yml
```diff
@@ -26,13 +26,10 @@ jobs:
       - name: Install Go dependencies
         run: go mod download
       
-      - name: Build and Test
+      - name: Build and Testga.
```
felipemadero, line 29: Testga?

.github/workflows/unit-test.yml
```diff
@@ -26,13 +26,10 @@ jobs:
...
+        working-directory: ${{ github.workspace }}
```
felipemadero, line 30: why do we need this now?

scripts/check_total_coverage.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
felipemadero, line 1: lets have a script called coverage.sh that is able to return unit, e2e or combined

scripts/check_total_coverage.sh
```diff
@@ -0,0 +1,23 @@
...
+
+coverage_dir=$(PWD)/coverage
```
felipemadero, line 3: please check previous to this, that the current dir is avalanche-cli repo dir, so PWD uses the correct assumption.

scripts/check_total_coverage.sh
```diff
@@ -0,0 +1,23 @@
...
+combined_coverage_file=$coverage_dir/combined.txt
+COVERAGE_THRESHOLD=15.0 # percentage threshold of code coverage required
```
felipemadero, line 5: lets not do any threshold check

scripts/check_total_coverage.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo "Generating coverage report in text format..."
+included_packages=$(go list ./... | grep -v /tests/ | grep -v '/sdk/') # not including 'tests' and 'sdk'
+go tool covdata merge -i=$coverage_dir/e2e,$coverage_dir/ut -o=$coverage_dir -pkg=${included_packages//$'\n'/,}
```
felipemadero, line 9: do we need th pkg param? why is that?

scripts/check_total_coverage.sh
```diff
@@ -0,0 +1,23 @@
...
+go tool covdata textfmt -i=./coverage -o $combined_coverage_file
+go tool cover -func $combined_coverage_file
+
+echo "Checking total coverage..."
+
+# print current test coverage from report
+current_test_coverage=$(go tool cover -func=$combined_coverage_file  | grep -e "total" | awk '{print $3}')
```
felipemadero, line 16: lets call go tool cover once (previous exec to go into a temp file)

scripts/check_total_coverage.sh
```diff
@@ -0,0 +1,23 @@
...
+echo "Current test coverage is $current_test_coverage"
+
+go tool cover -func=$combined_coverage_file  | grep -e "total" | \
```
felipemadero, line 19: lets avoid this check

scripts/run.e2e.sh
```diff
@@ -60,7 +60,7 @@ ACK_GINKGO_RC=true ~/go/bin/ginkgo build $extra_build_args ./tests/e2e
 
 if [ "${COVERAGE_MODE:-}" == true ]
 then
-    export GOCOVERDIR=coverage/e2e
+    export GOCOVERDIR=$(PWD)/coverage/e2e
```
felipemadero, line 63: please check CWD is avalanche-cli repo (parent of the parent of the script)

scripts/check_total_coverage.sh
felipemadero: (parent of the parent of the script)

scripts/unit_test.sh
```diff
@@ -1,4 +1,12 @@
 #!/usr/bin/env bash
 
-go test -v -coverprofile=coverage.out $(go list ./... | grep -v /tests/ | grep -v '/sdk/')
-#go tool cover -func=coverage.out
+set -ex
+
+coverage_dir=$PWD/coverage/ut # should be under the same parent folder as e2e coverage dir
```
felipemadero, line 5: please check CWD is avalanche-cli repo (parent of the parent of the script)

scripts/unit_test.sh
```diff
@@ -1,4 +1,12 @@
...
+
+echo "Re-creating unit test coverage directory: $coverage_dir"
+rm -rf $coverage_dir
+mkdir -p $coverage_dir
+
+go test -cover -v  $(go list ./... | grep -v /tests/ | grep -v '/sdk/') -args -test.gocoverdir=$coverage_dir
```
felipemadero, line 11: I don't believe we need to specify dir. you can just specify -coverprofile=$coveragedir/coverage.out

felipemadero (changes requested):

felipemadero (commented):
If possible, please add a way to check coverage only for given files or directories.
This is to check partial improvements over a given feature.

scripts/check_total_coverage.sh
felipemadero, line 9: please put combined into coverage/merged

felipemadero (changes requested):

scripts/check_total_coverage.sh
Tonix517: this param tells golang coverage tool about which pkg to analyze coverage. without this param the coverage gave an empty report

Tonix517: addressed

.github/workflows/unit-test.yml
Tonix517: i added `mkdir` call to create `coverage/ut` directory in the ut script. without this `mkdir` failed with permission denied. i will try `CWD` in a bit and see how it goes

scripts/check_total_coverage.sh
Tonix517: can do - if not urgent i'd leave this to my next PR

scripts/check_total_coverage.sh
Tonix517: as my above comment - ut script gave me 'permission denifed' failure without adding this `PWD`. still don't know why.

scripts/check_total_coverage.sh
Tonix517: i added this check to make sure that none of our future PR will decrease the coverage, to guarantee that coverage % will only go up.

.github/workflows/unit-test.yml
Tonix517: fixed

scripts/check_total_coverage.sh
Tonix517: addressed

scripts/check_total_coverage.sh
Tonix517: adding this check here to make sure our coverage in future won't decrease somehow. so every future PR in this repo should at least maintain the same coverage level

scripts/check_total_coverage.sh
felipemadero: can you rename this to coverage.sh ?

felipemadero: or check_coverage.sh

scripts/check_total_coverage.sh
felipemadero: Please check something like

```
if ! [[ "$0" =~ scripts/check_coverage.sh ]]; then
  echo "must be run from repository root"
  exit 1
fi
```

before any other command

scripts/check_total_coverage.sh
felipemadero: ok please remove the success line, it is confusing to have success with just 15%
also put the percent into a variable

scripts/check_total_coverage.sh
felipemadero: I believe I used merged without the pkg param and it just worked

scripts/run.e2e.sh
felipemadero: Please check something like

```
if ! [[ "$0" =~ scripts/run.e2e.sh ]]; then
  echo "must be run from repository root"
  exit 1
fi
```

before any other command

scripts/unit_test.sh
felipemadero: Please check something like

```
if ! [[ "$0" =~ scripts/unit_test.sh ]]; then
  echo "must be run from repository root"
  exit 1
fi
```

before any other command

tests/e2e/utils/helpers.go
```diff
@@ -79,6 +79,14 @@ func GetAPMDir() string {
 	return path.Join(usr.HomeDir, constants.APMDir)
 }
 
+func GetSnapshotsDir() string {
```
sukantoraymond, line 82: Seems that this is not part of this PR?

sukantoraymond:
How is e2e coverage calculation done? Unit test can be done with go test, but e2e?

scripts/check_total_coverage.sh
Tonix517: done - `coverage.sh`

scripts/check_total_coverage.sh
Tonix517: added in

scripts/check_total_coverage.sh
Tonix517: i just removed that percentage check for now.

scripts/run.e2e.sh
Tonix517: addressed

tests/e2e/utils/helpers.go
Tonix517: yea i accidentally checked in the commits i cherry-picked. reverted.

Tonix517:
ginkgo can be treated as a wrapper to `go test`. it exposes arguments that we use for coverage as `go test`

> How is e2e coverage calculation done? Unit test can be done with go test, but e2e?

felipemadero (approved):
LGTM

PR #2755, opened 2025-04-17
Deploy flags e2e

sukantoraymond:
Our current E2E for tests/e2e/testcases is pretty extensive for local network across all commands. However, it does not test flags input extensively for happy path and error path. 

For example, our E2E tests for blockchain deploy in https://github.com/ava-labs/avalanche-cli/blob/main/tests/e2e/testcases/subnet/sov/ currently tests that we are able to deploy to local network using PoA / PoS / Initialize Validator Manager. However, it does not test the multiple ways that we can achieve each objectives (for example we can use deploy an L1 with by setting num-local-machine=2 instead of using default value of 1). 

This PR proposes using json files as input that is fed into a general function (TestCommandWithJSONConfig), which can be reused for all blockchain subcommands (e.g. blockchain addValidator, etc) simply by feeding it with different JSON file. Reusing same core testing logic enables us to achieve extensive happy path and error path coverage without huge additional testing code.


tests/e2e/commands_e2e/blockchain/deploy/deploy_tests.json
```diff
@@ -0,0 +1,23 @@
...
+    "local": true,
+    "skip-icm-deploy": true,
+    "skip-update-check": true,
+    "blockchainName": "testSubnet"
+  },
+  "happyPath": [
+    {
+      "name": "local_deploy",
+      "flags": {}
+    }
```
sukantoraymond, line 12: we can set globalFlags in each testCases as well, which will override the globalFlags value

tests/e2e/commands_e2e/blockchain/deploy/suite.go
```diff
@@ -0,0 +1,68 @@
...
+	config *blockchain.TestJSONConfig
+	err    error
+)
+
+const ewoqEVMAddress = "0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"
+
+var _ = ginkgo.Describe("[Blockchain Deploy Flags]", ginkgo.Ordered, func() {
+	_ = ginkgo.BeforeEach(func() {
+		// Create test subnet config
+		commands.CreateSubnetEvmConfigSOV(subnetName, utils.SubnetEvmGenesisPath, ewoqEVMAddress)
```
sukantoraymond, line 30: We use ewoq here because we use ewoq as validator manager owner when we deploy in local network

tests/e2e/commands/blockchain/blockchain_command.go
```diff
@@ -0,0 +1,111 @@
...
+
+var avalancheBinaryPath = "./bin/avalanche"
+
+// SetAvalancheBinaryPath sets the path to the avalanche binary
+func SetAvalancheBinaryPath(path string) {
+	avalancheBinaryPath = path
+}
+
+// TestCommandWithJSONConfig tests a CLI command with flag inputs from a JSON file
+func TestCommandWithJSONConfig(command string, configPath string, testCase *TestCase) (string, error) {
```
Tonix517, line 35: love the idea - any plan to move it to some test common directories to enable it reusable by all other commands?

sukantoraymond: Thanks! Yes i will move it to external directories

sukantoraymond: Made it general

tests/e2e/commandse2e/blockchain/deploy/suite.go
```diff
@@ -0,0 +1,73 @@
...
+		commands.DeleteSubnetConfig(subnetName)
+	})
+	blockchainCmdArgs := []string{subnetName}
+	ginkgo.It("should successfully deploy a blockchain", func() {
+		// Run each happy path test case
+		for _, testCase := range config.HappyPath {
+			ginkgo.By(fmt.Sprintf("Running test case: %s", testCase.Name))
+			_, err = commandse2e.TestCommandWithJSONConfig(
+				commandse2e.BlockchainCmd,
+				"deploy",
```
felipemadero, line 50: this should be a constant on cmd/blockchaincmd/

tests/e2e/commandse2e/blockchain/deploy/suite.go
```diff
@@ -0,0 +1,73 @@
...
+		config, err = commandse2e.ReadTestConfig(deployTestJSONPath)
+		gomega.Expect(err).Should(gomega.BeNil())
+	})
+
+	ginkgo.AfterEach(func() {
+		commands.CleanNetwork()
+		// Cleanup test subnet config
+		commands.DeleteSubnetConfig(subnetName)
+	})
+	blockchainCmdArgs := []string{subnetName}
```
felipemadero, line 43: this should belong to the json test case IMO

tests/e2e/commandse2e/blockchain/deploy/suite.go
```diff
@@ -0,0 +1,73 @@
...
+	config *commandse2e.TestJSONConfig
+	err    error
+)
+
+const ewoqEVMAddress = "0x8db97C7cEcE249c2b98bDC0226Cc4C2A57BF52FC"
+
+var _ = ginkgo.Describe("[Blockchain Deploy Flags]", ginkgo.Ordered, func() {
+	_ = ginkgo.BeforeEach(func() {
+		// Create test subnet config
+		commands.CreateSubnetEvmConfigSOV(subnetName, ewoqEVMAddress, commands.PoA)
```
felipemadero, line 31: maybe this should be added to the json spec (how to setup required env)

PR #2757, opened 2025-04-21
[DPTN-1598] Adding localnetwork stop network e2e

Tonix517:
## Why this should be merged
https://ava-labs.atlassian.net/browse/DPTN-1598

## How this was tested
- [X] `./scripts/run.e2e.sh --filter "can stop a started network"`

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -248,6 +248,9 @@ func setupEnv() (string, error) {
 		return "", err
 	}
 	baseDir := filepath.Join(usr.HomeDir, constants.BaseDirName)
+	if utils.IsE2E() {
+		baseDir = filepath.Join(usr.HomeDir, constants.E2EBaseDirName)
+	}
 
 	// Create base dir if it doesn't exist
 	if err := os.MkdirAll(baseDir, os.ModePerm); err != nil {
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -17,8 +17,9 @@ const (
 
 	UbuntuVersionLTS = "24.04"
 
-	BaseDirName = ".avalanche-cli"
-	LogDir      = "logs"
+	BaseDirName    = ".avalanche-cli"
+	E2EBaseDirName = ".avalanche-cli/e2e"
+	LogDir         = "logs"
 
 	LocalNetworkMetaFile = "localNetworks.json"
 
--- a/tests/e2e/commands/network.go
+++ b/tests/e2e/commands/network.go
@@ -81,12 +81,15 @@ func StartNetworkWithVersion(version string) string {
 }
 
 /* #nosec G204 */
-func StopNetwork() {
-	cmd := exec.Command(
-		CLIBinary,
+func StopNetwork(stopCmdFlags ...string) {
+	stopCmdFlasg := append([]string{
 		NetworkCmd,
 		"stop",
-		"--"+constants.SkipUpdateFlag,
+		"--" + constants.SkipUpdateFlag,
+	}, stopCmdFlags...)
+	cmd := exec.Command(
+		CLIBinary,
+		stopCmdFlasg...,
 	)
 	output, err := cmd.CombinedOutput()
 	if err != nil {
@@ -95,4 +98,20 @@ func StopNetwork() {
 		utils.PrintStdErr(err)
 	}
 	gomega.Expect(err).Should(gomega.BeNil())
+}
+
+func GetNetworkStatus() (string, error) {
+	cmd := exec.Command(
+		CLIBinary,
+		NetworkCmd,
+		"status",
+		"--"+constants.SkipUpdateFlag,
+	)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		fmt.Println(cmd.String())
+		fmt.Println(string(output))
+		utils.PrintStdErr(err)
+	}
+	return string(output), err
 }
--- a/tests/e2e/commands/network.go
+++ b/tests/e2e/commands/network.go
@@ -81,7 +81,7 @@ func StartNetworkWithVersion(version string) string {
 }
 
 /* #nosec G204 */
-func StopNetwork(stopCmdFlags ...string) {
+func StopNetwork(stopCmdFlags ...string) error {
 	stopCmdFlasg := append([]string{
 		NetworkCmd,
 		"stop",
@@ -97,7 +97,7 @@ func StopNetwork(stopCmdFlags ...string) {
 		fmt.Println(string(output))
 		utils.PrintStdErr(err)
 	}
-	gomega.Expect(err).Should(gomega.BeNil())
+	return err
 }
 
 func GetNetworkStatus() (string, error) {
--- a/tests/e2e/utils/constants.go
+++ b/tests/e2e/utils/constants.go
@@ -3,8 +3,12 @@
 
 package utils
 
+import (
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+)
+
 const (
-	baseDir      = ".avalanche-cli"
+	baseDir      = constants.E2EBaseDirName
 	hardhatDir   = "./tests/e2e/hardhat"
 	confFilePath = hardhatDir + "/dynamic_conf.json"
 	greeterFile  = hardhatDir + "/greeter.json"
--- a/tests/e2e/utils/helpers.go
+++ b/tests/e2e/utils/helpers.go
@@ -79,6 +79,14 @@ func GetAPMDir() string {
 	return path.Join(usr.HomeDir, constants.APMDir)
 }
 
+func GetSnapshotsDir() string {
+	return filepath.Join(GetBaseDir(), constants.SnapshotsDirName)
+}
+
+func GetSnapshotPath(snapshotName string) string {
+	return filepath.Join(GetSnapshotsDir(), snapshotName)
+}
+
 func ChainConfigExists(subnetName string) (bool, error) {
 	cfgPath := filepath.Join(GetBaseDir(), constants.SubnetDir, subnetName, constants.ChainConfigFileName)
 	cfgExists := true
@@ -307,6 +315,13 @@ func DeleteAPMBin(vmid string) {
 
 	// ignore error, file may not exist
 	_ = os.RemoveAll(vmPath)
+}
+
+func DeleteSnapshot(snapshotName string) {
+	snapshotPath := path.Join(GetSnapshotsDir(), snapshotName)
+
+	// ignore error, file may not exist
+	_ = os.RemoveAll(snapshotPath)
 }
 
 func stdoutParser(output string, queue string, capture string) (string, error) {
@@ -614,6 +629,12 @@ func CheckSubnetEVMExists(version string) bool {
 func CheckAvalancheGoExists(version string) bool {
 	avagoPath := path.Join(GetBaseDir(), constants.AvalancheCliBinDir, constants.AvalancheGoInstallDir, "avalanchego-"+version)
 	_, err := os.Stat(avagoPath)
+	return err == nil
+}
+
+func CheckSnapshotExists(snapshotName string) bool {
+	snapshotPath := filepath.Join(GetSnapshotsDir(), snapshotName)
+	_, err := os.Stat(snapshotPath)
 	return err == nil
 }
 
--- a/tests/e2e/commands/cleanup.go
+++ b/tests/e2e/commands/cleanup.go
@@ -19,7 +19,7 @@ func DeleteE2EInventory() {
 	usr, err := user.Current()
 	gomega.Expect(err).Should(gomega.BeNil())
 	homeDir := usr.HomeDir
-	inventoryE2E := filepath.Join(homeDir, constants.BaseDirName, "nodes/inventories/", constants.E2EClusterName)
+	inventoryE2E := filepath.Join(homeDir, constants.E2EBaseDirName, "nodes/inventories/", constants.E2EClusterName)
 	fmt.Println("deleting: ", inventoryE2E)
 	err = os.RemoveAll(inventoryE2E)
 	gomega.Expect(err).Should(gomega.BeNil())
@@ -32,7 +32,7 @@ func DeleteNode(nodeID string) {
 	usr, err := user.Current()
 	gomega.Expect(err).Should(gomega.BeNil())
 	homeDir := usr.HomeDir
-	nodeE2E := filepath.Join(homeDir, constants.BaseDirName, "nodes", nodeID)
+	nodeE2E := filepath.Join(homeDir, constants.E2EBaseDirName, "nodes", nodeID)
 	fmt.Println("deleting: ", nodeE2E)
 	err = os.RemoveAll(nodeE2E)
 	gomega.Expect(err).Should(gomega.BeNil())
@@ -43,7 +43,7 @@ func DeleteE2ECluster() {
 	gomega.Expect(err).Should(gomega.BeNil())
 	homeDir := usr.HomeDir
 	relativePath := "nodes"
-	content, err := os.ReadFile(filepath.Join(homeDir, constants.BaseDirName, relativePath, constants.ClustersConfigFileName))
+	content, err := os.ReadFile(filepath.Join(homeDir, constants.E2EBaseDirName, relativePath, constants.ClustersConfigFileName))
 	if err == os.ErrNotExist {
 		return
 	}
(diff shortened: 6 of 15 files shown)


artemanosov (approved):
LGTM

tests/e2e/testcases/localnetwork/suite.go
```diff
@@ -0,0 +1,27 @@
...
+		out := commands.StartNetwork()
+		gomega.Expect(out).Should(gomega.ContainSubstring("Network ready to use"))
+
+		commands.StopNetwork()
+
+		out, err := commands.GetNetworkStatus()
+		gomega.Expect(err).ShouldNot(gomega.BeNil())
+		gomega.Expect(out).Should(gomega.ContainSubstring("network is not running"))
+	})
+})
```
sukantoraymond, line 27: I know this is currently still a draft, but few thoughts: 
- we can add local network start test (and we can test the multiple flags options here)
- we can add local network status test
- use stop local network test (including the two flags that are there (snapshot, etc))

Tonix517: sure can do

tests/e2e/commands/network.go
```diff
@@ -81,11 +81,30 @@ func StartNetworkWithVersion(version string) string {
 }
 
 /* #nosec G204 */
-func StopNetwork() {
+func StopNetwork(stopCmdFlags ...string) {
+	stopCmdFlasg := append([]string{
+		NetworkCmd,
+		"stop",
```
felipemadero, line 87: I wonder if we should start using a constants for the command names, defined under cmd/ tree

tests/e2e/commands/network.go
```diff
@@ -81,11 +81,30 @@ func StartNetworkWithVersion(version string) string {
 }
 
 /* #nosec G204 */
-func StopNetwork() {
+func StopNetwork(stopCmdFlags ...string) {
```
felipemadero, line 84: A good way to avoid having many functions for the different commands, is to have just one function
ExecuteCLICommand or something like that, which receives the cmd name and the flags
(it may only add the skip update flag)

tests/e2e/commands/network.go
```diff
@@ -81,11 +81,30 @@ func StartNetworkWithVersion(version string) string {
...
+		CLIBinary,
+		stopCmdFlasg...,
+	)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		fmt.Println(cmd.String())
+		fmt.Println(string(output))
+		utils.PrintStdErr(err)
+	}
+	gomega.Expect(err).Should(gomega.BeNil())
```
felipemadero, line 100: really this does not give possibility to check stop error conditions

tests/e2e/testcases/localnetwork/suite.go
```diff
@@ -0,0 +1,75 @@
...
+		out, err := commands.GetNetworkStatus()
+		gomega.Expect(err).ShouldNot(gomega.BeNil())
+		gomega.Expect(out).Should(gomega.ContainSubstring("network is not running"))
+	})
+
+	ginkgo.It("can start then stop a started network with --dont-save", func() {
+		out := commands.StartNetwork()
+		gomega.Expect(out).Should(gomega.ContainSubstring("Network ready to use"))
+
+		commands.StopNetwork("--dont-save")
```
felipemadero, line 56: this should be a constant on cmd/flags/ used also in cmd/ code

tests/e2e/testcases/localnetwork/suite.go
```diff
@@ -0,0 +1,75 @@
...
+		gomega.Expect(out).Should(gomega.ContainSubstring("Network ready to use"))
+
+		commands.StopNetwork()
+
+		out, err := commands.GetNetworkStatus()
+		gomega.Expect(err).ShouldNot(gomega.BeNil())
+		gomega.Expect(out).Should(gomega.ContainSubstring("network is not running"))
+	})
+
+	ginkgo.It("can start then stop a started network with --dont-save", func() {
```
felipemadero, line 52: this should use the same constant

tests/e2e/commands/network.go
felipemadero: anyway I recognize the e2e code will be easier to read if we have one function per command

tests/e2e/testcases/localnetwork/suite.go
felipemadero: --dont-save does something on the state, can this be checked? (really it does not create a snapshot)
so the cases for snapshots are:
- stop without args: create a default snapshot
- stop with --dont-save: dont create a default snapshot
- stop with --snapshot-name: create a snapshot with a different name
network clean beforehand will remove the default snapshot, the other snapshot need to be removed by hand 
(this is order to check it is created)

tests/e2e/testcases/localnetwork/suite.go
```diff
@@ -0,0 +1,75 @@
...
+		out := commands.StartNetwork()
+		gomega.Expect(out).Should(gomega.ContainSubstring("Network ready to use"))
+
+		commands.StopNetwork("--snapshot-name", "test-snapshot")
+
+		out, err := commands.GetNetworkStatus()
+		gomega.Expect(err).ShouldNot(gomega.BeNil())
+		gomega.Expect(out).Should(gomega.ContainSubstring("network is not running"))
+
+		// TODO: remove the saved snapshot
```
felipemadero, line 73: pls. but also check the snapshot is where expected

tests/e2e/testcases/localnetwork/suite.go
```diff
@@ -0,0 +1,75 @@
...
+		out, err = commands.GetNetworkStatus()
+		gomega.Expect(err).ShouldNot(gomega.BeNil())
+		gomega.Expect(out).Should(gomega.ContainSubstring("network is not running"))
+	})
+
+	ginkgo.It("can start then stop a started network", func() {
+		out := commands.StartNetwork()
+		gomega.Expect(out).Should(gomega.ContainSubstring("Network ready to use"))
+
+		commands.StopNetwork()
```
felipemadero, line 45: we need to be able to check command err return, and force errors conditions in this testsuite

tests/e2e/commands/network.go
felipemadero: in that later case make all functions call this generic one

tests/e2e/commands/network.go
```diff
@@ -81,11 +81,30 @@ func StartNetworkWithVersion(version string) string {
...
+}
+
+func GetNetworkStatus() (string, error) {
 	cmd := exec.Command(
 		CLIBinary,
 		NetworkCmd,
-		"stop",
+		"status",
```
felipemadero, line 107: should we use a constant defined on cmd/ ?

tests/e2e/testcases/localnetwork/suite.go
```diff
@@ -0,0 +1,75 @@
...
+		out, err := commands.GetNetworkStatus()
+		gomega.Expect(err).ShouldNot(gomega.BeNil())
+		gomega.Expect(out).Should(gomega.ContainSubstring("network is not running"))
+	})
+
+	ginkgo.It("can start then stop a started network with --snapshot-name", func() {
+		out := commands.StartNetwork()
+		gomega.Expect(out).Should(gomega.ContainSubstring("Network ready to use"))
+
+		commands.StopNetwork("--snapshot-name", "test-snapshot")
```
felipemadero, line 67: this needs to be a flag name constant in cmd/flags/

felipemadero: better if the test name is also a constant inside this file

felipemadero (commented):
last question: which is the improved coverage? only over cmd/networkcmd/stop.go code

tests/e2e/commands/network.go
Tonix517: same as above - it is a good idea and i'd do it in a follow up PR

tests/e2e/testcases/localnetwork/suite.go
```diff
@@ -0,0 +1,75 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
```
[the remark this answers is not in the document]
Tonix517: a small CON is that this import list will be very large: https://github.com/ava-labs/avalanche-cli/blob/main/tests/e2e/e2e_test.go#L12-L32

tests/e2e/testcases/localnetwork/suite.go
Tonix517: in e2e we are using the same `~/.avalanche-cli/snapshots` directory as user's daily CLI use. so e2e here may interfere with user's daily CLI uses, particularly the `default` snapshot. (btw what i need to do in this e2e here is to check existence and then removal) @felipemadero

Tonix517:
> last question: which is the improved coverage? only over cmd/networkcmd/stop.go code

will need to close https://github.com/ava-labs/avalanche-cli/pull/2754 first to get the report from GitHub actions. i'm looking at some github action marketplace to show code coverage info to PR: https://josh-ops.com/posts/github-code-coverage/#google_vignette

tests/e2e/testcases/localnetwork/suite.go
felipemadero: mmm your are totally right. options:
1) I am not opposite to temporal renaming of the default snapshot and then renaming it back again. testing is mainly
a dev team operation and we are expected to recover the default in case the "renaming it back again" part fails.
2) I am not opposite to have the snapshot dir being redefinable at the app level, so we on testing use a non production one.
as you wish (probably 2 is best no?)

Tonix517: Yes #2 sounds better indeed. btw do we have our CLI options categorized as public and internal using? to add such a new option, do we want to make it public to users too?

felipemadero:
> > last question: which is the improved coverage? only over cmd/networkcmd/stop.go code
> 
> will need to close #2754 first to get the report from GitHub actions. i'm looking at some github action marketplace to show code coverage info to PR: https://josh-ops.com/posts/github-code-coverage/#google_vignette

I am mainly interested in the coverage improvement mostly related to this PR:
- cmd/network/
- pkg/localnet/
So I believe a manual execution and analysis is needed to answer this.

tests/e2e/testcases/localnetwork/suite.go
felipemadero: Our public options are set at config file level, and there is not structure regarding public vs
internal options. Either is it public in the config file, or is internal in various places, and usually fixed.
But we don't have any public option to move something out of `~/.avalanche-cli`. 
I believe that atm, for less impact (eg in support), we can just keep it private, to continue 
having that centralization.

felipemadero: Btw, interested in ideas related to public vs internal options. Feel free to propose ideas on prototype PRs

Tonix517:
gave it a run: 
- `pkg/localnetwork` didn't have coverage change:
before_localnet

- `cmd/networkcmd/` does have increased coverage:

before
before_networkcmd

after
after_networkcmd



> > > last question: which is the improved coverage? only over cmd/networkcmd/stop.go code
> > 
> > 
> > will need to close #2754 first to get the report from GitHub actions. i'm looking at some github action marketplace to show code coverage info to PR: https://josh-ops.com/posts/github-code-coverage/#google_vignette
> 
> I am mainly interested in the coverage improvement mostly related to this PR:
> 
> * cmd/network/
> * pkg/localnet/
>   So I believe a manual execution and analysis is needed to answer this.

tests/e2e/testcases/localnetwork/suite.go
Tonix517: addressed.

tests/e2e/commands/network.go
Tonix517: addressed - now `err` is returned

tests/e2e/commands/network.go
Tonix517: we are calling `exec.Command` - is that what you are mentioning here?

tests/e2e/testcases/localnetwork/suite.go
Tonix517: @felipemadero please check this: https://github.com/ava-labs/avalanche-cli/pull/2761

Tonix517: @sukantoraymond made a valid point here: https://github.com/ava-labs/avalanche-cli/pull/2761#issuecomment-2825545133 given that there's already `IsE2E()` uses in code, that's the easiest change to use a separate e2e directory for testing purposes, which is also the least risky. 
i think we can do this `IsE2E` solution for now and leave public\internal CLI arguments for more discussions

@felipemadero

tests/e2e/commands/network.go
felipemadero: I mean, all the body inside here can be abstracted away in a separate function, with arguments:
- NetworkCmd
- "stop"
- stopCmdFlags

tests/e2e/testcases/localnetwork/suite.go
felipemadero: I don't agree with having separate special cmd flows or settings for e2e inside the code. It is happening rn I now, but that seems tech debt to me. Maybe we can use it as a solution, but have a deeper discussion on this.
Probably we want to be able to configure the CLI dir by flag, and use that everywhere in all E2E commands. 
But not using a hidden flag but a public one. Anyway, feel free to use E2E env var and continue thinking if
we want to do better.

cmd/root.go
```diff
@@ -258,6 +258,9 @@ func setupEnv() (string, error) {
 
 	// Create snapshots dir if it doesn't exist
 	snapshotsDir := filepath.Join(baseDir, constants.SnapshotsDirName)
+	if utils.IsE2E() {
```
felipemadero, line 261: probably we just want to change baseDir if IsE2E

pkg/constants/constants.go
```diff
@@ -103,7 +103,8 @@ const (
 
 	// it's unlikely anyone would want to name a snapshot `default`
 	// but let's add some more entropy
-	SnapshotsDirName = "snapshots"
+	SnapshotsDirName      = "snapshots"
+	E2ESnapshotSubDirName = "e2e" // under SnapshotsDirName
```
felipemadero, line 107: can we use a new base dir for E2E?

tests/e2e/utils/helpers.go
```diff
@@ -79,6 +79,14 @@ func GetAPMDir() string {
 	return path.Join(usr.HomeDir, constants.APMDir)
 }
 
+func GetSnapshotsDir() string {
```
felipemadero, line 82: can we use app obtained from GetApp() for this paths?

felipemadero (commented):
Almost ready

felipemadero:
> gave it a run:
> 
> * `pkg/localnetwork` didn't have coverage change:
> 
> before_localnet
> * `cmd/networkcmd/` does have increased coverage:
> 
> before before_networkcmd
> 
> after after_networkcmd
> 
> > > > last question: which is the improved coverage? only over cmd/networkcmd/stop.go code
> > > 
> > > 
> > > will need to close #2754 first to get the report from GitHub actions. i'm looking at some github action marketplace to show code coverage info to PR: https://josh-ops.com/posts/github-code-coverage/#google_vignette
> > 
> > 
> > I am mainly interested in the coverage improvement mostly related to this PR:
> > 
> > * cmd/network/
> > * pkg/localnet/
> >   So I believe a manual execution and analysis is needed to answer this.

thanks for this

felipemadero:
> https://github.com/ava-labs/storm/blob/main/.github/workflows/ci-tests.yaml#L80

ok pretty curious how this will work

pkg/constants/constants.go
Tonix517: addressed.

cmd/root.go
Tonix517: addressed. it is a better idea indeed, it allows us to separate other e2e related testing in future.

tests/e2e/utils/helpers.go
Tonix517: updated `baseDir` const in `tests/e2e/utils/constants.go` and we don't need this change anymore. all existing logic simply work

tests/e2e/commands/network.go
Tonix517: let me see how other NetworkCmd e2e tests look in my next PRs. will do this should I see this pattern in more tests

tests/e2e/utils/constants.go
```diff
@@ -4,7 +4,7 @@
 package utils
 
 const (
-	baseDir      = ".avalanche-cli"
+	baseDir      = ".avalanche-cli/e2e"
```
felipemadero, line 7: I would prefer to use E2EBaseDirName from the other package

felipemadero (approved):
LGTM

PR #2758, opened 2025-04-21
Artem/create e2e tests for transfer command

artemanosov:


tests/e2e/testcases/key/suite.go
```diff
@@ -283,61 +284,506 @@ var _ = ginkgo.Describe("[Key]", func() {
...
+
+				output, err = commands.ListKeys("local", true, true, "")
+				gomega.Expect(err).Should(gomega.BeNil())
+				_, keyBalance3, err := utils.ParseAddrBalanceFromKeyListOutput(output, keyName, "P-Chain")
+				gomega.Expect(err).Should(gomega.BeNil())
+				_, ewoqKeyBalance3, err := utils.ParseAddrBalanceFromKeyListOutput(output, ewoqKeyName, "P-Chain")
+				gomega.Expect(err).Should(gomega.BeNil())
+				gomega.Expect(ewoqKeyBalance1 - ewoqKeyBalance3).
+					Should(gomega.Equal(feeNAvax + amountNAvax))
+				gomega.Expect(keyBalance3 - keyBalance1).Should(gomega.Equal(amountNAvax))
```
sukantoraymond, line 353: this is a good test

cmd/keycmd/list.go
```diff
@@ -646,16 +646,21 @@ func formatCChainBalance(balance *big.Int) (string, error) {
...
 	}
-	// convert to nAvax
-	balance = balance.Div(balance, big.NewInt(int64(units.Avax)))
-	if balance.Cmp(big.NewInt(0)) == 0 {
+
+	// convert to nAvax with rounding
+	divisor := big.NewInt(int64(units.Avax))
+	half := big.NewInt(int64(units.Avax) - 1)
+	adjusted := new(big.Int).Add(balance, half)
+	result := new(big.Int).Div(adjusted, divisor)
```
sukantoraymond, line 654: not sure why we can't just leave balance = balance.Div(balance, big.NewInt(int64(units.Avax)))

cmd/keycmd/transfer.go
```diff
@@ -732,6 +763,13 @@ func exportFromP(
 		}
 		return err
 	}
+	pContext := wallet.P().Builder().Context()
+	pFeeCalculator := avagofee.NewDynamicCalculator(pContext.ComplexityWeights, pContext.GasPrice)
+	txFee, err := pFeeCalculator.CalculateFee(unsignedTx)
+	if err != nil {
+		return err
+	}
+	ux.Logger.PrintToUser("P-Chain Paid fee: %.9f", float64(txFee)/float64(units.Avax))
```
sukantoraymond, line 772: its great that we are printing which kind of fee is being paid (p chain / c chain, etc)

tests/e2e/testcases/key/suite.go
```diff
@@ -283,61 +284,506 @@ var _ = ginkgo.Describe("[Key]", func() {
...
+
+				output, err = commands.ListKeys("local", false, true, "")
+				gomega.Expect(err).Should(gomega.BeNil())
+				_, keyBalance3, err := utils.ParseAddrBalanceFromKeyListOutput(output, ewoqKeyName, "P-Chain")
+				gomega.Expect(err).Should(gomega.BeNil())
+				_, ewoqKeyBalance3, err := utils.ParseAddrBalanceFromKeyListOutput(output, ewoqKeyName, "C-Chain")
+				gomega.Expect(err).Should(gomega.BeNil())
+				gomega.Expect(ewoqKeyBalance1 - ewoqKeyBalance3).
+					Should(gomega.Equal(cChainFee + amountNAvax))
+				gomega.Expect(keyBalance3 - keyBalance1).Should(gomega.Equal(amountNAvax - pChainFee))
```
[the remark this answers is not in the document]
artemanosov: Yeah, I think I could refactor this, but left for it for later. Thanks for the comment though!

cmd/keycmd/transfer.go
sukantoraymond, line 772: would recommend to put AVAX at the end

sukantoraymond: its to differentiate when users pay with their blockchain gas token. Maybe AVAX (P-Chain)?

cmd/keycmd/list.go
artemanosov: Basic division here loses the precision. The problem is that division of big.Int does not round the numbers, which caused failure of assertions when the fee was calculated in tests. Using the logic that I added we make sure to round the number if needed which results in more accurate balance output. I can provide examples if needed

tests/e2e/testcases/key/suite.go
```diff
@@ -283,61 +284,506 @@ var _ = ginkgo.Describe("[Key]", func() {
...
+					keyName := "nokey"
+					cmd := commands.KeyTransferSend("ewoq", "", "", keyName, "0.1", "--c-chain-sender", "", "--c-chain-receiver", "", "", "")
+					output, err := cmd.CombinedOutput()
+
+					gomega.Expect(err).Should(gomega.HaveOccurred())
+					gomega.Expect(string(output)).
+						Should(gomega.ContainSubstring(fmt.Sprintf(".avalanche-cli/key/%s.pk: no such file or directory", keyName)))
+				})
+
+				ginkgo.It("should fail when amount provided amount is negative", func() {
```
sukantoraymond, line 612: lets add another one when amount is 0

tests/e2e/testcases/key/suite.go
```diff
@@ -283,61 +284,506 @@ var _ = ginkgo.Describe("[Key]", func() {
...
+					cmd := commands.KeyTransferSend("ewoq", "", "", "ewoq", "-0.1", "--c-chain-sender", "", "--c-chain-receiver", "", "", "")
+					output, err := cmd.CombinedOutput()
+
+					gomega.Expect(err).Should(gomega.HaveOccurred())
+					gomega.Expect(string(output)).
+						Should(gomega.ContainSubstring("amount must be positive"))
+				})
+
+				ginkgo.It("should fail to load sidecard when blockchain does not exist in subnets directory", func() {
```
sukantoraymond, line 621: sidecar

cmd/keycmd/transfer.go
artemanosov: We can add that👍

tests/e2e/commands/key.go
```diff
@@ -119,26 +122,62 @@ func ExportKeyToFile(keyName string, outputPath string) (string, error) {
 }
 
 /* #nosec G204 */
-func KeyTransferSend(keyName string, targetAddr string, amount string) (string, error) {
+func KeyTransferSend(keyName, ledger, destinationAddr, destinationKey, amount, senderChain, senderBlockchainName, receiverChain, receiverBlockchainName, originTransferrerAddress, destinationTransferrerAddress string) *exec.Cmd {
```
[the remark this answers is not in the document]
felipemadero: why are you returning the cmd? this sets a diff convention

felipemadero, line 125: maybe better to just receive all cmdline args here?

felipemadero: or a map flag name -> value

cmd/keycmd/transfer.go
```diff
@@ -975,6 +1019,8 @@ func exportFromC(
 		}
 		return err
 	}
+	ux.Logger.PrintToUser("C-Chain Paid fee: %.9f", float64(unsignedTx.Ins[0].Amount-amount)/float64(units.Avax))
```
felipemadero, line 1022: probably check Ins len before doing this

tests/e2e/testcases/key/suite.go
```diff
@@ -24,9 +24,10 @@ const (
 	testKeyWith0x   = "tests/e2e/assets/test_key_0x.pk"
```
felipemadero, line 24: I should move this file into something like tests/e2e/testcases/commands/avalanche_key_transfer.go
or some other convention

tests/e2e/testcases/key/suite.go
```diff
@@ -283,61 +284,506 @@ var _ = ginkgo.Describe("[Key]", func() {
...
+					"",
+					"",
+				)
+
+				output, err := cmd.CombinedOutput()
+
+				gomega.Expect(err).Should(gomega.HaveOccurred())
+				gomega.Expect(string(output)).
+					Should(gomega.ContainSubstring("transfer from P-Chain to Test-Chain is not supported"))
+			})
```
felipemadero, line 785: C -> X ?

felipemadero (commented):
very very nice and comprehensive.  final question: which is the coverage
improvement? just over cmd/keycmd/transfer.go code

tests/e2e/commands/interchain.go
```diff
@@ -0,0 +1,36 @@
...
+	"fmt"
+	"os/exec"
+
+	"github.com/ava-labs/avalanche-cli/pkg/constants"
+	"github.com/ava-labs/avalanche-cli/tests/e2e/utils"
+	"github.com/onsi/gomega"
+)
+
+const (
+	InterchainCMD = "interchain"
```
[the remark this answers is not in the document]
artemanosov: Yeah, I think we could refactor it in future PR. If we agree on Raymond's PR for one test command instead of multiple utils functions, one file with all commands as constants would work well

cmd/keycmd/list.go
```diff
@@ -646,16 +646,21 @@ func formatCChainBalance(balance *big.Int) (string, error) {
...
+	if result.Cmp(big.NewInt(0)) == 0 {
 		return "0", nil
 	}
 	balanceStr := ""
 	if useNanoAvax {
-		balanceStr = fmt.Sprintf("%9d", balance.Uint64())
+		balanceStr = fmt.Sprintf("%9d", result.Uint64())
 	} else {
-		balanceStr = fmt.Sprintf("%.9f", float64(balance.Uint64())/float64(units.Avax))
+		balanceStr = fmt.Sprintf("%.9f", float64(result.Uint64())/float64(units.Avax))
```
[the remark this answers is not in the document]
artemanosov: I would need to convert it back to big int. I think this is cleaner

tests/e2e/commands/key.go
```diff
@@ -57,14 +57,17 @@ func CreateKeyForce(keyName string) (string, error) {
...
+func ListKeys(network string, omitCChain bool, useNanoAvax bool, subnets string) (string, error) {
 	args := []string{KeyCmd, "list", "--" + network, "--" + constants.SkipUpdateFlag}
 	if omitCChain {
 		args = append(args, "--cchain=false")
 	}
 	if useNanoAvax {
 		args = append(args, "--use-nano-avax=true")
 	}
+	if subnets != "" {
+		args = append(args, "--subnets", "c,x,p,"+subnets)
```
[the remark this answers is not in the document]
artemanosov: don't see those flags there. As @Tonix517 mentioned it would be nice to have all commands and flags as constants so we could reference them either in utils or unified test function (from Raymond's PR)

tests/e2e/commands/key.go
```diff
@@ -57,14 +57,17 @@ func CreateKeyForce(keyName string) (string, error) {
 }
 
 /* #nosec G204 */
-func ListKeys(network string, omitCChain bool, useNanoAvax bool) (string, error) {
+func ListKeys(network string, omitCChain bool, useNanoAvax bool, subnets string) (string, error) {
 	args := []string{KeyCmd, "list", "--" + network, "--" + constants.SkipUpdateFlag}
```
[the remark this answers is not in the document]
artemanosov: not sure which one you're referring to here

cmd/keycmd/transfer.go
```diff
@@ -572,6 +586,29 @@ func interEvmSend(
...
+		return err
+	}
+
+	chainName, err := contract.GetBlockchainDesc(senderChain)
+	if err != nil {
+		return err
+	}
+	ux.Logger.PrintToUser("%s Paid fee: %.9f",
+		chainName,
+		calculateEvmFee(receipt.GasUsed, receipt.EffectiveGasPrice))
```
Tonix517, line 599: worth its own func? seeing the same logic is being called below

Tonix517: i meant the `GetBlockchainDesc->PrintToUsers->calculateEvmFee` call sequence. no strong opinions here actually, just out of sake of aesthetics

cmd/keycmd/transfer.go
```diff
@@ -423,6 +443,8 @@ func intraEvmSend(
 		if err != nil {
 			return err
 		}
+	} else if amountFlt < 0 {
```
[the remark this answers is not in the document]
artemanosov: what for? we handle case when amount is 0 (prompt), if amount is <0, we throw an error, for the reset (amount is positive) we continue

tests/e2e/commands/key.go
```diff
@@ -57,14 +57,17 @@ func CreateKeyForce(keyName string) (string, error) {
 }
 
 /* #nosec G204 */
-func ListKeys(network string, omitCChain bool, useNanoAvax bool) (string, error) {
+func ListKeys(network string, omitCChain bool, useNanoAvax bool, subnets string) (string, error) {
```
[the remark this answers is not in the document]
artemanosov: that's true, but I thought that 4 args is not that much comparing to transfer for example which takes a lot of input arguments.

tests/e2e/commands/key.go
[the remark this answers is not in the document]
felipemadero: I am trying us to avoid directly using strings like "list" "blockchain" for commands, but instead to define constants for them, for example on cmd/constants/ and use those both in cmd specs and in tests. Not sure if this should be solved in this PR, but we may want to have a standard way of doing this.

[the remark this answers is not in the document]
artemanosov: Yea, I get the point, I think it's fair. We just don't have all the consts and I feel like it is out of scope for this PR, since it's focused on tests of key transfer

tests/e2e/commands/interchain.go
```diff
@@ -0,0 +1,36 @@
+package commands
```
felipemadero, line 1: you need to add license here to avoid CI linting to fail

tests/e2e/commands/contract.go
```diff
@@ -0,0 +1,48 @@
+package commands
```
felipemadero, line 1: you need to add license here to avoid CI linting to fail

felipemadero (commented):
You are having a lot of issue on CI due to yarn install failure. Please fix this so all is green https://github.com/ava-labs/avalanche-cli/actions/runs/14672388268/job/41181712445?pr=2758

cmd/keycmd/transfer.go
artemanosov: There are one more occurrence of similar block, I think it's not worth to extract

PR #2759, opened 2025-04-22
Added bls-public-key and  bls-proof-of-possession cli options to blockchain changeWeight and removeValidator

digithrone:
avalanche blockchain <name> changeWeight and removeValidator lacked option to add bls public key and proof of possessio via cli args.

## Why this should be merged
It allows for remote changeWeight and validator removal without the need to open port 9650 and configured rpc access as validators are isolated and normally do not allow rpc access

## How this was tested
I have performed a rebalancing of testnet environment for the client with single 100 weight node and rebalanced to 5 validator nodes all  of weight 100 by using addValidator, changeWeight and removeValidator. 

## How is this documented
The fixes are trivial so no additional documentation is needed.

### DIFF
--- a/cmd/blockchaincmd/change_weight.go
+++ b/cmd/blockchaincmd/change_weight.go
@@ -64,6 +64,8 @@ The L1 has to be a Proof of Authority L1.`,
 	cmd.Flags().Uint64Var(&newWeight, "weight", 0, "set the new staking weight of the validator")
 	cmd.Flags().BoolVarP(&useEwoq, "ewoq", "e", false, "use ewoq key [fuji/devnet only]")
 	cmd.Flags().StringVar(&nodeIDStr, "node-id", "", "node-id of the validator")
+	cmd.Flags().StringVar(&publicKey, "bls-public-key", "", "set the BLS public key of the validator to add")
+	cmd.Flags().StringVar(&pop, "bls-proof-of-possession", "", "set the BLS proof of possession of the validator to add")
 	cmd.Flags().StringVar(&nodeEndpoint, "node-endpoint", "", "gather node id/bls from publicly available avalanchego apis on the given endpoint")
 	cmd.Flags().BoolVarP(&useLedger, "ledger", "g", false, "use ledger instead of key (always true on mainnet, defaults to false on fuji/devnet)")
 	cmd.Flags().StringSliceVar(&ledgerAddresses, "ledger-addrs", []string{}, "use the given ledger addresses")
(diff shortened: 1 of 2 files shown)


cmd/blockchaincmd/change_weight.go
```diff
@@ -64,6 +64,8 @@ The L1 has to be a Proof of Authority L1.`,
 	cmd.Flags().Uint64Var(&newWeight, "weight", 0, "set the new staking weight of the validator")
 	cmd.Flags().BoolVarP(&useEwoq, "ewoq", "e", false, "use ewoq key [fuji/devnet only]")
 	cmd.Flags().StringVar(&nodeIDStr, "node-id", "", "node-id of the validator")
+	cmd.Flags().StringVar(&publicKey, "bls-public-key", "", "set the BLS public key of the validator to add")
+	cmd.Flags().StringVar(&pop, "bls-proof-of-possession", "", "set the BLS proof of possession of the validator to add")
```
felipemadero, line 68: pls change the description to something related to the change weight command, eg "BLS proof of possession of the validator that is to change weight"

cmd/blockchaincmd/remove_validator.go
```diff
@@ -68,6 +68,8 @@ these prompts by providing the values with flags.`,
 	cmd.Flags().BoolVarP(&useLedger, "ledger", "g", false, "use ledger instead of key (always true on mainnet, defaults to false on fuji)")
 	cmd.Flags().StringSliceVar(&ledgerAddresses, "ledger-addrs", []string{}, "use the given ledger addresses")
 	cmd.Flags().StringVar(&nodeIDStr, "node-id", "", "node-id of the validator")
+	cmd.Flags().StringVar(&publicKey, "bls-public-key", "", "set the BLS public key of the validator to add")
```
felipemadero, line 71: this command does not use this variables at all

felipemadero (commented):
currently it is possible to avoid the node to have open http api, CLI asks for pop on
change weight, and remove validator asks for node id or receive it from flags (remove validator does not use bls info). 
what is ok is that change weight was not having a flag for pop, which this PR adds

digithrone:
I removed unnecessary change in removeValidator and also removed public key options so only pop is added to changeWeight command. Also updated PR title and description to properly reflect code changes.

cmd/blockchaincmd/change_weight.go
felipemadero: pls change the description to something related to the change weight command, eg "BLS proof of possession of the validator that is to change weight"

felipemadero:
> I removed unnecessary change in removeValidator and also removed public key options so only pop is added to changeWeight command. Also updated PR title and description to properly reflect code changes.

Please also consider https://github.com/ava-labs/avalanche-cli/pull/2759#discussion_r2054540299

digithrone:
I changed the description so it is compatible with node-id description.

PR #2763, opened 2025-04-24
fix empty node id in validator list

felipemadero:
## Why this should be merged
Fixes cases of empty node ID in `avalanche validator list` when the node uses all the balance.
Also makes the following command to not need to use the validator manager:
- avalanche validator list
- avalanche validator getBalance
- avalanche validator increaseBalance


Tonix517 (approved):
i assume confusions like this can be reduced when we get to the code reorg around `pkg` and `sdk` dirs.

felipemadero:
> i assume confusions like this can be reduced when we get to the code reorg around `pkg` and `sdk` dirs.

Indeed. Problem here is that I don't want sdk to import pkg anymore, no I am doing to partial moves.
There is a ticket to fully move pkg/utils into sdk which will make thing things harder to happen.

PR #2764, opened 2025-04-24
Make upgradecmd to work

felipemadero:
## Why this should be merged
`avalanche blockchain upgrade generate` always exit with error trying to connect to local network. Also, a given upgrade is never applied to local clusters or local networks.
this PR fixes issues related to L1 configurations:
- all blockchain configurations are now applied on deploy (upgrade was missing, global node settings was missing)
- all blockchain configuration updates are now applied on restart (all was missing)
- global node settings are applied on local network restart
- upgrade generate is fixed so it does not try to check for balances on local network


sukantoraymond:
Since I don't have a lot of context in this, I think having an E2E for upgrade that includes this change (we already have `testcases/upgrade/non-sov/`, but im pretty sure those e2e don't cover this change) will go a long way towards understanding how `upgrade` is meant to be used.

sukantoraymond (approved):

PR #2765, opened 2025-04-24
Updating all copyright year to 2025

Tonix517:


cmd/blockchaincmd/add_validator.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2022, Ava Labs, Inc. All rights reserved.
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
```
Tonix517, line 1: given that we already have `LICENSE` file at the repo root, we actually do not need to have this header in every source file.

sukantoraymond: we actuallly have a check somehwere in ci that checks that we have copyright, but i think we can keep it

artemanosov (approved):

PR #2771, opened 2025-04-28
Remove duplicate ICM flags in deploy

sukantoraymond:
Remove duplicate flags to minimize flag options to users


Tonix517 (approved):
q: for breaking changes like this, do we need to communicate with customers somehow?

sukantoraymond:
> q: for breaking changes like this, do we need to communicate with customers somehow?

Ideally we would have deprecation notice. The duplication here happened when teleporter and relayer were renamed to ICM and thats why we added the flags, but i think enough time has passed to just remove references to teleporter and relayer

cmd/blockchaincmd/deploy.go
```diff
@@ -175,13 +175,8 @@ so you can take your locally tested Blockchain and deploy it on Fuji or Mainnet.
 
 	icmGroup := flags.RegisterFlagGroup(cmd, "ICM Flags", "show-icm-flags", false, func(set *pflag.FlagSet) {
 		set.BoolVar(&icmSpec.SkipICMDeploy, "skip-icm-deploy", false, "Skip automatic ICM deploy")
-		set.BoolVar(&icmSpec.SkipICMDeploy, "skip-local-teleporter", false, "skip automatic ICM deploy on local networks [to be deprecated]")
-		set.BoolVar(&icmSpec.SkipICMDeploy, "skip-teleporter-deploy", false, "skip automatic ICM deploy")
-		set.BoolVar(&icmSpec.SkipICMDeploy, "noicm", false, "skip automatic ICM deploy")
```
felipemadero, line 180: isn't --noicm better that any other option? or --icm

sukantoraymond: noicm is less clear than skip-icm-deploy

felipemadero (commented):
can you also fix cmd/commands.md ?

PR #2774, opened 2025-04-29
Remove unused commands

sukantoraymond:
Reduce the amount of code we need to support E2E


cmd/blockchaincmd/validators.go
```diff
@@ -1,124 +0,0 @@
...
-	"github.com/ava-labs/avalanche-cli/pkg/networkoptions"
-	"github.com/ava-labs/avalanche-cli/pkg/subnet"
-	"github.com/ava-labs/avalanchego/ids"
-	"github.com/ava-labs/avalanchego/vms/platformvm"
-	"github.com/olekukonko/tablewriter"
-	"github.com/spf13/cobra"
-)
-
-// avalanche blockchain validators
-func newValidatorsCmd() *cobra.Command {
```
sukantoraymond, line 23: we already have avalanche validators list

Tonix517:
Do we collect user metrics, like which cmd user used, and errors occurred etc.? if so that metrics data will be useful for us to make decisions like this. 

btw there may be privacy considerations if we want to do this.

Tonix517:
https://us.posthog.com/project/23813

> Do we collect user metrics, like which cmd user used, and errors occurred etc.? if so that metrics data will be useful for us to make decisions like this.
> 
> btw there may be privacy considerations if we want to do this.

Tonix517 (approved):

felipemadero (changes requested):
I don't believe we remove all the commands. Publish for sure. But no validators and stats. Those should be integrated somewhere else.

felipemadero:
> I don't believe we remove all the commands. Publish for sure. But no validators and stats. Those should be integrated somewhere else.

Most probably we want to merge that code into validator command group, but not fully remove it

cmd/blockchaincmd/validators.go
felipemadero: we need to integrate this into validators list, then remove. this is for another PR

PR #2789, opened 2025-05-01
Remove use local nodes

sukantoraymond:
Remove num-local-nodes, since we can just use existing flag num-bootstrap-validators


felipemadero (approved):

Tonix517 (approved):
LGTM. do we have Release Notes showing potential breaking changes like this to customers?

sukantoraymond:
> LGTM. do we have Release Notes showing potential breaking changes like this to customers?

Yes, we will update our commands.md as well as relay this in release notes

PR #2784, opened 2025-05-01
Adding localnetwork e2e tests: start\status\clean

Tonix517:
## Why this should be merged
https://ava-labs.atlassian.net/browse/DPTN-1625
https://ava-labs.atlassian.net/browse/DPTN-1626
https://ava-labs.atlassian.net/browse/DPTN-1627

## How this was tested
- [X] ran new e2e test cases locally

PR #2792, opened 2025-05-05
[DPTN-1653] Adding e2e test for local node start

Tonix517:
## Why this should be merged
https://ava-labs.atlassian.net/browse/DPTN-1653

## How this was tested
- [X] locally ran e2e script with filter

NOTE: will need to rebase when https://github.com/ava-labs/avalanche-cli/pull/2784 is merged.

### DIFF
--- a/tests/e2e/commands/node.go
+++ b/tests/e2e/commands/node.go
@@ -15,6 +15,7 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/ava-labs/avalanche-cli/pkg/constants"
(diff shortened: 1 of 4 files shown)


tests/e2e/commands/node.go
```diff
@@ -207,6 +208,25 @@ func NodeImport(filename string, clusterName string) string {
 	return runCmd(cmd, ExpectSuccess)
 }
 
+func NodeLocalStart(clusterName string) (string, error) {
```
felipemadero, line 211: start has a good amount of flags that can be tested

felipemadero (commented):
please take into account testing the flags

PR #2795, opened 2025-05-06
add check for zero balance on add validator

felipemadero:
## Why this should be merged
Avoids posterior failure or infinite interation


Tonix517 (approved):

sukantoraymond (approved):
i would say selected key has zero balance

PR #2797, opened 2025-05-09
chore: make function comment match function name

yetyear:
## Why this should be merged

make function comment match function name


sukantoraymond (approved):

yetyear:
Thanks for your review and approved.

PR #2798, opened 2025-05-09
Add e2e tests for icm deploy and sendMsg commands

artemanosov:


tests/e2e/testcases/icm/sendMsg/suite.go
```diff
@@ -0,0 +1,522 @@
...
+			sendMsgFlags := utils.TestFlags{
+				"key": ewoqKeyName,
+			}
+
+			hexEncodedMessage := "0x48656c6c6f20576f726c64" // "Hello World" in hex
+			sendMessageArgs := []string{
+				cChain,
+				subnetName,
+				hexEncodedMessage,
+				"--hex-encoded",
```
sukantoraymond, line 371: @felipemadero any reason why we would ever need hex-encoded flag?

tests/e2e/testcases/icm/sendMsg/suite.go
```diff
@@ -0,0 +1,522 @@
...
+		commands.CleanNetwork()
+		err := utils.DeleteConfigs(subnetName)
+		gomega.Expect(err).Should(gomega.BeNil())
+		utils.DeleteCustomBinary(subnetName)
+
+		err = utils.DeleteConfigs(subnet2Name)
+		gomega.Expect(err).Should(gomega.BeNil())
+		utils.DeleteCustomBinary(subnet2Name)
+	})
+
```
sukantoraymond, line 35: seems that all these tests are for non sov blockchain -> (subnets, not L1s), which is a good start. 

Can we have tests for L1s too? 

This means: 
- CreateSubnetEvmConfigSOV
 - DeploySubnetLocallySOV -> this one should have icm already deployed automatically unless u disable it with a flag

sukantoraymond:
Right now we have three commands that do the same thing: 
- interchain messenger
- icm
- teleporter

Can we just keep one and remove two?

@felipemadero

tests/e2e/testcases/icm/sendMsg/suite.go
```diff
@@ -0,0 +1,522 @@
...
+				cChain,
+				subnetName,
+				message,
+			}
+
+			output, err := utils.TestCommand(utils.ICMCmd, "sendMsg", sendMessageArgs, globalFlags, sendMsgFlags)
+			gomega.Expect(err).Should(gomega.BeNil())
+			gomega.Expect(output).Should(gomega.ContainSubstring(fmt.Sprintf("Delivering message \"%s\" from source blockchain \"%s\"", message, cChain)))
+			gomega.Expect(output).Should(gomega.ContainSubstring(fmt.Sprintf("Waiting for message to be delivered to destination blockchain \"%s\"", subnetName)))
+			gomega.Expect(output).Should(gomega.ContainSubstring("Message successfully Teleported!"))
```
sukantoraymond, line 83: I think these three lines can be abstracted into a function

sukantoraymond, line 83: also i think we need to change messaing of "Message successfully Teleported!" -> sth like Message successfully delivered

@felipemadero

tests/e2e/testcases/icm/sendMsg/suite.go
```diff
@@ -0,0 +1,522 @@
...
+
+		err = utils.DeleteConfigs(subnet2Name)
+		gomega.Expect(err).Should(gomega.BeNil())
+		utils.DeleteCustomBinary(subnet2Name)
+	})
+
+	ginkgo.Context("with valid input", func() {
+		ginkgo.It("should send a message from c-chain to subnet", func() {
+			commands.CreateSubnetEvmConfigNonSOV(subnetName, utils.SubnetEvmGenesisPath, false)
+			commands.DeploySubnetLocallyNonSOV(subnetName)
```
sukantoraymond, line 39: why not place these two in beforeEach

tests/e2e/testcases/icm/sendMsg/suite.go
felipemadero: to provide any byte data as a test message, not just strings. needed for more complex interactions with
contracts, eg to simulate contract calls

tests/e2e/testcases/icm/sendMsg/suite.go
felipemadero: I ignore if we want to use this specialized functions, or more generic ones with appropriate args. I tend to prefer the last option as it is in the way of less funtions

felipemadero:
> Right now we have three commands that do the same thing:
> 
> * interchain messenger
> * icm
> * teleporter
> 
> Can we just keep one and remove two?
> 
> @felipemadero

i believe we can now remove teleporter. probably keep both interchain messenger and icm as an alias.
if the desire is to keep one, just icm ofc. I would only remove teleporter

tests/e2e/testcases/icm/sendMsg/suite.go
felipemadero: can you include this in the PR @artemanosov ?

tests/e2e/testcases/icm/sendMsg/suite.go
```diff
@@ -0,0 +1,522 @@
...
+			output, err := utils.TestCommand(utils.ICMCmd, "sendMsg", sendMessageArgs, globalFlags, sendMsgFlags)
+			gomega.Expect(err).Should(gomega.BeNil())
+			gomega.Expect(output).Should(gomega.ContainSubstring(fmt.Sprintf("Delivering message \"%s\" from source blockchain \"%s\"", message, subnet2Name)))
+			gomega.Expect(output).Should(gomega.ContainSubstring(fmt.Sprintf("Waiting for message to be delivered to destination blockchain \"%s\"", subnetName)))
+			gomega.Expect(output).Should(gomega.ContainSubstring("Message successfully Teleported!"))
+
+			commands.StopRelayer()
+		})
+
+		ginkgo.It("should send a message from subnet to subnet", func() {
```
sukantoraymond, line 200: whats the diff between this and the one before

tests/e2e/commands/relayer.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/ava-labs/avalanche-cli/tests/e2e/utils"
+	"github.com/onsi/gomega"
+)
+
+/* #nosec G204 */
+func StopRelayer() {
+	cmdArgs := []string{InterchainCMD, "relayer", "stop", "--local"}
+	cmdArgs = append(cmdArgs, "--"+constants.SkipUpdateFlag)
+
+	cmd := exec.Command(CLIBinary, cmdArgs...)
```
felipemadero, line 20: why you dont use TestCommand here?

tests/e2e/testcases/icm/sendMsg/suite.go
```diff
@@ -0,0 +1,522 @@
...
+
+			// Send a message
+			globalFlags := utils.GlobalFlags{
+				"local":             true,
+				"skip-update-check": true,
+			}
+
+			sendMsgFlags := utils.TestFlags{
+				"key":        ewoqKeyName,
+				"source-rpc": rpcs2[0],
```
sukantoraymond, line 311: If i use source rpc and dest rpc, i technically don't even need subnet names in the command argument anymore? @felipemadero

tests/e2e/testcases/icm/deploy/suite.go
```diff
@@ -0,0 +1,441 @@
...
+			commands.CleanNetwork()
+			err := utils.DeleteConfigs(subnetName)
+			gomega.Expect(err).Should(gomega.BeNil())
+			utils.DeleteCustomBinary(subnetName)
+		})
+		ginkgo.It("should deploy ICM contracts into c-chain", func() {
+			globalFlags := utils.GlobalFlags{
+				"local":             true,
+				"skip-update-check": true,
+			}
```
sukantoraymond, line 37: globalFlags can be defined on top, so it doesn't have to be defined in every testcases

tests/e2e/testcases/icm/deploy/suite.go
```diff
@@ -0,0 +1,441 @@
...
+	ginkgo.Context("with valid input", func() {
+		ginkgo.BeforeEach(func() {
+			commands.StartNetwork()
+		})
+
+		ginkgo.AfterEach(func() {
+			commands.CleanNetwork()
+			err := utils.DeleteConfigs(subnetName)
+			gomega.Expect(err).Should(gomega.BeNil())
+			utils.DeleteCustomBinary(subnetName)
```
felipemadero, line 31: not sure on the need for this

tests/e2e/testcases/icm/sendMsg/suite.go
```diff
@@ -0,0 +1,522 @@
...
+			gomega.Expect(err).Should(gomega.BeNil())
+			gomega.Expect(output).Should(gomega.ContainSubstring(fmt.Sprintf("Delivering message \"%s\" from source blockchain \"%s\"", message, subnet2Name)))
+			gomega.Expect(output).Should(gomega.ContainSubstring(fmt.Sprintf("Waiting for message to be delivered to destination blockchain \"%s\"", subnetName)))
+			gomega.Expect(output).Should(gomega.ContainSubstring("Message successfully Teleported!"))
+
+			commands.StopRelayer()
+		})
+
+		ginkgo.It("should send a message from subnet to subnet", func() {
+			commands.CreateSubnetEvmConfigNonSOV(subnetName, utils.SubnetEvmGenesisPath, false)
```
felipemadero, line 201: is this repeated?

tests/e2e/testcases/icm/deploy/suite.go
```diff
@@ -0,0 +1,441 @@
...
+		ginkgo.It("should deploy ICM contracts into c-chain", func() {
+			globalFlags := utils.GlobalFlags{
+				"local":             true,
+				"skip-update-check": true,
+			}
+			testFlags := utils.TestFlags{
+				"key": ewoqKeyName,
+			}
+			commandArguments := []string{
+				"--c-chain",
```
sukantoraymond, line 42: shouldn't this be a flag

sukantoraymond: anyways seems interesting that using c chain flagand not using it in the test case after this both result in "ICM Messenger successfully deployed to C-Chain".

Should we make it so that it will deploy to C-Chain always even without flag -> this means we remove the flag @felipemadero

tests/e2e/testcases/icm/deploy/suite.go
```diff
@@ -0,0 +1,441 @@
...
+				"key":              ewoqKeyName,
+			}
+			commandArguments := []string{
+				"--c-chain",
+			}
+
+			_, err := utils.TestCommand(utils.ICMCmd, "deploy", commandArguments, globalFlags, testFlags)
+			gomega.Expect(err).Should(gomega.BeNil())
+
+			output, err := utils.TestCommand(utils.ICMCmd, "deploy", append(commandArguments, "--force-registry-deploy"), globalFlags, utils.TestFlags{
```
sukantoraymond, line 292: @felipemadero do we need this flag --force-registry-deploy? Why dont we always deploy registry even if messenger is already deployed by default?

tests/e2e/testcases/icm/deploy/suite.go
```diff
@@ -0,0 +1,441 @@
...
+			testFlags := utils.TestFlags{
+				"key": ewoqKeyName,
+			}
+			commandArguments := []string{
+				"--c-chain",
+			}
+
+			output, err := utils.TestCommand(utils.ICMCmd, "deploy", commandArguments, globalFlags, testFlags)
+			gomega.Expect(err).Should(gomega.BeNil())
+			gomega.Expect(output).
```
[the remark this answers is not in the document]
sukantoraymond: I think we should find out if there are contract in that location -> also good knowledge for us to have anyways

tests/e2e/testcases/icm/sendMsg/suite.go
felipemadero: the command can be improved in this sense, but currently the blockchain names are required.
the origin of this flags is to support situations where the blockchain conf does not contain the rpc.
I believe we can plan an improved ux for the command, but probably is not part of the current effort.
we can make a ticket

tests/e2e/testcases/icm/deploy/suite.go
felipemadero: --c-chain specifies deploy on c-chain only. it is the specification of the l1. it can be either c-chain or some blockchain name so the flag is needed.
but, for local network, c-chain is automatically deployed even if the blockchain name is another

felipemadero: this not happens on fuji/mainnet, there you only deploy where indicated

tests/e2e/testcases/icm/deploy/suite.go
felipemadero: by default it tries to avoid deploying a registry if there is a registry already deployed. the command can be improved
I believe we can make a ticket for this. ideally we should be able to find out if a registry is present, we don't 
have yet a way to do this, we can only check if a messenger is deployed.

felipemadero: that is, by default it will assume a registry was deployed if a messenger is present. but there are cases where
the messenger is there but the registry not, so you need to force the deploy. open to suggestions in appropriate ticket based on this comments

tests/e2e/testcases/icm/sendMsg/suite.go
artemanosov: Do we want to do tests for SOV blockchains now or do we want to just add tests for non SOV blockchains to cover more commands?

tests/e2e/testcases/icm/sendMsg/suite.go
artemanosov: Not a fan of exporting assertions. I think some redundancy in tests is okay, as long as it makes them more readable

tests/e2e/testcases/icm/sendMsg/suite.go
artemanosov: good call

tests/e2e/testcases/icm/sendMsg/suite.go
artemanosov: Mistake. Will remove

tests/e2e/testcases/icm/deploy/suite.go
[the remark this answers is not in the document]
artemanosov: How would we do that?

tests/e2e/testcases/icm/sendMsg/suite.go
sukantoraymond: what exactly are hex encoded flags? I will update the flag description so that its more easily understood by users @felipemadero

tests/e2e/testcases/icm/deploy/suite.go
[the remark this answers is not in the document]
sukantoraymond: @felipemadero are there functions we can implement to check if a contract exists in a blockchain?

tests/e2e/testcases/icm/sendMsg/suite.go
artemanosov: I believe the user provides the string in a hex format and it is sent that way, so it's more flexible. Not sure if it's actually used much

tests/e2e/testcases/icm/deploy/suite.go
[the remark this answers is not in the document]
felipemadero: you can use ContractAlreadyDeployed from sdk/evm/evm.go

tests/e2e/testcases/icm/sendMsg/suite.go
felipemadero: We want to switch to using SOV , non SOV is planned for deprecation

PR #2799, opened 2025-05-09
remove all previous blockchain conf on create --force

felipemadero:
## Why this should be merged
Currently, when using --force to overwrite a blockchain conf, it may happen that secondary blockchain
conf is reused, such as subnet json, blockchain json, etc.

## How this works
This PR ensures a blockchain is removed when using create with --force arg


artemanosov (approved):

sukantoraymond:
lets write e2e for this

felipemadero:
> lets write e2e for this

there is work to be done on e2e for create command, so will be done there

PR #2800, opened 2025-05-09
add spec of staking key/cert/bls for new nodes on local cluster

felipemadero:
## Why this should be merged
Currently, on deploy and convert, no staking key/cert/bls can be specified, always obtaining fresh ids.
But a expected use case is to be able to specify info from a previous validator.
This is also pretty useful for testing, in order to repeat cluster node details.

## How this works
Adds flags to deploy/convert and use the values as appropriate on the lib calls.

## How this was tested
tests on local environment

### DIFF
--- a/cmd/blockchaincmd/deploy.go
+++ b/cmd/blockchaincmd/deploy.go
@@ -71,6 +71,9 @@ var (
 	avagoBinaryPath                 string
 	numBootstrapValidators          int
 	numLocalNodes                   int
+	stakingTLSKeyPaths              []string
+	stakingCertKeyPaths             []string
+	stakingSignerKeyPaths           []string
 	httpPorts                       []uint
 	stakingPorts                    []uint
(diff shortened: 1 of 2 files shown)


artemanosov (approved):

cmd/blockchaincmd/deploy.go
```diff
@@ -71,6 +71,9 @@ var (
 	avagoBinaryPath                 string
 	numBootstrapValidators          int
 	numLocalNodes                   int
+	stakingTLSKeyPaths              []string
+	stakingCertKeyPaths             []string
+	stakingSignerKeyPaths           []string
```
sukantoraymond, line 76: is stakingSignerKey BLS Key?

felipemadero: yes

sukantoraymond (approved):

PR #2818, opened 2025-05-21
remove `avalanche teleporter` cmd

felipemadero:
## Why this should be merged
currently we have `avalanche icm` as a convenient shorthand, and the naming was deprecated


artemanosov (approved):

cmd/root.go
```diff
@@ -95,16 +95,11 @@ in with avalanche blockchain create myNewBlockchain.`,
 	// add node command
 	rootCmd.AddCommand(nodecmd.NewCmd(app))
 
-	// add teleporter command
-	subcmd := messengercmd.NewCmd(app)
-	subcmd.Use = "teleporter"
-	rootCmd.AddCommand(subcmd)
-
 	// add interchain command
 	rootCmd.AddCommand(interchaincmd.NewCmd(app))
```
sukantoraymond, line 99: should remove interchain too, just keep icm

Issue #2629, opened 2025-02-24
ACP-99


sukantoraymond:
Implement ACP-99 based on https://github.com/avalanche-foundation/ACPs/discussions/165#discussioncomment-11512285

Issue #2640, opened 2025-02-26
ACP - 99 impelementation


sukantoraymond:
https://github.com/avalanche-foundation/ACPs/pull/170

PR #2821, opened 2025-05-27
use validator manager v2_0_0 as default

felipemadero:
## Why this should be merged
Closes https://github.com/ava-labs/avalanche-cli/issues/2629 https://github.com/ava-labs/avalanche-cli/issues/2640

Main difference with CLI 1.0.0 implementation:

- On 1.0.0 there is only one validator manager, one proxy, one proxy admin, both for PoS and PoA cases. For PoA, the proxy, proxy admin, and manager are all set in genesis. For PoS, the proxy, proxy admin, validator messages lib and reward calculator are set in genesis, while the manager is deployed afterwards.
- On 2.0.0, there is only one validator manager, one proxy and one proxy admin for PoA, but two proxies, two proxy admins, a validator manager, and a specialized validator manager for PoS. For PoA, the proxy, proxy admin, and manager are all set in genesis. For PoS, the two proxies, the two proxy admins, the validator messages lib and reward calculator are set in genesis, while the manager and the specialized managers are deployed afterwards. So, not always a manager is having a fixed address. But the proxies have.

Validator Manager Proxy Address: 0x0FEEDC0DE0000000000000000000000000000000 (PoA)
Specialized Manager Proxy Address: 0x100C0DE1C0FFEE00000000000000000000000000 (PoS)

For PoS, a Validator Manager Proxy Address is also present at the FEEDCODE address, but is not expected to be
typically accessed by the user.

### DIFF
--- a/cmd/blockchaincmd/convert.go
+++ b/cmd/blockchaincmd/convert.go
@@ -3,7 +3,6 @@
 package blockchaincmd
 
 import (
-	"errors"
 	"fmt"
 	"math/big"
 	"os"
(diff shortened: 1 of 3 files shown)


Tonix517 (approved):
approved to unblock merging - assuming all tests pass

PR #2830, opened 2025-06-04
Adding cron job for code coverage

Tonix517:
## Why this should be merged
shell scripts authored by @felipemadero 

## How this works
the new workflow will run every 24 hrs at 0:0 UTC

## How this was tested
https://github.com/ava-labs/avalanche-cli/actions/runs/15453123406/job/43499719206?pr=2830

### DIFF
--- a/.github/workflows/code-cov-report.yml
+++ b/.github/workflows/code-cov-report.yml
@@ -1,0 +1,49 @@
+name: Code Coverage Cron Report
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *' # Run daily at midnight UTC
+  
+jobs:
+  code_cov_report:
+    name: Code Coverage Report
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: "16.x"
+        
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Install Go dependencies
+        run: go mod download
+      
+      - name: Build
+        working-directory: ${{ github.workspace }}
+        run: |
+          scripts/build.sh
+        env:
+          CGO_CFLAGS: "-O -D__BLST_PORTABLE__" # Set the CGO flags to use the portable version of BLST
+
+      - name: Unit Tests
+        working-directory: ${{ github.workspace }}
+        run: |
+          scripts/unit_test.sh
+
+      - name: E2E Tests
+        working-directory: ${{ github.workspace }}
+        run: |
+          scripts/run_e2e_suites_for_cov.sh
+
+      # at the moment the coverage report is printed to the console
+      - name: Generate Code Coverage Report
+        run: |
+          scripts/report_coverage.sh
--- a/.github/workflows/code-cov-report.yml
+++ b/.github/workflows/code-cov-report.yml
@@ -26,13 +26,6 @@ jobs:
       - name: Install Go dependencies
         run: go mod download
       
-      - name: Build
-        working-directory: ${{ github.workspace }}
-        run: |
-          scripts/build.sh
-        env:
-          CGO_CFLAGS: "-O -D__BLST_PORTABLE__" # Set the CGO flags to use the portable version of BLST
-
       - name: Unit Tests
         working-directory: ${{ github.workspace }}
         run: |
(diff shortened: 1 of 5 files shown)


scripts/run_e2e_suites_for_cov.sh
```diff
@@ -0,0 +1,38 @@
+
+export COVERAGE_MODE=true
+export LEDGER_SIM=true
+
+rm -rf coverage/e2e
```
felipemadero, line 5: the provided script was just an example, this line for example needs PWD, and before that, a check that PWD is correct needs to be done (similar to other scripts)

scripts/run_e2e_suites_for_cov.sh
```diff
@@ -0,0 +1,38 @@
...
+killall avalanchego
```
felipemadero, line 6: this should be avoided, is a cleanup I added for easy repeated tests but may kill undesired processes

scripts/run_e2e_suites_for_cov.sh
```diff
@@ -0,0 +1,38 @@
...
+rm -f /tmp/testKey.pk
```
felipemadero, line 7: add a TODO to this line, because really the associated test should be improved, so it execution can be repeated without issues

scripts/report_coverage.sh
```diff
@@ -0,0 +1,56 @@
+mkdir -p coverage/merged
+go tool covdata merge -i=coverage/e2e,coverage/ut -o coverage/merged
+
+go tool covdata textfmt -i coverage/merged/ -o profile.txt
+
+cat profile.txt\
+	| grep -v github.com/ava-labs/avalanche-cli/internal/mocks > profile.tmp
```
felipemadero, line 7: lets filter out just mocks, not github.com/ava-labs/avalanche-cli/internal/mocks, a new mock was added in a different location

scripts/report_coverage.sh
```diff
@@ -0,0 +1,56 @@
...
+total_functions=$(wc -l coverage.txt | awk '{print $1}')
+covered_functions=$(cat coverage.txt | grep -v '\t0.0%$' | wc -l | awk '{print $1}')
+coverage=$(tail -1 coverage.txt | awk '{print $NF}')
+
+echo
+echo NON CLOUD NON SDK
+echo Total functions: $total_functions
+echo Covered functions: $covered_functions
+echo Coverage: $coverage
+
```
felipemadero, line 56: lets add also an html generation

felipemadero: most useful for local execution atm

scripts/report_coverage.sh
```diff
@@ -0,0 +1,56 @@
...
+	| grep -v github.com/ava-labs/avalanche-cli/pkg/ssh\
+	| grep -v github.com/ava-labs/avalanche-cli/pkg/docker\
+	| grep -v github.com/ava-labs/avalanche-cli/pkg/ansible > profile.tmp
+cat profile.txt | grep github.com/ava-labs/avalanche-cli/cmd/nodecmd/local.go >> profile.tmp
+cat profile.txt | grep github.com/ava-labs/avalanche-cli/pkg/node/local.go >> profile.tmp
+mv profile.tmp profile.txt
+
+go tool cover -func profile.txt > coverage.txt
+total_functions=$(wc -l coverage.txt | awk '{print $1}')
+covered_functions=$(cat coverage.txt | grep -v '\t0.0%$' | wc -l | awk '{print $1}')
```
felipemadero, line 34: this seems broken, why covered functions == total functions? it was not the case on local macos

.github/workflows/code-cov-report.yml
```diff
@@ -0,0 +1,49 @@
...
+        
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+
+      - name: Install Go dependencies
+        run: go mod download
+      
+      - name: Build
```
felipemadero, line 29: probably this step is not needed, e2e just build the binary I believe so (also the binary needs to be coverage ready)

scripts/report_coverage.sh
felipemadero: if continue being the same, probably we want to remove this info

scripts/report_coverage.sh
```diff
@@ -0,0 +1,56 @@
+mkdir -p coverage/merged
```
felipemadero, line 1: need to check where it is executing (PWD)
also, probably add PWD to each accessed file and properly locate them

.github/workflows/code-cov-report.yml
```diff
@@ -0,0 +1,49 @@
...
+        working-directory: ${{ github.workspace }}
+        run: |
+          scripts/unit_test.sh
+
+      - name: E2E Tests
+        working-directory: ${{ github.workspace }}
+        run: |
+          scripts/run_e2e_suites_for_cov.sh
+
+      # at the moment the coverage report is printed to the console
```
felipemadero, line 46: which are other options? may a ticket be added for the relevant ones?

Tonix517:
> there is a lot of e2e failure. why is that?

there was some failures that we recently fixed. now i see all e2e test cases are passing

PR #2832, opened 2025-06-05
Check that CLI is at least specified min version

sukantoraymond:
When CLI has higher min version specified
```
AVL-7H2W7V:avalanche-cli raymondsukanto$ ./bin/avalanche blockchain
Error: CLI version is required to be at least v1.8.11, current CLI version is v1.8.10, please upgrade CLI by calling `curl -sSfL https://raw.githubusercontent.com/ava-labs/avalanche-cli/main/scripts/install.sh | sh -s`

```

This is required because if we update our schema for [dependency versioning](https://raw.githubusercontent.com/ava-labs/avalanche-cli/main/versions/latest.json), CLI will need to know how to process these changes to function.


Tonix517 (approved):

pkg/version/min_cli_version.go
```diff
@@ -0,0 +1,45 @@
...
+	if !strings.HasPrefix(minVersion, "v") {
+		minVersion = "v" + minVersion
+	}
+	if !strings.HasPrefix(version, "v") {
+		version = "v" + version
+	}
+
+	versionComparison := semver.Compare(version, minVersion)
+	if versionComparison == -1 {
+		return fmt.Errorf("CLI version is required to be at least %s, current CLI version is %s, please upgrade CLI by calling `curl -sSfL https://raw.githubusercontent.com/ava-labs/avalanche-cli/main/scripts/install.sh | sh -s`", minVersion, version)
```
felipemadero, line 42: can you also add `avalanche update` to the upgrade instructions?

felipemadero: this as the first option, and the curl as second one

sukantoraymond: addreessed

PR #2849, opened 2025-06-11
Deprecating --hard option in clean command

Tonix517:


sukantoraymond:
Screenshot 2025-06-12 at 4 16 07 PM
So clean network -- hard will actually remove the entire bin directory. It contains all the binaries listed above. Why do we want to delete hard option? @felipemadero

cmd/networkcmd/clean.go
```diff
@@ -59,11 +50,9 @@ func clean(*cobra.Command, []string) error {
...
 	}
 
-	if hard {
-		ux.Logger.PrintToUser("hard clean requested via flag, removing all downloaded avalanchego and plugin binaries")
-		binDir := filepath.Join(app.GetBaseDir(), constants.AvalancheCliBinDir)
-		cleanBins(binDir)
-	}
+	ux.Logger.PrintToUser("removing all downloaded avalanchego and plugin binaries")
+	binDir := filepath.Join(app.GetBaseDir(), constants.AvalancheCliBinDir)
+	cleanBins(binDir)
```
felipemadero, line 55: I mean, we don't even need this functionality just don;t do it. Lets keep the downloaded binaries after a clean.

tests/e2e/testcases/network/suite.go
```diff
@@ -117,7 +117,7 @@ var _ = ginkgo.Describe("[Network]", ginkgo.Ordered, func() {
 		commands.DeleteSubnetConfig(subnetName)
 	})
 
-	ginkgo.It("clean hard deletes plugin binaries non SOV", func() {
+	ginkgo.It("clean deletes plugin binaries non SOV", func() {
```
felipemadero, line 120: we don't need this one, if we are not removing binaries anymore

felipemadero:
> Screenshot 2025-06-12 at 4 16 07 PM So clean network -- hard will actually remove the entire bin directory. It contains all the binaries listed above. Why do we want to delete hard option? @felipemadero

we don't want to continue deleted this, I also made the same comment in parallel

cmd/networkcmd/clean.go
Tonix517: removed this block

felipemadero (approved):
LGTM but needs to pass CI

PR #2856, opened 2025-06-14
improve proposer vm setup

felipemadero:
## Why this should be merged
currently some users are seeing errors similar to this:
```
client.SendTransaction failure at step 1 ... nonce too low: next nonce 18, tx nonce 17
``1
this corresponds to a failed nonce estimation before creating a second tx on the proposer vm setup.

## How this works
This PR aleviates this by keeping a local copy of the block number and nonce sequence, altogether 
with adding a timed delay between activation txs, and checking agains API if the nonce is even greater
(to try to reduce issues where the account was externally used).

## How this was tested

## How is this documented

### DIFF
--- a/sdk/evm/evm.go
+++ b/sdk/evm/evm.go
@@ -678,11 +678,11 @@ func (client Client) CreateDummyBlocks(
 	txSigner := types.LatestSignerForChainID(chainID)
 	blockNumber, err := client.BlockNumber()
 	if err != nil {
-		return fmt.Errorf("client.BlockNumber failure: %w", err)
+		return fmt.Errorf("unable to get block number: %w", err)
 	}
 	nonce, err := client.NonceAt(addr.Hex())
 	if err != nil {
-		return fmt.Errorf("client.NonceAt failure: %w", err)
+		return fmt.Errorf("unable to get nonce: %w", err)
 	}
 	for i := 0; i < numBlocks; i++ {
 		// it may be the case that we hit an outdated node with the rpc, so lets not fully trust the API
--- a/sdk/evm/evm.go
+++ b/sdk/evm/evm.go
@@ -676,14 +676,27 @@ func (client Client) CreateDummyBlocks(
 	}
 	gasPrice := big.NewInt(legacy.BaseFee)
 	txSigner := types.LatestSignerForChainID(chainID)
+	blockNumber, err := client.BlockNumber()
+	if err != nil {
+		return fmt.Errorf("client.BlockNumber failure: %w", err)
+	}
+	nonce, err := client.NonceAt(addr.Hex())
+	if err != nil {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: sdk/evm/evm.go [hunk 1]
```diff
@@ -676,14 +676,27 @@ func (client Client) CreateDummyBlocks(
 	}
 	gasPrice := big.NewInt(legacy.BaseFee)
 	txSigner := types.LatestSignerForChainID(chainID)
+	blockNumber, err := client.BlockNumber()
+	if err != nil {
+		return fmt.Errorf("client.BlockNumber failure: %w", err)
+	}
+	nonce, err := client.NonceAt(addr.Hex())
+	if err != nil {
+		return fmt.Errorf("client.NonceAt failure: %w", err)
```
sukantoraymond, line 685: same here should we rename it to "unable to get nonce "

sdk/evm/evm.go
```diff
@@ -676,14 +676,27 @@ func (client Client) CreateDummyBlocks(
...
+	}
 	for i := 0; i < numBlocks; i++ {
-		prevBlockNumber, err := client.BlockNumber()
-		if err != nil {
+		// it may be the case that we hit an outdated node with the rpc, so lets not fully trust the API
+		if blockNumberFromAPI, err := client.BlockNumber(); err != nil {
 			return fmt.Errorf("client.BlockNumber failure at step %d: %w", i, err)
+		} else if blockNumberFromAPI > blockNumber {
+			// changes from outside
+			blockNumber = blockNumberFromAPI
```
sukantoraymond, line 693: why not we always use blockNumberFromAPI instead?

sdk/evm/evm.go
```diff
@@ -676,14 +676,27 @@ func (client Client) CreateDummyBlocks(
...
 		}
-		nonce, err := client.NonceAt(addr.Hex())
-		if err != nil {
+		if nonceFromAPI, err := client.NonceAt(addr.Hex()); err != nil {
 			return fmt.Errorf("client.NonceAt failure at step %d: %w", i, err)
+		} else if nonceFromAPI > nonce {
+			// changes from outside
+			nonce = nonceFromAPI
```
sukantoraymond, line 699: same here why not we always use nonceFromAPI?

sdk/evm/evm.go
felipemadero: because API can be outdated, and we have locally better information in that case. we now a new tx is already accepted.

sdk/evm/evm.go
felipemadero: same response. nonce from API can be outdated, in this case we trust in local logic. on the other hand,
if the nonce is not outdated but > local one, we trust on the API

sukantoraymond (approved):

Issue #2873, opened 2025-07-02
[Bug]: Getting a permission error when trying to create a node on GCP


rochana-atapattu:
**Describe the bug**
I followed the steps after `avalanche node create test` command, mostly with default. The VM images seem to be private.

**To Reproduce**
run `avalanche node create test --gcp`
select default values.

**Expected behavior**

**Logs**
Error: googleapi: Error 403: Required 'compute.images.list' permission for 'projects/avalabs-experimental', forbidden
[07-02|16:36:07.201] INFO ux/output.go:43 Error: googleapi: Error 403: Required 'compute.images.list' permission for 'projects/avalabs-experimental', forbidden

**Operating System**
macOS

felipemadero:
This error occurs because the CLI attempts to use VM images from a private Ava Labs GCP project (avalabs-experimental) that external users don't have access to.

With the CLI now in maintenance mode, GCP cloud support won't be actively developed by the team. If you find a fix (e.g., using public images or allowing users to specify their own image project), we welcome PRs and will review them.

In the meantime, AWS cloud provisioning may be a more reliable option if that works for your use case.

PR #2874, opened 2025-07-02
fix: handle if calling to list regions fail

rochana-atapattu:
## Why this should be merged
handling a missed error check

## How this works
propagating the error and handle it 

## How this was tested
running the command to different inputs, can't validate it full because if #2873


felipemadero (approved):

felipemadero:
Thank you for the contribution! We've included your fix in #2994.

We couldn't merge your PR directly due to branch protection requiring GPG-signed commits, so we applied the changes on our end.

Great catch on the missing error handling!

PR #2878, opened 2025-07-03
fix acp99 pos import

felipemadero:
## Why this should be merged
Currently, `avalanche blockchain import public --rpc https://subnets.avax.network/beam/mainnet/rpc --mainnet` does not recognize beam as a PoS L1, but PoA. And incorrect validator manager contract is stored on sidercar.


cmd/blockchaincmd/import_public.go
```diff
@@ -221,17 +222,29 @@ func importBlockchain(
...
 			}
-			printFunc("  Validation Kind: %s", sc.ValidatorManagement)
 			if sc.ValidatorManagement == validatormanagertypes.ProofOfAuthority {
 				owner, err := contract.GetContractOwner(rpcURL, common.HexToAddress(validatorManagerAddress))
 				if err != nil {
 					return models.Sidecar{}, nil, err
 				}
-				sc.ValidatorManagerOwner = owner.String()
+				// check if the owner is a specialized validator manager
+				validatorManagement := validatorManagerSDK.GetValidatorManagerType(rpcURL, owner)
```
sukantoraymond, line 235: Can you provide example of how we could get to this point? Don't really understand it

sdk/validatormanager/root.go
```diff
@@ -322,22 +321,18 @@ func InitializeValidatorsSet(
...
-		return validatormanagertypes.UndefinedValidatorManagement, err
+	if _, err := PoSWeightToValue(rpcURL, managerAddress, 0); err == nil {
+		return validatormanagertypes.ProofOfStake
+	}
+	// verify it is PoA
+	if _, err := validator.GetValidationID(rpcURL, managerAddress, ids.EmptyNodeID); err == nil {
+		return validatormanagertypes.ProofOfAuthority
 	}
-	return validatormanagertypes.ProofOfStake, nil
+	return validatormanagertypes.UndefinedValidatorManagement
```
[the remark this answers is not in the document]
felipemadero: previous implementation assumed all validators (PoA/PoS) responded to GetValidationID. this is not the case
anymore with ACP99. now only PoA responds to GetValidationID and PoS responds to PoSWeightToValue

cmd/blockchaincmd/import_public.go
felipemadero: added a explanatory comment

cmd/blockchaincmd/import_public.go
```diff
@@ -191,6 +191,7 @@ func importBlockchain(
 	if err != nil {
 		return models.Sidecar{}, nil, err
 	}
+	printFunc("  Sovereign: %v", !subnetInfo.IsPermissioned)
```
sukantoraymond, line 194: I dont think we need to print this. Let's just print if blockchain is Subnet (non sovereign)

PR #2898, opened 2025-07-30
Refactor contract sdk

sukantoraymond:
While moving sdk out into its own repo, foudn that validator manager sdk is using pkg/contract. This PR is made to replace pkg/contract to sdk/contract

### DIFF
--- a/pkg/validatormanager/weight_update.go
+++ b/pkg/validatormanager/weight_update.go
@@ -147,7 +147,7 @@ func InitValidatorWeightChange(
 			ux.Logger.PrintToUser("Validator weight change initialized. InitiateTxHash: %s", tx.Hash())
 		}
 	} else {
-		logger.Error(logging.LightBlue.Wrap("The validator weight change process was already initialized. Proceeding to the next step"))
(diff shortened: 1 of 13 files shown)


pkg/validatormanager/weight_update.go
```diff
@@ -57,6 +61,7 @@ func InitializeValidatorWeightChange(
 
 func InitValidatorWeightChange(
 	ctx context.Context,
+	logger logging.Logger,
```
sukantoraymond, line 64: Seems that currently we have printFunc -> where we use ux.Logger.PrintToUser. Do we want logger or printFunc? @felipemadero

pkg/ux/output.go
```diff
@@ -18,20 +18,20 @@ import (
 var Logger *UserLog
 
 type UserLog struct {
-	log    logging.Logger
+	Log    logging.Logger
```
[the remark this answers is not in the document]
felipemadero: np

pkg/validatormanager/weight_update.go
felipemadero: can you remove the printFunc usage in this PR?

Issue #2718, opened 2025-04-03
avoid using context.Background() outside of main() init()


felipemadero:
Based on a comment from @ARR4N .

Ref https://github.com/ava-labs/avalanche-cli/pull/2717#discussion_r2026827712 

[more](https://google.github.io/styleguide/go/decisions.html#:~:text=It%20is%20very%20rare%20for%20code%20in%20the%20middle%20of%20a%20callchain%20to%20require%20creating%20a%20base%20context%20of%20its%20own%20using%20context.Background().%20Always%20prefer%20taking%20a%20context%20from%20your%20caller%2C%20unless%20it%E2%80%99s%20the%20wrong%20context.)

PR #2904, opened 2025-08-04
use sdk context everywhere except unit tests

felipemadero:
## Why this should be merged
Closes https://github.com/ava-labs/avalanche-cli/issues/2718

Currently context.Background() is used on some parts, and in other parts direct context.WithTimeout that does
not take into account trapped interruptions.


cmd/blockchaincmd/join.go
```diff
@@ -325,8 +325,9 @@ func writeAvagoChainConfigFiles(
 
 func checkIsValidating(subnetID ids.ID, nodeID ids.NodeID, pClient platformvm.Client) (bool, error) {
 	// first check if the node is already an accepted validator on the subnet
-	ctx := context.Background()
 	nodeIDs := []ids.NodeID{nodeID}
+	ctx, cancel := sdkutils.GetAPIContext()
```
sukantoraymond, line 329: how do we decide if GetAPIContext or GetAPILargeContext

cmd/keycmd/transfer.go
```diff
@@ -931,9 +946,11 @@ func cToPSend(
 	usingLedger bool,
 	amount uint64,
 ) error {
+	ctx, cancel := sdkutils.GetTimedContext(constants.WalletCreationTimeout)
```
sukantoraymond, line 949: why not just use getlargeapicontext here

pkg/blockchain/helper.go
```diff
@@ -61,7 +61,7 @@ func GetAggregatorNetworkUris(app *application.Avalanche, clusterName string) ([
 
 func UrisToPeers(uris []string) ([]info.Peer, error) {
 	peers := []info.Peer{}
-	ctx, cancel := utils.GetANRContext()
+	ctx, cancel := sdkutils.GetTimedContext(3 * time.Minute)
```
sukantoraymond, line 64: seems that we need constnat for anr context

sukantoraymond (approved):

cmd/blockchaincmd/join.go
felipemadero: large context is used for txs that has been observed to delay on the blockchain request. typically 
a request that takes more time than expected because of data intensive operation.

cmd/keycmd/transfer.go
felipemadero: because wallet creation can retrieve P/X/C utxos for an account and depending on the account, the operation
may take its time

PR #2909, opened 2025-08-07
add prepare create subnet and chain tx

felipemadero:

### DIFF
--- a/sdk/pchain/new_create_chain_tx.go
+++ b/sdk/pchain/new_create_chain_tx.go
@@ -1,0 +1,66 @@
+// Copyright (C) 2025, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+package pchain
+
+import (
+	"fmt"
+
+	"github.com/ava-labs/avalanche-cli/sdk/multisig"
+	"github.com/ava-labs/avalanche-cli/sdk/wallet"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/formatting/address"
+	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
+)
+
+// CreateChainTxParams contains all parameters needed to create a CreateChainTx
+type CreateChainTxParams struct {
+	// SubnetID is the ID of the subnet where the chain will be created
+	SubnetID string
+	// VMID is the ID of the virtual machine to be associated to the chain
+	VMID string
+	// ChainName is the chain name
+	ChainName string
+	// Genesis details the contents of the genesis block
+	Genesis []byte
+	// SubnetAuthKeys are the subset of the subnet control keys used to
+	// authorize the operation
+	SubnetAuthKeys []string
+}
+
+// NewCreateChainTx creates uncommitted CreateChainTx
+// keychain in wallet will be used to build the transaction
+func NewCreateChainTx(
+	client wallet.Wallet,
+	createChainTxParams CreateChainTxParams,
+) (*multisig.Multisig, error) {
+	fxIDs := make([]ids.ID, 0)
+	subnetAuthKeys, err := address.ParseToIDs(createChainTxParams.SubnetAuthKeys)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing control keys: %w", err)
+	}
+	subnetID, err := ids.FromString(createChainTxParams.SubnetID)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing subnet ID: %w", err)
+	}
+	vmID, err := ids.FromString(createChainTxParams.VMID)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing vm ID: %w", err)
+	}
+	// NOT SURE ON THIS! we are modifying the client as a side effect
+	// We may want to just get options for the builder, as in original deployer
+	if err := client.SetSubnetAuthMultisig(subnetAuthKeys); err != nil {
+		return nil, err
+	}
+	unsignedTx, err := client.P().Builder().NewCreateChainTx(
+		subnetID,
+		createChainTxParams.Genesis,
+		vmID,
+		fxIDs,
+		createChainTxParams.ChainName,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error building tx: %w", err)
+	}
+	tx := txs.Tx{Unsigned: unsignedTx}
+	return multisig.New(&tx), nil
+}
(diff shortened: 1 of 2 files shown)


sdk/pchain/send_transaction.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"time"
+
+	"github.com/ava-labs/avalanche-cli/sdk/multisig"
+	"github.com/ava-labs/avalanche-cli/sdk/utils"
+	"github.com/ava-labs/avalanche-cli/sdk/wallet"
+	"github.com/ava-labs/avalanchego/wallet/subnet/primary/common"
+)
+
+func SendTransaction(
```
sukantoraymond, line 14: Why don't we just move sending transaction to wallet like https://github.com/ava-labs/avalanche-tooling-sdk-go/blob/fef92d1127dc37da50a9307a78663b81a04b971e/wallet/wallet.go#L99?

sdk/keychain/keychain.go
```diff
@@ -49,6 +50,20 @@ type Ledger struct {
 	LedgerIndices []uint32
 }
 
+func PrivateKeyToAvalancheAccount(
```
sukantoraymond, line 53: why do we want this? Why don't we just reuse https://github.com/ava-labs/avalanche-tooling-sdk-go/blob/fef92d1127dc37da50a9307a78663b81a04b971e/wallet/p-chain/examples/convertSubnetToL1Tx.go#L51?

It handles ledger too

sukantoraymond: Now that i think about it, we should create wallet object directly from private key / private key path, no need for intermediary keychain / we can abstract it out

sdk/keychain/keychain.go
```diff
@@ -126,13 +139,65 @@ func (kc *Keychain) AddLedgerAddresses(addresses []string) error {
...
+		indices, err := kc.Ledger.LedgerDevice.FindFunds(network, amount, 0)
 		if err != nil {
 			return err
 		}
 		return kc.AddLedgerIndices(indices)
 	}
 	return fmt.Errorf("keychain is not ledger enabled")
 }
+
+func (kc *Keychain) P(
```
sukantoraymond, line 156: can we get address directly from avalanche go Wallet.P() object instead?

sukantoraymond: Seems that it will be something like
```
key := wallet.Keychain.Keychain // This gives you access to the underlying key
pAddress, err := key.P(network.HRP) // network.HRP would be "fuji" for testnet, "avax" for mainnet
```

sukantoraymond: Point of consideration: Should this be in wallet object instead?

sdk/key/soft_key.go
```diff
@@ -283,12 +283,12 @@ func (m *SoftKey) Save(p string) error {
 	return os.WriteFile(p, []byte(m.PrivKeyHex()), constants.UserOnlyWriteReadPerms)
 }
 
-func (m *SoftKey) P(networkHRP string) (string, error) {
-	return address.Format("P", networkHRP, m.privKey.PublicKey().Address().Bytes())
+func (m *SoftKey) P(network network.Network) (string, error) {
```
sukantoraymond, line 286: we already have this P object, why do we need another keychain P chain function?

sdk/pchain/prepare_create_chain_tx.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/ava-labs/avalanche-cli/sdk/multisig"
+	"github.com/ava-labs/avalanche-cli/sdk/wallet"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/formatting/address"
+	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
+)
+
+// PrepareCreateChainTx creates uncommitted CreateChainTx
+// keychain in wallet will be used to build the transaction
+func PrepareCreateChainTx(
```
sukantoraymond, line 17: also dont think we need to use the same name as typescript SDK. NewCreateChain that we have works better IMO

sdk/pchain/prepare_create_chain_tx.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+
+	"github.com/ava-labs/avalanche-cli/sdk/multisig"
+	"github.com/ava-labs/avalanche-cli/sdk/wallet"
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/formatting/address"
+	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
+)
+
```
sukantoraymond, line 14: thinking this should be on wallet

sukantoraymond: https://github.com/ava-labs/avalanche-tooling-sdk-go/blob/fef92d1127dc37da50a9307a78663b81a04b971e/wallet/p-chain/txs/convertSubnetToL1Tx.go

sdk/pchain/prepare_create_chain_tx.go
```diff
@@ -0,0 +1,55 @@
...
+		return nil, fmt.Errorf("error parsing control keys: %w", err)
+	}
+	subnetID, err := ids.FromString(subnetIDStr)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing subnet ID: %w", err)
+	}
+	vmID, err := ids.FromString(vmIDStr)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing vm ID: %w", err)
+	}
```
sukantoraymond, line 37: i guess in here we are deciding to let user use string instead of ids.ID. Im not opposed to it

sdk/pchain/send_transaction.go
felipemadero: I have not issues in moving this into wallet, but I have a slight preference for non OOP style, which is also the TS SDK style

felipemadero: anyway, we can have both options (if we find out the TS SDK provides both options)

sdk/keychain/keychain.go
felipemadero: I am trying to align to SDK. they have this account abstraction, very similar to the keychain. Also the function
is there. I believe it is handy. Anyway, not opposite to also having it in wallet. But I would have separate function
for the options:
- ledger
- path
- string

sdk/keychain/keychain.go
felipemadero: I believe the account can have it. Not opposed to also have it on wallet.

PR #2913, opened 2025-08-08
Fix pos

felipemadero:
## Why this should be merged
Fixes issues of PoS implementation:
- changes current default PoS flag values of weight to value factor 1, minimumStakeAmount 1, maximumStakeAmount 1000, to 1e18, 1e18, 1000e18, so the default staking unit used at flags is now the token, instead of wei (really there was no way to specify stake of say 20 tokens due to parsing limitations in flags, so most of the CLI PoS operations in the past revolved around 20 wei)
- churn rate check is not based in vmc settings 
- enable setting of bootstrap validator weight, so as to enable better control in the replacement of weight with stakes, if migrating from PoA to PoS while keeping the same validators (you can now use a weight of 20 if you later on want to stake 20 tokens in the same validator after a migration)
- enable specifying the private key to be used for the staker when adding validator, and the validator owner (staker again) when removing validators is PoS (previously only the PoA validator manager owner was doing the stake)
- display and verify the maximum available weight (stake) when prompting for weight while adding a validator
- require change weight command to be only available for PoA
- require the provision of a native minter precompile admin when initializing PoS, for the case the contract is not yet a native minter 
- for PoS deploy on deploy command, also deploy a reward calculator
- add columns for owner, stake, start time, end time, ready to be removel, to validator list
- add check for validator owner previous to initializing removal
- add check for appriate end time previous to initializing removal
- move precompiles functions to SDK
- add check that the native minter is on and that the PoS is enabled previous to initializing it


cmd/blockchaincmd/add_validator.go
```diff
@@ -124,6 +126,8 @@ Testnet or Mainnet.`,
 	cmd.Flags().StringVar(&nodeEndpoint, "node-endpoint", "", "gather node id/bls from publicly available avalanchego apis on the given endpoint")
 	cmd.Flags().Uint64Var(&weight, validatorWeightFlag, uint64(constants.DefaultStakeWeight), "set the weight of the validator")
 	cmd.Flags().StringVar(&validatorManagerOwner, "validator-manager-owner", "", "force using this address to issue transactions to the validator manager")
+	stakerPrivateKeyFlags.SetFlagNames("staker-private-key", "staker-key", "staker-genesis-key")
```
sukantoraymond, line 129: we shouldnt have multiple key names for the same thing IMO. One key name is easier to users.
maybe just staker key

sukantoraymond: also might not be secure in case users are using CLI in backend to expose private key, maybe tell users to first create the key through key create and then using the key name as input

cmd/blockchaincmd/add_validator.go
```diff
@@ -124,6 +126,8 @@ Testnet or Mainnet.`,
...
+	stakerPrivateKeyFlags.AddToCmd(cmd, "as source of staking funds")
```
sukantoraymond, line 130: how is the provider of this fund different from the provider of the fund for previous steps (provider of initial balance to validators, PoA)

cmd/blockchaincmd/add_validator.go
```diff
@@ -252,15 +257,43 @@ func addValidator(cmd *cobra.Command, args []string) error {
...
+		)
+		if err != nil {
+			return err
+		}
+		allowedChange := float64(totalWeight*uint64(churnSettings.MaximumChurnPercentage)) / 100.0
 		if !cmd.Flags().Changed(validatorWeightFlag) {
 			weight, err = app.Prompt.CaptureWeight(
-				"What weight would you like to assign to the validator?",
-				func(uint64) error { return nil },
+				fmt.Sprintf("What weight would you like to assign to the validator (max=%d)?", uint64(allowedChange)),
```
sukantoraymond, line 282: can u pls explain what weight change here means? How is this different from the weight assigned during PoA? Why are we changing the weight again during PoS?

cmd/blockchaincmd/add_validator.go
```diff
@@ -252,15 +257,43 @@ func addValidator(cmd *cobra.Command, args []string) error {
...
+				func(weight uint64) error {
+					if float64(weight) > allowedChange {
+						return fmt.Errorf("can't make change: desired validator weight %d exceeds max allowed weight change of %d", weight, uint64(allowedChange))
```
sukantoraymond, line 285: can we pls have a test in e2e that tests this? where we try to add a validator with weight more than allowed

cmd/blockchaincmd/change_weight.go
```diff
@@ -189,7 +194,15 @@ func setWeight(_ *cobra.Command, args []string) error {
...
-		allowedChange := float64(totalWeight) * constants.MaxL1TotalWeightChange
+		churnSettings, err := validatormanagersdk.GetChurnSettings(
+			validatorManagerRPCEndpoint,
+			common.HexToAddress(validatorManagerAddress),
+		)
+		if err != nil {
+			return err
+		}
+
+		allowedChange := float64(totalWeight*uint64(churnSettings.MaximumChurnPercentage)) / 100.0
```
sukantoraymond, line 205: make this into a function since it was also used in add validator?

cmd/blockchaincmd/remove_validator.go
```diff
@@ -76,6 +76,8 @@ these prompts by providing the values with flags.`,
 	cmd.Flags().StringVar(&validatorManagerOwner, "validator-manager-owner", "", "force using this address to issue transactions to the validator manager")
 	cmd.Flags().StringVar(&initiateTxHash, "initiate-tx-hash", "", "initiate tx is already issued, with the given hash")
 	cmd.SetHelpFunc(flags.WithGroupedHelp([]flags.GroupedFlags{sigAggGroup}))
+	stakerPrivateKeyFlags.SetFlagNames("staker-private-key", "staker-key", "staker-genesis-key")
```
sukantoraymond, line 79: same here about one key name

cmd/blockchaincmd/remove_validator.go
```diff
@@ -280,21 +282,56 @@ func removeValidatorSOV(
...
+			)
+			if err != nil {
+				return err
+			}
+		}
+		address, err := evm.PrivateKeyToAddress(privateKey)
+		if err != nil {
+			return err
+		}
+		from = address.Hex()
```
sukantoraymond, line 317: lets make all these into a function

sdk/validatormanager/validator_manager_poa.go
```diff
@@ -61,9 +60,97 @@ func PoAValidatorManagerInitialize(
...
+	}
+	return getValidatorReturn, nil
+}
+
+type ChurnSettings struct {
+	ChurnPeriodSeconds     uint64
+	MaximumChurnPercentage uint8
+}
+
+func GetChurnSettings(
```
sukantoraymond, line 121: is churn Poa / pos related? Maybe we should have validator_manager.go instead

tests/e2e/testcases/subnet/sov/addRemoveValidatorPoS/suite.go
```diff
@@ -177,153 +177,3 @@ var _ = ginkgo.Describe("[AddRemove Validator SOV L1 Manager PoS]", func() {
 		commands.DeleteSubnetConfig(utils.BlockchainName)
 	})
 })
-
-var _ = ginkgo.Describe("[AddRemove Validator SOV External Manager PoS]", func() {
```
sukantoraymond, line 181: why are we removing this

cmd/blockchaincmd/add_validator.go
felipemadero: I believe it is of convenience to provide options to input a key name, or just directly the private key, or better genesis key, probably this is for another PR or discussion as there are many places where this options are received

tests/e2e/testcases/subnet/sov/addRemoveValidatorPoS/suite.go
felipemadero: sov on c-chain stopped working as external manager is on c-chain which does not have native minter precompile. it was working before because previously there was no reward paiment so the precompile issue was not hit

sdk/validatormanager/validator_manager_pos.go
```diff
@@ -135,3 +170,118 @@ func PoSWeightToValue(
...
+		return getStakingValidatorReturn, fmt.Errorf("invalid type for delegationFeeBips output of getStakingValidator: expected uint16, got %T", out[1])
+	}
+	getStakingValidatorReturn.MinStakeDuration, ok = out[2].(uint64)
+	if !ok {
+		return getStakingValidatorReturn, fmt.Errorf("invalid type for minStakeDuration output of getStakingValidator: expected uint64, got %T", out[2])
+	}
+	getStakingValidatorReturn.UptimeSeconds, ok = out[3].(uint64)
+	if !ok {
+		return getStakingValidatorReturn, fmt.Errorf("invalid type for uptimeSeconds output of getStakingValidator: expected uint64, got %T", out[3])
+	}
```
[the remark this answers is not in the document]
felipemadero: not understanding, how should be the implementation

sdk/validatormanager/validator_manager_poa.go
```diff
@@ -61,9 +60,97 @@ func PoAValidatorManagerInitialize(
...
+	if len(out) != 1 {
+		return getValidatorReturn, fmt.Errorf("incorrect number of outputs for getValidator: expected 1 got %d", len(out))
+	}
+	var ok bool
+	getValidatorReturn, ok = abi.ConvertType(out[0], new(GetValidatorReturn)).(*GetValidatorReturn)
+	if !ok {
+		return getValidatorReturn, fmt.Errorf("invalid type for output of getValidator: expected GetValidatorReturn, got %T", out[0])
+	}
+	return getValidatorReturn, nil
+}
```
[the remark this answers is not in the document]
felipemadero: in validator list we take a lot of info from pchain, this function interacts with the L1. the important part that we get from here is the start time inside the validator manager, not available otherwise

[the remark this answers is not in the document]
felipemadero: when interacting with pchain is it useful to get the remaining pchain validator balance

cmd/blockchaincmd/add_validator.go
felipemadero: , probably this is for another PR or discussion as there are many places where this options are accepted.

cmd/blockchaincmd/change_weight.go
felipemadero: the function is useful itself, but not useful in change weight because the logic use the intermediate variables
more than once. but adding a function just to be used in add validator and as general utility on SDK

PR #2942, opened 2025-09-09
chore: remove redundant words

keeghcet:
## Why this should be merged

remove redundant words


felipemadero:
Thank you for the contribution! These typo fixes were already included in #2994 (from an earlier PR with the same changes).

Appreciate you taking the time!

PR #2948, opened 2025-09-18
Custom vm path for devnet

sukantoraymond:
Copies custom vm binary specified in custom vm path flag for `node devnet wiz` command to remote hosts


felipemadero (commented):
This is a really nice add on, but we should try to separate the local binary from the cloud binary as they are not necessarily going to be the same and we probably need both.

Some extra comments:
1) the binary specified by vmPath should be already available at the ~/.avalanche-cli/vms/, so not need to persist the original path in the sidecar
2) persisting a path given by the user in sidecar exposes to later fail, if the user removes this file, it is not managed by CLI so  should be copied into ~/.avalanche-cli, probably 
3) this brakes local neworks tests for the vm, because you are passing in a binary for aws, that is stored and then used in local networks

